Issue #245, opened 2026-01-18
Pointer parameters for value types cause compilation errors with rvalue arguments


jlaustill:
## Description

When passing expressions (rvalues) to functions that take value-type parameters like `u8`, the transpiler generates code that doesn't compile. The transpiler converts value parameters to pointer parameters, but then generates call sites that pass expressions instead of addressable lvalues.

## Reproduction

```cnx
scope example {
    u32 process(u32 crc, u8 byte) {
        return crc ^ byte;
    }

    u32 test(u32 value) {
        u32 crc <- 0;
        // This fails - passing expression to u8 parameter
        crc <- this.process(crc, value & 0xFF);
        return crc;
    }
}
```

## Generated C++ (problematic)

```cpp
static uint32_t example_process(uint32_t* crc, uint8_t* byte) {
    return (*crc) ^ (*byte);
}

uint32_t example_test(uint32_t* value) {
    uint32_t crc = 0;
    // ERROR: cannot convert 'int' to 'uint8_t*'
    crc = example_process(&crc, (*value) & 0xFF);
    return crc;
}
```

## Compiler Error

```
error: invalid conversion from 'int' to 'uint8_t*' {aka 'unsigned char*'} [-fpermissive]
```

## Expected Behavior

Either:
1. Generate value parameters (not pointers) for primitive types, OR
2. Generate temporary variables at call sites for rvalue arguments

## Workaround

Use explicit temporary variables:

```cnx
u8 b <- value & 0xFF;
crc <- this.process(crc, b);
```

## Environment

- C-Next version: 0.1.10
- Platform: Linux

jlaustill:
Fixed in PR #247

PR #247, opened 2026-01-19
fix: wrap rvalue expressions in compound literals for pointer params (Issue #245)

jlaustill:
## Summary

- Fixes Issue #245 where rvalue expressions (like `value & 0xFF`) passed to functions with value-type parameters (`u8`) generated invalid C code
- The transpiler now wraps ALL rvalues in compound literals when passed to pointer parameters, not just literals
- Change in `_generateFunctionArg()` at `CodeGenerator.ts:3190`: removed `isLiteralExpression()` check since any expression reaching that point is an rvalue that needs wrapping

## Test plan

- [x] Added test `tests/functions/rvalue-arg.test.cnx` that demonstrates the fix
- [x] Generated C code now uses `&(uint8_t){value & 0xFF}` instead of raw `value & 0xFF`
- [x] Fixed code compiles cleanly with `gcc -Wall -Werror` (no pointer-from-integer warnings)
- [x] Full test suite passes (614/614 tests)

### DIFF
--- a/tests/functions/rvalue-arg.test.cnx
+++ b/tests/functions/rvalue-arg.test.cnx
@@ -10,7 +10,7 @@ u32 main() {
     u32 crc <- 0;
 
     // Test: passing bitwise expression (rvalue) to u8 parameter
-    // Should generate: process(&crc, &(uint8_t){(*value) & 0xFF})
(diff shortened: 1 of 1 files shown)


jlaustill:
Note: This PR fixes #245 (using auto-close syntax for GitHub to link and close the issue when merged).

Issue #252, opened 2026-01-19
C++ mode: enum and bool types not converted when passed to u8 parameters


jlaustill:
## Description

When passing C++ typed enum (`enum class` / `enum : uint8_t`) or `bool` struct members to a function expecting `u8`, the generated code passes the wrong pointer type.

## Reproduction

```cnx
// Given C++ header with:
// enum EPressureType : uint8_t { ... };
// enum EThermocoupleType : uint8_t { ... };
// struct Config { bool enabled; EPressureType ptype; EThermocoupleType tctype; };

scope example {
    u32 process(u32 crc, u8 byte) {
        return crc ^ byte;
    }

    public u32 test(const Config cfg) {
        u32 crc <- 0;
        crc <- this.process(crc, cfg.enabled);   // bool
        crc <- this.process(crc, cfg.ptype);     // EPressureType  
        crc <- this.process(crc, cfg.tctype);    // EThermocoupleType
        return crc;
    }
}
```

## Generated C++ (problematic)

```cpp
// These pass the wrong pointer types:
crc = example_process(&crc, &config->enabled);       // bool*
crc = example_process(&crc, &config->ptype);         // EPressureType*
crc = example_process(&crc, &config->tctype);        // EThermocoupleType*
```

## Compiler Errors

```
error: cannot convert 'const bool*' to 'uint8_t*' {aka 'unsigned char*'}
error: cannot convert 'const EPressureType*' to 'uint8_t*' {aka 'unsigned char*'}
error: cannot convert 'const EThermocoupleType*' to 'uint8_t*' {aka 'unsigned char*'}
```

## Expected Behavior

Generate explicit conversion, e.g.:
```cpp
uint8_t _tmp = static_cast<uint8_t>(config->enabled);
crc = example_process(&crc, &_tmp);
```

## Notes

- C++ typed enums (`enum Foo : uint8_t`) are distinct types from `uint8_t`
- `bool` is also a distinct type from `uint8_t` in C++
- This requires either explicit casts or temporary variables

## Environment

- C-Next version: 0.1.11
- Mode: `--cpp`

Issue #251, opened 2026-01-19
C++ mode: const correctness when passing struct members to functions


jlaustill:
## Description

When a function takes a `const` struct parameter and passes its members to another function, the generated code passes `const T*` where `T*` is expected.

## Reproduction

```cnx
scope example {
    u32 process(u32 crc, u8 byte) {
        return crc ^ byte;
    }

    public u32 test(const MyStruct data) {
        u32 crc <- 0;
        crc <- this.process(crc, data.value);  // data.value is u8
        return crc;
    }
}
```

## Generated C++ (problematic)

```cpp
static uint32_t example_process(uint32_t* crc, uint8_t* byte) {
    return (*crc) ^ (*byte);
}

uint32_t example_test(const MyStruct* data) {
    uint32_t crc = 0;
    // ERROR: const uint8_t* passed to uint8_t*
    crc = example_process(&crc, &data->value);
    return crc;
}
```

## Compiler Error

```
error: invalid conversion from 'const uint8_t*' to 'uint8_t*' [-fpermissive]
```

## Expected Behavior

Either:
1. Generate `const uint8_t*` parameter when the value is read-only
2. Generate a copy to a local variable before passing

## Environment

- C-Next version: 0.1.11
- Mode: `--cpp`

jlaustill:
## Fix Implemented

**Branch:** `fix/cpp-struct-member-args-251-252`

### The Fix

Added `needsCppMemberConversion()` helper in `CodeGenerator.ts` that detects when passing a const struct member to a non-const parameter in C++ mode.

When detected, generates a temp variable to break the const chain:
```cpp
// Before (broken):
crc = process(&crc, &(*cfg).value);  // const uint8_t* to uint8_t*

// After (fixed):
uint8_t _cnx_tmp_0 = static_cast<uint8_t>((*cfg).value);
crc = process(&crc, &_cnx_tmp_0);
```

### Test
`tests/functions/const-struct-member-cpp.test.cnx` - validates const struct member passing

### Files Changed
- `src/codegen/CodeGenerator.ts` - added helper method and updated `_generateFunctionArg()`
- `scripts/test-utils.ts` - improved C++ mode detection in validation utilities

PR #253, opened 2026-01-19
fix: use temp variables instead of compound literals in C++ mode

### LINKED ISSUE #250: C++ mode: Compound literals are not addressable lvalues
## Description

The v0.1.11 fix for issue #245 uses C99 compound literals to handle rvalue arguments:

```cpp
crc = crc32_crcByte(&crc, &(uint8_t){config->magic & 0xFF});
```

However, in C++ mode (`--cpp`), compound literals are rvalues and cannot have their address taken. This generates:

```
error: lvalue required as unary '&' operand
```

## Additional Issues Found

1. **const correctness**: When passing struct members from a `const` parameter:
```cpp
// Generated - const uint8_t* passed to uint8_t*
crc = crc32_crcByte(&crc, &config->version);
```
Error: `invalid conversion from 'const uint8_t*' to 'uint8_t*'`

2. **Enum/bool type mismatches**: Enum and bool members passed to u8 functions:
```cpp
crc = crc32_crcByte(&crc, &config->pressureType);  // EPressureType*, not uint8_t*
crc = crc32_crcByte(&crc, &config->egtEnabled);    // bool*, not uint8_t*
```

## Suggested Fix for C++ Mode

Generate temporary variables instead of compound literals:

```cpp
// Instead of:
crc = crc32_crcByte(&crc, &(uint8_t){config->magic & 0xFF});

// Generate:
uint8_t _tmp1 = static_cast<uint8_t>(config->magic & 0xFF);
crc = crc32_crcByte(&crc, &_tmp1);
```

Or consider pass-by-value for small types in C++ mode.

## Environment

- C-Next version: 0.1.11
- Mode: `--cpp`
- Platform: Linux
- Compiler: g++ 13.x with `-std=c++17`

## Related

Follow-up to #245 which is fixed for C mode but not C++ mode.

jlaustill:
## Summary
- Fixes C++ mode compilation error where compound literals `&(uint8_t){...}` fail with "lvalue required as unary '&' operand"
- In C++ mode, generates temporary variables instead of compound literals for rvalue arguments
- Adds `cppMode` option to CodeGenerator that Pipeline sets based on `cppDetected` flag

## Test plan
- [x] Added test `tests/functions/rvalue-arg-cpp.test.cnx` with C++ header (typed enum) to force g++ compilation
- [x] Test validates both bitwise expressions (`value & 0xFF`) and arithmetic expressions (`a + b`) work correctly
- [x] All 616 tests pass

Fixes #250


jlaustill:
**Re: Reviewer Questions**

1. **Issues #251 (const correctness) and #252 (enum/bool types)** - These are separate issues and should be addressed in separate PRs. This PR focuses specifically on the compound literal rvalue issue.

2. **Edge cases** - The loop condition edge case has been addressed in the follow-up commit. Short-circuit and ternary expression evaluation changes remain as documented known limitations, but are low risk in practice.

PR #265, opened 2026-01-20
feat: add E0908 flow analysis for c_ variable NULL-check enforcement

jlaustill:
## Summary

Implements stricter NULL-check tracking for `c_` prefixed variables as part of ADR-046:

- **Re-assignment resets state**: Re-assigning a `c_` variable resets its NULL-check state (new value could be NULL)
- **While loop condition tracking**: Variables checked in while conditions are considered checked inside the loop body
- **E0908 enforcement**: Using `c_` variables before NULL-checking now reports E0908 error

## Implementation Details

Adds `INullableVariableState` interface and flow analysis with:
- State save/restore for if/while control flow
- Conservative semantics (restore pre-if state after if without else)
- While condition tracking via `inWhileCondition` flag

## Test plan

- [x] `e0908-use-before-check.test.cnx` - basic use-before-check error
- [x] `e0908-reassignment-resets.test.cnx` - re-assignment resets NULL-check state  
- [x] `valid-while-null-check.test.cnx` - while condition provides NULL check
- [x] All 635 tests pass


jlaustill:
Updated to add while loop condition tracking for E0908 flow analysis.

When a `c_` variable is checked in a while condition (e.g., `while (c_file != NULL)`), the variable is now considered NULL-checked inside the while body.

Also updated `valid-c-prefix-reassignment` test to require NULL check after reassignment (stricter behavior).

All 641 tests pass.

PR #278, opened 2026-01-21
feat: implement pass-by-value for small unmodified primitives

### LINKED ISSUE #269: Consider pass-by-value for small types instead of pointers
## Problem

C-Next generates pointer parameters for all value types, even small ones like `u8`, `u16`, `f32`. This requires wrapper functions when calling from C++ code that has values rather than pointers.

## Current behavior

C-Next source:
```cnx
public u8 lowByte(u16 value) {
    return value[0, 8];
}
```

Generated C++:
```cpp
uint8_t j1939_encode_lowByte(uint16_t* value) {
    return (((*value)) & 0xFFU);
}
```

Calling from C++ requires:
```cpp
uint16_t temp = j1939_encode_temp16bit(temperatureC);
msg.buf[0] = j1939_encode_lowByte(&temp);  // Must take address
```

## Workaround

Thin wrappers to preserve clean call sites:
```cpp
static inline bool isSpnEnabled(const AppConfig* cfg, uint16_t spn) {
    return spn_check_isSpnEnabled(cfg, &spn);
}
```

## Suggestion

Consider an option or heuristic to generate pass-by-value for small primitive types (≤ 8 bytes) where the parameter isn't modified. This would improve ergonomics when calling C-Next functions from existing C++ code.

## Why this matters

- Reduces boilerplate wrapper code
- More natural interop with existing C++ codebases
- Small types are efficiently passed by value on all modern ABIs

jlaustill:
## Summary

- Implements pass-by-value optimization for small unmodified primitive parameters (u8, i8, u16, i16, u32, i32, u64, i64, bool)
- Adds modification detection with fixed-point call graph analysis
- Updates code generation to omit pointer indirection for eligible parameters
- Ensures header/implementation signature consistency

### Analysis Phase
Added `analyzeParameterModifications()` in CodeGenerator that:
1. Walks AST to detect direct parameter modifications (assignment, compound ops, increment/decrement)
2. Uses fixed-point algorithm to propagate modification status through call graph
3. Marks parameters as pass-by-value only if they are unmodified AND small primitives

### Code Generation Changes
- `generateParameter()`: Omits `*` for pass-by-value params
- `generatePrimaryExpr()` and `doGenerateAssignmentTarget()`: Omit `(*...)` dereference
- Call site generation: Omits `&` address-of for pass-by-value arguments
- `HeaderGenerator`: Updated to use pass-by-value info for consistent signatures
- `Pipeline`: Wires pass-by-value info from CodeGenerator to HeaderGenerator

### Parameters That Still Pass by Pointer
- Modified directly or transitively (passed to function that modifies)
- Struct types (always by pointer per ADR-006)
- Float types (already pass by value per ADR-010)
- Array types (already pass naturally per C semantics)

## Test Plan

- [x] All 644 existing tests pass
- [x] New test at `tests/pass-by-value/small-unmodified.test.cnx` validates:
  - Unmodified parameters pass by value
  - Modified parameters still pass by pointer
  - Transitively modified parameters pass by pointer
  - Mixed parameter lists work correctly
- [x] Header/implementation signature consistency verified

Closes #269

### DIFF
--- a/src/codegen/CodeGenerator.ts
+++ b/src/codegen/CodeGenerator.ts
@@ -1816,6 +1816,40 @@ export default class CodeGenerator implements IOrchestrator {
       }
     }
 
+    // Issue #269: Handle switch statements - modifications can occur in any case
+    if (stmt.switchStatement()) {
+      const switchStmt = stmt.switchStatement()!;
+      // Check switch expression for calls
+      this.walkExpressionForCalls(funcName, paramSet, switchStmt.expression());
+      // Walk each case block
+      for (const caseCtx of switchStmt.switchCase()) {
+        this.walkBlockForModifications(
+          funcName,
+          [...paramSet],
+          caseCtx.block(),
+        );
+      }
+      // Walk default case if present
+      const defaultCase = switchStmt.defaultCase();
+      if (defaultCase) {
+        this.walkBlockForModifications(
+          funcName,
+          [...paramSet],
+          defaultCase.block(),
+        );
+      }
+    }
+
+    // ADR-050: Handle critical statements - recurse into the block
+    if (stmt.criticalStatement()) {
+      const criticalStmt = stmt.criticalStatement()!;
+      this.walkBlockForModifications(
+        funcName,
+        [...paramSet],
+        criticalStmt.block(),
+      );
+    }
+
     // Recurse into nested blocks
     if (stmt.block()) {
       this.walkBlockForModifications(funcName, [...paramSet], stmt.block()!);
--- a/tests/pass-by-value/small-unmodified.expected.c
+++ b/tests/pass-by-value/small-unmodified.expected.c
@@ -5,6 +5,7 @@
 
 #include <stdint.h>
 #include <stdbool.h>
+#include <cmsis_gcc.h>
 
 // test-execution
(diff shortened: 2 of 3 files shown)


jlaustill:
## Code Review Feedback Addressed

Based on the code review, the following issues were fixed:

### Critical Fix
- **Header/Implementation Signature Mismatch**: Updated `tests/pass-by-value/small-unmodified.test.h` to use correct pass-by-value signatures that match the implementation:
  - `getLowByte(uint16_t value)` instead of `getLowByte(uint16_t* value)`
  - `getHighByte(uint16_t value)` instead of `getHighByte(uint16_t* value)`
  - `addTwo(uint32_t a, uint32_t b)` instead of `addTwo(uint32_t* a, uint32_t* b)`
  - `invert(bool flag)` instead of `invert(bool* flag)`
  - `addAndIncrement(uint32_t* a, uint32_t b)` - correctly has `a` as pointer (modified) and `b` as value (unmodified)
  - Fixed `main` return type to `int`

### Important Fix
- **Removed unused `passByValue` field from TParameterInfo**: The field was added but never used since pass-by-value status is tracked in the `passByValueParams` map instead. Added documentation explaining this design choice.

All 644 tests pass after these fixes.

PR #285, opened 2026-01-21
fix: inline scope const values instead of generating local variables (#282)

### LINKED ISSUE #282: Scope constants generated as local variables instead of static or inline
## Version
v0.1.13

## Observation
When using `const` declarations at scope level, the transpiler generates them as local variables inside each function that uses them, rather than as `static const` or inlined values.

## Example

presets.cnx:
```cnx
scope presets {
    const f32 AEM_COEFF_A <- 1.485995686e-03;
    const f32 BOSCH_COEFF_A <- 1.40e-03;
    const f32 GM_COEFF_A <- 1.29e-03;

    public f32 ntcCoeffA(u8 preset) {
        if (preset = 0) { return this.AEM_COEFF_A; }
        if (preset = 1) { return this.BOSCH_COEFF_A; }
        if (preset = 2) { return this.GM_COEFF_A; }
        return 0.0;
    }
}
```

Generated presets.cpp:
```cpp
float presets_ntcCoeffA(uint8_t preset) {
    float presets_AEM_COEFF_A = 1.485995686e-03;   // All 3 declared
    float presets_BOSCH_COEFF_A = 1.40e-03;        // but only 1 is
    float presets_GM_COEFF_A = 1.29e-03;           // used per call
    if (preset == 0) { return presets_AEM_COEFF_A; }
    if (preset == 1) { return presets_BOSCH_COEFF_A; }
    if (preset == 2) { return presets_GM_COEFF_A; }
    return 0.0;
}
```

## cppcheck warning
```
style: The scope of the variable 'presets_AEM_COEFF_A' can be reduced. [variableScope]
```

## Suggestions

1. **Inline the values** directly in the return statements (preferred for small constants)
2. **Use `static const`** at file scope for larger data
3. **Declare inside branches** only where used

## Impact
- 12 additional cppcheck style warnings in our codebase
- Minor runtime inefficiency (unused variables initialized)
- Not a safety issue, just a code quality observation

jlaustill:
## Summary
- Fixes scope-level `const` declarations being incorrectly transpiled as local variables inside functions
- Private const values are now inlined directly at use sites
- Public const values are now emitted with proper `const` prefix at file scope
- Resolves cppcheck `variableScope` warnings for scope consts

Closes #282

## Test plan
- [x] Added new test: `tests/scope/scope-const-inline.test.cnx`
- [x] All 647 tests pass
- [x] Verified private const values are inlined (e.g., `return 255;` instead of local variable)
- [x] Verified public const values have `const` prefix at file scope

### DIFF
--- a/tests/scope/scope-const-inline.expected.c
+++ b/tests/scope/scope-const-inline.expected.c
@@ -42,6 +42,25 @@ uint8_t ConstTest_getCounter(void) {
     return ConstTest_counter;
 }
 
+uint8_t ConstTest_getHexFlags(void) {
+    return 0xAB;
+}
+
+uint8_t ConstTest_getBinMask(void) {
+    return 0b11001010;
+}
+
+uint32_t ConstTest_getHexAddr(void) {
+    return 0x40001000;
+}
+
+uint16_t ConstTest_multipleRefs(void) {
+    uint16_t a = 255;
+    uint16_t b = 256;
+    uint16_t c = 510;
+    return a + b + c;
+}
+
 int main(void) {
     uint8_t maxVal = ConstTest_getMaxValue();
     if (maxVal != 255) return 1;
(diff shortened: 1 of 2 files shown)


jlaustill:
## Additional Test Coverage

I've added edge case tests to improve coverage for the const inlining feature:

### New Test Cases

1. **Hex literal const values**
   ```cnx
   const u8 HEX_FLAGS <- 0xAB;
   const u32 HEX_ADDR <- 0x40001000;
   ```
   Verifies that hex format is preserved when inlining (e.g., `return 0xAB;`)

2. **Binary literal const values**
   ```cnx
   const u8 BIN_MASK <- 0b11001010;
   ```
   Verifies that binary format is preserved when inlining

3. **Multiple references to the same const**
   ```cnx
   public u16 multipleRefs() {
       u16 a <- this.MAX_VALUE;
       u16 b <- this.MAX_VALUE + 1;
       u16 c <- this.MAX_VALUE * 2;
       return a + b + c;
   }
   ```
   Verifies that each reference is independently inlined and expressions are correctly evaluated

### Results

All 647 tests pass ✅

The implementation correctly handles all edge cases - literal formats are preserved and multiple references work as expected.

PR #292, opened 2026-01-21
fix: add C++ template syntax parsing support (issue #291)

### LINKED ISSUE #291: C++ template syntax not parsed
## Description

C-Next fails to parse C++ template instantiation syntax. When attempting to declare a variable using a C++ template type, the parser fails with "no viable alternative".

## Reproduction

```cnx
#include "FlexCAN_T4.h"

scope test_can {
    FlexCAN_T4<CAN1, RX_SIZE_256, TX_SIZE_16> canBus;
    
    public void init() {
        this.canBus.begin();
    }
}
```

## Error

```
Error: no viable alternative at input 'FlexCAN_T4<'
```

## Expected Behavior

C-Next should parse C++14 template syntax and generate equivalent C++ code. This would allow converting files like `J1939Bus.cpp` which use templated types.

## Context

This is blocking conversion of embedded code that uses templated libraries like FlexCAN_T4, Adafruit sensor libraries, etc. Currently these files must remain as handwritten C++ because the template declarations can't be expressed in C-Next.

## Environment

- C-Next version: 0.1.15
- Platform: Linux

jlaustill:
## Summary

- Adds grammar rules to parse C++ template instantiation syntax like `FlexCAN_T4<CAN1, RX_SIZE_256, TX_SIZE_16>`
- Enables interop with templated C++ libraries in embedded systems (FlexCAN_T4, Adafruit sensors, etc.)
- Updates test infrastructure to detect templates and use g++ compiler

## Changes

**Grammar (`grammar/CNext.g4`):**
- Added `templateType` rule: `IDENTIFIER '<' templateArgumentList '>'`
- Added `templateArgumentList` and `templateArgument` rules
- Supports nested templates, primitive types, identifiers, and integer literals

**Test Infrastructure (`scripts/test-utils.ts`):**
- Updated `requiresCpp14` to detect template syntax and use g++ instead of gcc

**Tests:**
- `simple-template.test.cnx` - Multiple and integer parameters
- `single-param-template.test.cnx` - Single type parameter
- `external-template-interop.test.cnx` - Templates with C-Next scopes

## Test plan

- [x] All 653 tests pass
- [x] New template tests compile with g++
- [x] Template syntax is passed through unchanged to C++ output

Closes #291


jlaustill:
## Review Feedback Addressed

**Nested template test added** - New test case `nested-template.test.cnx` covers `Container<Pair<Element, Element> >` and similar patterns.

**Regex detection** - Verified the current regex still detects nested templates (matches inner template which is sufficient for g++ detection).

**Note on `>>` ambiguity:** Nested templates require a space before `>>` in C-Next source (`Container<Pair<T, U> >`) due to lexer tokenization. The transpiler outputs valid C++11 syntax without the space. This is the same limitation as pre-C++11 C++ compilers.

All 654 tests pass.

Issue #232, opened 2026-01-18
Scope variables using static storage makes functions non-thread-safe


jlaustill:
## Description

Scope-level variables are transpiled to `static` variables at file scope, which makes functions non-reentrant and not thread-safe.

## Generated Code

```cpp
/* Scope: crc32 */
static uint32_t crc32_offset = 0;
static uint8_t crc32_length = 0;

uint32_t crc32_calculateChecksum(const AppConfig* config) {
    crc32_offset = 0;  // Reset at start
    // ... uses crc32_offset and crc32_length throughout
}
```

## Problem

If `crc32_calculateChecksum()` is called from multiple threads simultaneously (or re-entered via interrupt), the shared static variables will be corrupted.

## Source CNX

```cnx
scope crc32 {
    u32 offset <- 0;
    u8 length <- 0;
    
    public u32 calculateChecksum(const AppConfig config) {
        this.offset <- 0;
        // ...
    }
}
```

## Possible Solutions

1. **Use function-local variables** instead of static when the variable is only used within one function
2. **Document the thread-safety implications** of scope variables
3. **Provide an option** like `scope crc32 [reentrant]` to generate thread-safe code
4. **Use thread-local storage** (`_Thread_local` / `thread_local`) as an option

## Impact

- For single-threaded embedded systems: Low risk, but still a footgun
- For RTOS or multi-core systems: Can cause subtle, hard-to-debug corruption
- For interrupt-driven code: Re-entrancy issues if ISR calls the function

## Environment

- c-next version: 0.1.9

jlaustill:
Starting work on this issue. The plan is to implement usage analysis to detect single-function scope variables and make them function-local instead of file-scope static.

**Approach:**
1. Add tracking in SymbolCollector for which scope variables are used in which functions
2. Variables used in only ONE function → local variable (reentrant)
3. Variables used in MULTIPLE functions → file-scope static (intentionally shared state)

This preserves the shared-state use case while fixing the reentrancy bug for single-function variables.

PR #319, opened 2026-01-22
fix: scope variables now persist across calls (Issue #313)

### LINKED ISSUE #313: Scope-level variables incorrectly generated as local variables instead of static
## Bug Description

Scope-level variables in C-Next are inconsistently generated - some become static module-level variables (correct), while others become local variables inside functions (incorrect).

## Reproduction

Given this C-Next code:

```cnx
scope Ossm {
    AppData appData;
    AppConfig appConfig;
    IntervalTimer sensorTimer;
    volatile bool sensorUpdateReady <- false;
    elapsedMillis halfSecondMillis;
    elapsedMillis oneSecondMillis;
    
    public void sendJ1939Messages() {
        if (this.halfSecondMillis >= 500) {
            // ...
            this.halfSecondMillis <- 0;
        }
    }
    
    public void setup() {
        // uses this.appConfig, this.sensorTimer
    }
}
```

## Expected Output

All scope-level variables should be static at module scope:

```c
static AppData Ossm_appData = {};
static AppConfig Ossm_appConfig = {};
static IntervalTimer Ossm_sensorTimer = {};
static bool Ossm_sensorUpdateReady = false;
static elapsedMillis Ossm_halfSecondMillis = {};
static elapsedMillis Ossm_oneSecondMillis = {};

void Ossm_sendJ1939Messages(void) {
    if (Ossm_halfSecondMillis >= 500) {
        // ...
        Ossm_halfSecondMillis = 0;
    }
}
```

## Actual Output

Some variables are static (correct), others are local inside functions (incorrect):

```c
static AppData Ossm_appData = {};
static bool Ossm_sensorUpdateReady = false;

void Ossm_sendJ1939Messages(void) {
    elapsedMillis Ossm_halfSecondMillis = {};  // BUG: local, not static!
    elapsedMillis Ossm_oneSecondMillis = {};   // BUG: local, not static!
    if (Ossm_halfSecondMillis >= 500) {        // Always false - just initialized!
        // ...
    }
}

void Ossm_setup(void) {
    struct AppConfig Ossm_appConfig = {};      // BUG: local, not static!
    IntervalTimer Ossm_sensorTimer = {};       // BUG: local, not static!
}
```

## Impact

- Variables lose state between function calls
- Timers and elapsed time tracking completely broken
- Configuration lost after setup() returns

## C-Next Version

v0.1.15

jlaustill:
## Summary
- Removed the Issue #232 "single-function variable optimization" that violated ADR-016
- Scope variables used by only one function were incorrectly converted to function-local variables, causing them to reset on each call
- All scope variables now correctly emit at file scope as `static` declarations, persisting across function calls

## Changes
- Removed `ISingleFunctionVar` interface and `localVarsForFunction` tracking from `ScopeGenerator.ts`
- Removed single-function detection logic that skipped file-scope emission
- Removed local variable injection code that injected variables inside function bodies
- Added new test `scope-single-function-state.test.cnx` to verify persistence
- Updated 25 expected output files to reflect correct behavior

## Test Plan
- [x] New test validates scope variables persist across function calls
- [x] All 661 tests pass
- [x] Manual execution verification: test returns exit code 0 (was returning 2 before fix)

Closes #313

### DIFF
--- a/tests/conditions/condition-no-function-call.expected.c
+++ b/tests/conditions/condition-no-function-call.expected.c
@@ -23,7 +23,7 @@ static inline uint32_t cnx_clamp_sub_u32(uint32_t a, uint64_t b) {
     return result;
 }
 
-/* test-execution */
+// test-execution
 // Issue #254: Valid conditions without function calls
 // These should compile and run successfully
 uint32_t getValue(void) {
--- a/tests/conditions/condition-no-function-call.test.cnx
+++ b/tests/conditions/condition-no-function-call.test.cnx
@@ -1,4 +1,4 @@
-/* test-execution */
+// test-execution
 // Issue #254: Valid conditions without function calls
 // These should compile and run successfully
 
--- a/tests/const-params/auto-const-pointer.expected.c
+++ b/tests/const-params/auto-const-pointer.expected.c
@@ -5,7 +5,7 @@
 
 #include <stdint.h>
 
-/* test-execution */
+// test-execution
 // Tests: Issue #268 - automatic const for unmodified pointer parameters
 // Read-only parameter - should get const
 void readOnly(uint32_t value) {
--- a/tests/const-params/auto-const-pointer.test.cnx
+++ b/tests/const-params/auto-const-pointer.test.cnx
@@ -1,4 +1,4 @@
-/* test-execution */
+// test-execution
 // Tests: Issue #268 - automatic const for unmodified pointer parameters
 
 // Read-only parameter - should get const
--- a/tests/const-params/pass-through-modification.expected.c
+++ b/tests/const-params/pass-through-modification.expected.c
@@ -5,7 +5,7 @@
 
 #include <stdint.h>
 
-/* test-execution */
+// test-execution
 // Tests: Issue #268 - pass-through modification tracking for auto-const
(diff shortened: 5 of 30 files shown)


jlaustill:
## Historical Context: Why Issue #232 Optimization Existed

Issue #232 identified a **real concern**: scope variables using static storage makes functions non-reentrant and not thread-safe. The Issue proposed four solutions:

1. Use function-local variables for single-function scope vars ← **This was implemented (incorrectly)**
2. Document the thread-safety implications
3. Provide a `scope [reentrant]` option
4. Use thread-local storage

**Why solution #1 was wrong**: ADR-016 *intentionally* defines scope variables as having static-like persistence. Silently converting them to function-local variables changes observable behavior - a counter that should increment across calls now always returns 1.

**The correct approach** would be:
- **Option 2**: Document that scope variables are intentionally static (like C static vars)
- **Option 3**: Add explicit syntax like `scope Foo [reentrant]` for users who want thread-safe, reentrant behavior

If users want reentrant functions, they should declare variables inside the function in C-Next, not as scope members.

This PR removes the incorrect optimization. A future issue could explore adding explicit reentrant scope support if needed.

PR #598, opened 2026-02-02
fix: array indexing on parameter generates array access (#579)

### LINKED ISSUE #579: Array indexing on parameter generates bit manipulation instead of array access
## Summary

When a function parameter is used with array indexing (`buf[i] <- value`), but declared without explicit array syntax (`u8 buf` instead of `u8 buf[8]`), the transpiler generates incorrect code that treats array indexing as bit manipulation.

## Version

cnext v0.1.51

## Minimal Reproducer

```c-next
scope Test {
    void fillBuffer(u8 buf) {
        for (u8 i <- 0; i < 8; i +<- 1) {
            buf[i] <- 0xFF;  // Using buf as an array
        }
    }

    public void run() {
        u8 buffer[8];
        this.fillBuffer(buffer);
    }
}
```

## Generated (incorrect)

```c
static void Test_fillBuffer(uint8_t* buf) {
    for (uint8_t i = 0; i < 8; i += 1) {
        // BUG: Treats buf[i] <- 0xFF as bit manipulation on the pointer
        buf = (buf & ~(1 << i)) | ((0xFF ? 1 : 0) << i);
    }
}
```

## Expected

```c
static void Test_fillBuffer(uint8_t* buf) {
    for (uint8_t i = 0; i < 8; i += 1) {
        buf[i] = 0xFF;
    }
}
```

## Compilation Error

```
error: invalid operands to binary & (have 'uint8_t *' and 'int')
```

## Analysis

The transpiler confuses array indexing (`buf[i]`) with bit extraction syntax (`value[start, width]`). When the index is a variable, it should recognize this as array access, not bit manipulation.

## Workaround

Explicitly declare the parameter with array syntax: `void fillBuffer(u8 buf[8])`

jlaustill:
## Summary

Fixes #579 - When a function parameter is declared as `u8 buf` (without explicit array syntax) but used with array indexing like `buf[i]`, the transpiler now correctly generates array access instead of bit manipulation.

**Root causes fixed:**
1. Non-array parameters used with subscript access now become pointers (not pass-by-value)
2. Expression path now uses array access for parameter subscripts
3. Fixed `currentStructType` initialization for primitive array bit access

**Changes:**
- Add `isParameter` field to `TTypeInfo` to track function parameters
- Create shared `SubscriptClassifier` utility for consistent array/bit classification
- Track subscript access on parameters to ensure they become pointers in C
- Update expression path to use array access for parameter subscripts
- Fix primitive array bit access (e.g., `matrix[i][j][bit]`)

## Test plan

- [x] New test file `tests/params/param-array-indexing.test.cnx` covers all 3 root causes
- [x] Test 1-6: Parameter array indexing (read/write with variable/constant index)
- [x] Test 7: Bit range on parameter stays as bit extraction (no over-correction)
- [x] Test 8: Local array bit access after consuming dimensions
- [x] All 881 integration tests pass
- [x] All 1607 unit tests pass

### DIFF
--- a/src/transpiler/output/codegen/CodeGenerator.ts
+++ b/src/transpiler/output/codegen/CodeGenerator.ts
@@ -2520,53 +2520,20 @@ export default class CodeGenerator implements IOrchestrator {
     const ternary = expr.ternaryExpression();
     if (ternary) {
       for (const orExpr of ternary.orExpression()) {
-        this.walkOrExpressionForSubscriptAccess(funcName, paramSet, orExpr);
+        this.walkOrExpression(orExpr, (unaryExpr) => {
+          this.handleSubscriptAccess(funcName, paramSet, unaryExpr);
+        });
       }
     }
   }
 
   /**
-   * Issue #579: Walk an orExpression tree for subscript access.
-   */
-  private walkOrExpressionForSubscriptAccess(
-    funcName: string,
-    paramSet: Set<string>,
-    orExpr: Parser.OrExpressionContext,
-  ): void {
-    for (const andExpr of orExpr.andExpression()) {
-      for (const eqExpr of andExpr.equalityExpression()) {
-        for (const relExpr of eqExpr.relationalExpression()) {
-          for (const bitOrExpr of relExpr.bitwiseOrExpression()) {
-            for (const bitXorExpr of bitOrExpr.bitwiseXorExpression()) {
-              for (const bitAndExpr of bitXorExpr.bitwiseAndExpression()) {
-                for (const shiftExpr of bitAndExpr.shiftExpression()) {
-                  for (const addExpr of shiftExpr.additiveExpression()) {
-                    for (const mulExpr of addExpr.multiplicativeExpression()) {
-                      for (const unaryExpr of mulExpr.unaryExpression()) {
-                        this.walkUnaryExpressionForSubscriptAccess(
-                          funcName,
-                          paramSet,
-                          unaryExpr,
-                        );
-                      }
-                    }
-                  }
-                }
-              }
-            }
-          }
-        }
-      }
-    }
-  }
-
-  /**
-   * Issue #579: Walk a unary expression for subscript access.
+   * Issue #579: Handle subscript access on a unary expression.
    * Only tracks single-index subscript access (which could be array access).
    * Two-index subscript (e.g., value[start, width]) is always bit extraction,
    * so it doesn't require the parameter to become a pointer.
    */
-  private walkUnaryExpressionForSubscriptAccess(
+  private handleSubscriptAccess(
     funcName: string,
     paramSet: Set<string>,
     unaryExpr: Parser.UnaryExpressionContext,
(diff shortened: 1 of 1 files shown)


jlaustill:
## PR #598 Review: Array Indexing on Parameters

**Overall: Request Changes** - The fix has merit but introduces regressions in two test files.

### Summary

| Area | Changes | Impact |
|------|---------|--------|
| `CodeGenerator.ts` | +162/-16 | Subscript tracking, expression path fix |
| `SubscriptClassifier.ts` | +96 new | Shared classification logic |
| `TSubscriptKind.ts` | +13 new | Type discriminant |
| `TTypeInfo.ts` | +1 | Add `isParameter` field |
| `AssignmentClassifier.ts` | +29/-29 | Use shared classifier |
| New test | +245 | `param-array-indexing.test.cnx` |
| Modified tests | +62/-16 | Two existing tests changed |

### What's Good

1. **Shared `SubscriptClassifier`** - Centralizes array vs bit classification logic
2. **`isParameter` tracking** - Enables correct detection of pointer parameters
3. **Comprehensive new test** - `param-array-indexing.test.cnx` covers the main fix with 13 validations
4. **Subscript access tracking** - `subscriptAccessedParameters` correctly identifies single-index subscript

### Issues Found

#### 1. `small-unmodified.test.c` - Inconsistent signature/body

The source has bit range extraction:
```c-next
u8 getLowByte(u16 value) {
    return value[0, 8];  // TWO-index subscript = bit extraction
}
```

The new expected output:
```c
uint8_t getLowByte(const uint16_t* value) {  // Pointer!
    return ((value) & 0xFFU);  // But no dereference!
}
```

**Problem**: If `value` is now a pointer, `(value) & 0xFFU` operates on the pointer address, not the value. This would compile but produce garbage results.

The `value[0, 8]` is TWO-index subscript (bit range), which should NOT trigger pointer conversion. Only SINGLE-index subscript should cause this per the issue description.

#### 2. `const-expression-chain.test.c` - Bit access regression

The source has:
```c-next
u8 matrix[4][4];
bool bit <- matrix[ROW][COL][FIELD_BIT];  // Bit access on array element
```

Old expected (correct):
```c
bool bit = ((matrix[ROW][COL] >> FIELD_BIT) & 1);
```

New expected (broken):
```c
bool bit = matrix[ROW][COL][FIELD_BIT];
```

**Problem**: `matrix[ROW][COL]` is `uint8_t` (a scalar). You cannot subscript a scalar in C - this is invalid code that would either fail to compile or read from garbage memory.

The new test `param-array-indexing.test.cnx` correctly handles this same pattern:
```c
return ((matrix[row][col] >> bit) & 1);  // Correct!
```

So the bug appears to be that `SubscriptClassifier` is checking the primary identifier's type (`matrix` = array) instead of the intermediate result's type (`matrix[ROW][COL]` = scalar).

#### 3. Unrelated changes in `const-expression-chain.test.c`

```diff
-const uint32_t INDEX_0 = 0;
+extern const uint32_t INDEX_0 = 0;
```

Adding `extern` to const definitions wasn't mentioned in the PR description.

### Suggested Fixes

1. **For issue #1**: The `subscriptAccessedParameters` tracking should correctly exclude two-index subscript. Verify that `value[0, 8]` doesn't trigger the tracking logic, or there's another bug causing the signature change.

2. **For issue #2**: When classifying subscripts after array dimensions are consumed, use the intermediate type (scalar after indexing), not the original array type.

3. **Revert the const-expression-chain.test.c expected output** to the correct bit extraction pattern.

jlaustill:
Thanks for the thorough review! You're correct - the committed `.test.c` files were stale from an intermediate development state before I fixed the two-index subscript exclusion.

I've pushed a fix commit (80d2ae3d) that updates the generated test files with correct output:

**Issue #1 - `small-unmodified.test.c`**: Now correctly shows `uint16_t value` (pass-by-value) instead of `const uint16_t* value` (pointer). The two-index subscript `value[0, 8]` correctly stays as bit extraction.

**Issue #2 - `const-expression-chain.test.c`**: Now correctly shows `((matrix[ROW][COL] >> FIELD_BIT) & 1)` for bit access instead of the invalid `matrix[ROW][COL][FIELD_BIT]`.

**Issue #3 - `extern` on const**: This appears to be a pre-existing difference between the `.expected.c` and `.test.c` files in the repo. The expected file already had `extern`, so this isn't a regression from my changes.

The core fix is working correctly - the stale files were just an artifact of my development process where I staged files before the final fix was complete.

PR #606, opened 2026-02-02
fix(transpiler): use DependencyGraph for file ordering (#580)

### LINKED ISSUE #580: Incorrect const inference when parameter is only passed to mutating function
## Summary

In certain cases, the transpiler incorrectly marks a struct parameter as `const` even when it's passed to a function that mutates it. This causes compilation errors when the called function expects a non-const reference.

## Version

cnext v0.1.51

## Observed Behavior

In a multi-file project, two functions with identical parameter types generate different constness:

### Source (both take `AppConfig config`):

```c-next
// Gets incorrectly marked const
void handleReset(AppConfig config) {
    Serial.println("Resetting to defaults...");
    global.CommandHandler.reset(config);  // reset() takes non-const, mutates config
    Serial.println("OK");
}

// Correctly marked non-const
void handleNtcPreset(AppConfig config) {
    if (global.parsed.count < 3) {
        Serial.println("ERR");
        return;
    }
    u8 input <- global.parsed.data[1];
    u8 preset <- global.parsed.data[2];
    u8 errorCode <- global.CommandHandler.applyNtcPreset(config, input, preset);
}
```

### Generated (C++ mode):

```cpp
// BUG: const despite calling non-const function
static void SerialCommandHandler_handleReset(const AppConfig& config) {
    CommandHandler_reset(config);  // ERROR: can't bind const to non-const ref
}

// CORRECT: non-const
static void SerialCommandHandler_handleNtcPreset(AppConfig& config) {
    // ...
    CommandHandler_applyNtcPreset(config, input, preset);
}
```

## Compilation Error

```
error: binding reference of type 'AppConfig&' to 'const AppConfig' discards qualifiers
```

## Pattern

Functions incorrectly marked const appear to:
- Have minimal control flow (no early returns before the mutating call)
- Have no local variable assignments before the mutating call
- Only pass the parameter through to another function

## Minimal Reproducer

Could not isolate in a single file. Bug manifests in the OSSM project with cnext v0.1.51:
```bash
cd ossm && cnext src/ && pio run
```

The affected file is `src/Domain/SerialCommandHandler.cnx`, functions `handleReset` and `handleSave`.

jlaustill:
## Summary
- Replace conditional reversal heuristic with proper topological sorting using `DependencyGraph`
- Fix incorrect const inference when C++ mode is detected from headers (not --cpp flag)
- Add CLI test for the C++ header detection scenario

## Root Cause
The file ordering reversal was conditional on `cppDetected` being true at line 160, but `cppDetected` could be set later during Stage 2 (header collection). Projects like OSSM that don't use `--cpp` but have C++ headers would process files in the wrong order, causing transitive const inference to fail.

## Changes
- Use `DependencyGraph` in `discoverSources()` to track file dependencies as includes are resolved
- Return files sorted topologically (dependencies first) via `getSortedFiles()`
- Remove the conditional reversal heuristic entirely
- Add CLI test that verifies C++ detection from headers triggers correct const inference

## Test plan
- [x] All 885 integration tests pass
- [x] All 1723 unit tests pass
- [x] All 28 CLI tests pass (including new issue #580 test)
- [x] Linting passes

Closes #580

### DIFF
--- a/src/cli/CleanCommand.ts
+++ b/src/cli/CleanCommand.ts
@@ -3,9 +3,10 @@
  * Deletes generated files (.c, .cpp, .h, .hpp) that have matching .cnx sources
  */
 
-import { basename, join, relative, resolve } from "node:path";
-import { existsSync, statSync, unlinkSync } from "node:fs";
+import { basename, join, resolve } from "node:path";
+import { unlinkSync } from "node:fs";
 import InputExpansion from "../transpiler/data/InputExpansion";
+import PathResolver from "../transpiler/data/PathResolver";
 
 /**
  * Command to clean generated output files
@@ -48,6 +49,13 @@ class CleanCommand {
       ? resolve(headerOutDir)
       : resolvedOutDir;
 
+    // Issue #586: Use PathResolver for consistent path calculation
+    const pathResolver = new PathResolver({
+      inputs,
+      outDir,
+      headerOutDir,
+    });
+
     let deletedCount = 0;
 
     // For each .cnx file, calculate and delete generated files
@@ -55,7 +63,7 @@ class CleanCommand {
       const baseName = basename(cnxFile).replace(/\.cnx$|\.cnext$/, "");
 
       // Calculate relative path from input directories
-      const relativePath = this.getRelativePath(cnxFile, inputs);
+      const relativePath = pathResolver.getRelativePathFromInputs(cnxFile);
 
       // Code files (.c and .cpp) go to outDir
       const codeExtensions = [".c", ".cpp"];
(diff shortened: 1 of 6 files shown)


jlaustill (commented):
## Code Review: PR #606 - Use DependencyGraph for file ordering (#580)

### Overall Assessment: ✅ Approve

This is an excellent fix that replaces a flawed heuristic with a proper solution using existing infrastructure.

### Root Cause Analysis

The bug was subtle:
```typescript
// OLD CODE (lines 158-161)
if (this.cppDetected) {
  cnextFiles.reverse();
}
```

**Problem**: `cppDetected` could be `false` at this point but set to `true` later during Stage 2 (header collection). Projects using C++ headers without `--cpp` flag would skip the reversal, causing incorrect file ordering and broken transitive const inference.

### Solution Quality

The fix is architecturally sound:

1. **Leverages existing `DependencyGraph`** - Uses proper topological sorting via `@n1ru4l/toposort` instead of a brittle heuristic
2. **Tracks dependencies during discovery** - Each include relationship is recorded: `depGraph.addDependency(cnxPath, includePath)`
3. **Deterministic ordering** - Files are always sorted with dependencies first, regardless of when C++ mode is detected
4. **Handles cycles gracefully** - Falls back to arbitrary order with a warning (existing DependencyGraph behavior)

### Code Changes

**Transpiler.ts changes are clean:**
- Adds `fileByPath` map to track `IDiscoveredFile` objects by resolved path
- Builds dependency graph as includes are resolved
- Returns topologically sorted files

```typescript
// Track dependency: cnxFile depends on cnxInclude
depGraph.addDependency(cnxPath, includePath);
```

**CLI test is comprehensive:**
- Reproduces the exact bug scenario (C++ detected from headers, not `--cpp` flag)
- Verifies transitive const inference works correctly
- Tests both Handler and Modifier output

### No Issues Found

The implementation is correct. A few observations (not blockers):

1. **Path normalization** - Uses `resolve()` consistently for paths, which is good
2. **Warning propagation** - `depGraph.getWarnings()` is properly pushed to `this.warnings`
3. **Test isolation** - Test creates temp directory and cleans up properly

### Minor Suggestion (optional)

The comment at line 153-155 could be simplified:
```typescript
// Issue #580: Files are sorted by DependencyGraph in discoverSources()
// using topological sorting (dependencies first).
```

### Summary

| Aspect | Rating |
|--------|--------|
| Bug fix correctness | ✅ Excellent |
| Solution approach | ✅ Proper (uses existing infra) |
| Test coverage | ✅ Comprehensive CLI test |
| Code quality | ✅ Clean |

This is a textbook example of replacing a heuristic with a proper algorithmic solution. Ready to merge.

Issue #355, opened 2026-01-22
Bug: Array member via pointer generates invalid static_cast in C++ mode


jlaustill:
## Description

When accessing an array member through a pointer-converted struct parameter in C++ mode, the transpiler generates an invalid `static_cast` that tries to cast an array to a scalar type.

## Minimal Reproduction

**CanMessage.h:**
```c
#ifndef CAN_MESSAGE_H
#define CAN_MESSAGE_H

#include <stdint.h>

typedef struct CAN_message_t {
    uint32_t id;
    uint8_t buf[8];
    uint8_t len;
} CAN_message_t;

#endif
```

**test.cnx:**
```c-next
#include "CanMessage.h"

scope Test {
    public void processBuffer(const u8 data[8], u8 len) {
    }
    
    void handleMessage(const CAN_message_t msg) {
        this.processBuffer(msg.buf, msg.len);
    }
}
```

**Command:**
```bash
cnext test.cnx --cpp
```

## Actual Output (v0.1.26)

```cpp
static void Test_handleMessage(const CAN_message_t* msg) {
    uint8_t _cnx_tmp_0 = static_cast<uint8_t>(msg->buf);  // INVALID!
    Test_processBuffer(&_cnx_tmp_0, msg->len);
}
```

**Compile error:**
```
error: invalid 'static_cast' from type 'const uint8_t [8]' to type 'uint8_t'
```

## Expected Output

```cpp
static void Test_handleMessage(const CAN_message_t* msg) {
    Test_processBuffer(msg->buf, msg->len);  // Array decays to pointer
}
```

## Notes

- This is specific to **C++ mode** (`--cpp` flag or when C++ features are detected)
- In C mode, the same code generates `&msg->buf` which compiles but has wrong semantics
- The transpiler correctly converts `const CAN_message_t msg` to `const CAN_message_t* msg`
- But it then incorrectly handles array member access through that pointer
- Related to closed issues #342 and #347 which may have been incompletely fixed

jlaustill:
**Additional Context from #342:**

Looking at the fix comments on #342, the root cause fix was:
> Modified `CppSymbolCollector.collectSimpleDeclaration()` to detect **anonymous class specifiers** in typedefs

But `CAN_message_t` uses a **named** struct pattern:
```c
typedef struct CAN_message_t { ... } CAN_message_t;  // Named struct
```

Not the anonymous pattern that was fixed:
```c
typedef struct { ... } Name;  // Anonymous struct
```

[...]

Issue #331, opened 2026-01-22
Overlapping include paths cause duplicate symbol detection for same file


jlaustill:
## C-Next Version
v0.1.22

## Summary
When include paths overlap (e.g., both `src/Display` and `src` are in the include path), the transpiler discovers the same `.cnx` file multiple times via different path resolutions and incorrectly flags it as a symbol conflict with itself.

### Directory structure:
```
/tmp/test/
└── src/
    └── Display/
        ├── Utils.cnx
        └── App.cnx
```

### Utils.cnx:
```c-next
scope Utils {
    public u8 add(u8 x) {
        return x + 1;
    }
}
```

### App.cnx:
```c-next
#include <Utils.cnx>

scope App {
    public u8 test() {
        return global.Utils.add(5);
    }
}
```

### Command (with overlapping include paths):
```bash
cd /tmp/test
cnext --include src/Display --include src src/Display/App.cnx
```

### Error Output:
```
Include paths:
  /tmp/test/src/Display
  src/Display
  src
Conflict: Symbol conflict: 'Utils' is defined multiple times in C-Next:
  /tmp/test/src/Display/Utils.cnx:1
  /tmp/test/src/Display/Utils.cnx:1
  /tmp/test/src/Display/Utils.cnx:1
```

Note: The **same file and line** appears **3 times** in the conflict message.

## Expected Behavior
The transpiler should canonicalize file paths before checking for duplicates. If the same physical file is discovered via multiple include paths, it should be processed only once.

## Root Cause Analysis
With include paths `[src/Display, src]`, the file `Utils.cnx` can be found as:
1. `src/Display/Utils.cnx` (direct from first path)
2. `Display/Utils.cnx` (via `src` path)

The transpiler appears to be adding symbols without first checking if the resolved file path was already processed.

## Impact
This blocks any project with a typical directory structure where:
- `src/SubDir` is an include path (for local includes)
- `src` is also an include path (for cross-module includes like `#include <SubDir/File.cnx>`)

## Suggested Fix
Before processing a discovered file, canonicalize its path (resolve to absolute path) and check if it was already processed.

## Workaround
None known - the overlapping include paths are automatically inferred by the transpiler based on the input file location.

jlaustill:
🔍 **Starting investigation**

Beginning systematic debugging of this issue:
1. Reproducing the bug locally
2. Tracing the root cause through the codebase
3. Writing tests to verify the bug exists
4. Implementing a fix

Will update with findings.

PR #620, opened 2026-02-02
refactor: migrate to established libraries for reduced maintenance

jlaustill:
## Summary

- **Terminal colors**: Migrated from custom ANSI escape codes to `chalk`
  - Created shared `scripts/colors.ts` helper
  - Updated 5 script files to use chalk's function-based API

- **ConfigLoader**: Migrated from custom file search to `cosmiconfig`
  - Simplified implementation (~52 → ~42 lines)
  - Proper parent directory search with `stopDir: "/"`

- **FileDiscovery**: Migrated from custom recursive scan to `fast-glob`
  - Converted RegExp excludes to glob ignore patterns
  - Preserved Issue #331 deduplication for overlapping directories
  - Preserved Issue #355 `.pio/build` exclusion handling

- **CacheManager**: Migrated from custom JSON I/O to `flat-cache` v6
  - Version invalidation clears old cache before creating new
  - Updated tests for flat-cache's file naming convention (no `.json` extension)

## Dependencies Added
```
chalk, cosmiconfig, fast-glob, flat-cache
```

## Test plan
- [x] All 888 integration tests pass
- [x] All 1967 unit tests pass
- [x] All 28 CLI tests pass
- [x] ConfigLoader tests verify parent directory search
- [x] CacheManager tests verify version invalidation
- [x] FileDiscovery handles overlapping directories correctly

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,7 +11,6 @@
       "dependencies": {
         "@n1ru4l/toposort": "^0.0.1",
         "antlr4ng": "^3.0.16",
-        "chalk": "^5.6.2",
         "cosmiconfig": "^9.0.0",
         "fast-glob": "^3.3.3",
         "flat-cache": "^6.1.20",
@@ -27,6 +26,7 @@
         "@types/yargs": "^17.0.35",
         "@vitest/coverage-v8": "^4.0.18",
         "antlr4ng-cli": "^2.0.0",
+        "chalk": "^5.6.2",
         "dependency-cruiser": "^17.3.7",
         "husky": "^9.1.7",
         "knip": "^5.82.1",
@@ -1828,6 +1828,7 @@
       "version": "5.6.2",
       "resolved": "https://registry.npmjs.org/chalk/-/chalk-5.6.2.tgz",
       "integrity": "sha512-7NzBL0rN6fMUW+f7A6Io4h40qQlG+xGmtMxfbnH/K7TAtt8JQWVQK+6g0UXKMeVJoyV5EkkNsErQ8pVD3bLHbA==",
+      "dev": true,
       "license": "MIT",
       "engines": {
         "node": "^12.17.0 || ^14.13 || >=16.0.0"
--- a/package.json
+++ b/package.json
@@ -79,6 +79,7 @@
     "@types/yargs": "^17.0.35",
     "@vitest/coverage-v8": "^4.0.18",
     "antlr4ng-cli": "^2.0.0",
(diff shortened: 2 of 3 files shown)


jlaustill:
## Code Review

**Overall Assessment: Looks good with a few issues to address.**

### Issues Found

**1. `chalk` should be a devDependency, not dependency**

`chalk` is only used in `scripts/` files which are development/testing tools:
- `scripts/colors.ts`, `scripts/test-cli.js`

These don't run in production when users invoke `cnext`. The other three libraries (`cosmiconfig`, `fast-glob`, `flat-cache`) are correctly in dependencies since they're used in runtime code.

**2. CacheManager migration test no longer tests actual migration**

The test `"should migrate old format cache entries"` (line 622) was modified but now just sets a new entry and reads it back:
```typescript
// Use the new manager API to set an entry, then manually modify it
cacheManager.setSymbols(...);
await cacheManager.flush();
// Reload with new manager - entry should be accessible
```
This doesn't test migration from the old mtime-based format to the new cacheKey format. The `migrateOldEntries()` method exists but has no test coverage now.

**3. FileDiscovery's `regexToGlob` has limited coverage**

The method only handles 4 specific patterns with a naive fallback:
```typescript
return `**/*${src.replace(/\\/g, "")}*/**`;
```
This works because the codebase only uses those 4 patterns, but could break if someone passes a different pattern. Consider adding a warning or documenting this limitation.

### Minor Issues

**4. `Colors.reset` method is misleading**
```typescript
static reset(text: string): string {
  return text; // Does nothing
}
```
This differs from the old `"\x1b[0m"` behavior. While chalk handles resets automatically, having a no-op method named `reset` is confusing. Consider removing it if unused, or renaming to `identity`/`none`.

**5. Orphaned type file?**

`ICacheSymbols` import was removed from CacheManager.ts. Verify `src/transpiler/types/ICacheSymbols.ts` isn't now orphaned dead code.

### What Looks Good

- Issue references preserved (#331, #355) showing behavior was maintained
- Same `searchPlaces` order in ConfigLoader as before
- Error handling preserved (`console.warn` for unreadable directories)
- Proper deduplication logic maintained for overlapping directories
- Test updates for flat-cache's filename convention (no `.json` extension)
- Removed ~165 lines of manual implementations

### Suggested Actions

1. Move `chalk` from `dependencies` to `devDependencies` in package.json
2. Either restore meaningful migration testing or add a comment explaining flat-cache handles format internally
3. Run `npx knip` to check for orphaned files after the refactor

PR #641, opened 2026-02-03
fix(transpiler): consolidate dual code paths (Issue #634)

### LINKED ISSUE #634: refactor: consolidate Transpiler.run() and transpileSource() dual code paths
## Summary

The Transpiler class has two entry points that must be kept manually synchronized, which is a code smell and source of potential bugs:

1. **`Transpiler.run()`** — CLI entry point (multi-file processing)
2. **`Transpiler.transpileSource()`** — Test framework entry point (single-file)

## Problem

When adding new features, developers must ensure both code paths are updated correctly. This has caused bugs in the past (see CLAUDE.md warnings) and requires explicit documentation to prevent regressions.

From CLAUDE.md:
> When adding features involving cross-file symbols (enums, structs, types):
> 1. Test with `npm test` (uses `transpileSource()`) — may pass with incomplete implementation
> 2. Verify with `npx tsx src/index.ts` (uses `run()`) — tests the full transpiler
> 3. Ensure both paths receive the same symbol information

The fact that tests can pass while CLI fails (or vice versa) indicates an architectural issue.

## Current Architecture

```
run()                          transpileSource()
  │                                  │
  ├─ Multi-file discovery            ├─ Single file
  ├─ Symbol collection               ├─ Optional context
  ├─ Dependency ordering             │
  │                                  │
  └──────────┬───────────────────────┘
             │
             ▼
      codeGenerator.generate()
```

While both paths converge at `generate()`, the setup phase differs significantly, leading to potential state inconsistencies.

### Option A: Single Entry Point with Configuration
Refactor to a single `transpile()` method that accepts a configuration object:

```typescript
interface TranspileOptions {
  files: string[];           // Single or multiple
  context?: CrossFileContext; // Optional symbol context
  mode: 'cli' | 'api' | 'test';
}

transpile(options: TranspileOptions): TranspileResult
```

### Option B: Extract Common Pipeline
Create a shared pipeline class that both entry points use:

```typescript
class TranspilePipeline {
  constructor(config: PipelineConfig) {}
  
  addFile(file: string): void
  addContext(ctx: CrossFileContext): void
  execute(): TranspileResult
}
```

### Option C: Strategy Pattern
Use strategy pattern to handle the differences:

```typescript
interface TranspileStrategy {
  collectSymbols(): ISymbolInfo
  orderFiles(): string[]
  generateOutput(): string
}

class CliStrategy implements TranspileStrategy { ... }
class TestStrategy implements TranspileStrategy { ... }
```

## Acceptance Criteria

- [ ] Single code path for transpilation logic
- [ ] Tests and CLI use the same underlying implementation
- [ ] No manual synchronization warnings needed in CLAUDE.md
- [ ] Existing tests continue to pass
- [ ] No performance regression

## References

- Issue #561: Cross-file modification tracking (required syncing `accumulatedModifications`)
- Issue #632: Float-to-int clamping (fortunately converged at `generate()`)
- CLAUDE.md "Cross-File Testing" and "Transpiler Code Paths" sections

jlaustill:
## Summary

- **Unified symbol collection timing**: Both `run()` and `transpileSource()` standalone now collect symbols BEFORE code generation (previously `run()` did this before, but standalone mode did it after)
- **Fixed SymbolTable not being cleared**: Calling `run()` twice on the same Transpiler instance now properly resets state
- **Fixed state not reset in standalone mode**: Repeated `transpileSource()` calls now maintain proper isolation
- **Eliminated duplicate `CNextResolver.resolve()` call**: Standalone mode was calling this twice; now reuses the result

## Test plan

- [x] Added 14 tests in `DualCodePaths.test.ts` verifying:
  - Single file transpilation parity between `run()` and `transpileSource()`
  - Struct, enum, scope, and const array handling identical in both paths
  - Multi-file cross-file references work correctly
  - State isolation between calls
  - Error handling parity
  - Symbol collection timing doesn't break type resolution
- [x] All 900 integration tests pass
- [x] All 2270 unit tests pass

Closes #634

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/src/transpiler/__tests__/DualCodePaths.test.ts
+++ b/src/transpiler/__tests__/DualCodePaths.test.ts
@@ -5,7 +5,7 @@
  * output for the same input, ensuring the refactoring doesn't break anything.
  */
 
-import { describe, it, expect, beforeEach } from "vitest";
+import { describe, it, expect, beforeEach, afterEach } from "vitest";
 import Transpiler from "../Transpiler";
 import ITranspilerConfig from "../types/ITranspilerConfig";
 import { mkdtempSync, writeFileSync, rmSync } from "node:fs";


jlaustill:
Thanks for the review!

**Fixed:** Added missing `afterEach` import in commit 7304e234.

**Regarding the `strlen()` regression test suggestion:** Good point. I investigated this and found that the original comment was likely outdated or overly cautious. The key insight is that `CodeGenerator.generate()` receives main file symbol info via the `symbolInfo` parameter (created from `TSymbolInfoAdapter.convert()` at line 899), not through the `symbolTable`. The `symbolTable` is primarily for cross-file type resolution.

The consolidation tests I added (`struct defined and used in same file`, `enum defined and used as function parameter`, `nested struct types`) verify that type resolution works correctly with the unified timing. These tests would catch any regression where types become "placeholder comments."

If you'd like, I can add a specific `strlen()` test case, though the existing tests should cover the underlying type resolution concern.

Issue #665, opened 2026-02-04
Scope array return with enum index generates bit-shift instead of array access


jlaustill:
## Description

When a scope function returns a scope-level array element indexed by an enum parameter, the transpiler generates a bit-shift expression instead of proper array access.

## Reproducer

```c-next
// test.cnx
enum EIndex {
    A,
    B,
    COUNT
}

scope Test {
    f32 values[global.EIndex.COUNT];

    public f32 get(EIndex idx) {
        return this.values[idx];  // Bug: generates (values >> idx) & 1
    }

    public void set(EIndex idx, f32 val) {
        this.values[idx] <- val;  // Works correctly
    }
}
```

## Expected Output

```c
float Test_get(EIndex idx) {
    return Test_values[idx];
}
```

## Actual Output

```c
float Test_get(EIndex idx) {
    return ((Test_values >> idx) & 1);
}
```

## Notes

- The `set` function generates correct array access code: `Test_values[idx] = val;`
- The bug appears to be specific to **returning** a scope array element
- The bit-shift pattern `>> idx & 1` suggests the transpiler is treating this as a bit extraction rather than array indexing
- This blocks the data-driven J1939 implementation in OSSM which needs `SensorValues.get(EValueId id)`

## Version

cnext v0.1.56

PR #666, opened 2026-02-04
fix(codegen): scope array return with enum index generates array access (#665)

jlaustill:
## Summary

- Fixes issue #665 where returning `this.values[idx]` from a scope function with enum-sized array generated bit extraction instead of array access
- Root cause: `trackVariableTypeWithName()` set `isArray` based on dimension resolution success, not array syntax presence
- Fix: Check for array syntax first (`arrayDim !== null && arrayDim.length > 0`), matching `trackVariableType()` pattern

## Test plan

- [x] Added comprehensive integration test with 18 validation points covering:
  - Enum-sized array with enum index (original bug)
  - Enum-sized array with integer index
  - Literal-sized array (regression check)
  - Multiple arrays in same scope
  - Return with index arithmetic
- [x] Added unit test specifically for issue #665 scenario
- [x] All 901 integration tests pass
- [x] All 3321 unit tests pass

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -274,7 +274,6 @@
       "integrity": "sha512-RsUFrSB0oQHEBnR8yarKIReUPwSu2ROpbjhdVKi4T/nQhMaS+TnIQPBwkMtb2r8A1KS2Hijw4D/4bV/XHoFQWw==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "engines": {
         "node": ">=20"
       }
@@ -356,8 +355,7 @@
       "resolved": "https://registry.npmjs.org/@cspell/dict-css/-/dict-css-4.0.19.tgz",
       "integrity": "sha512-VYHtPnZt/Zd/ATbW3rtexWpBnHUohUrQOHff/2JBhsVgxOrksAxJnLAO43Q1ayLJBJUUwNVo+RU0sx0aaysZfg==",
       "dev": true,
-      "license": "MIT",
-      "peer": true
+      "license": "MIT"
     },
     "node_modules/@cspell/dict-dart": {
       "version": "2.3.2",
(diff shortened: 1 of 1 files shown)


jlaustill:
## PR #666 Code Review: Scope Array Return Fix

**Overall Assessment: Approve** ✅

This PR correctly fixes issue #665 which was blocking the OSSM data-driven J1939 implementation.

---

### Root Cause Analysis

The bug was in `trackVariableTypeWithName()`:

```typescript
// Old (buggy):
const arrayDimensions = this._evaluateArrayDimensions(arrayDim);
const isArray = arrayDimensions !== undefined && arrayDimensions.length > 0;
```

When array size is an enum member like `global.EIndex.COUNT`, `_evaluateArrayDimensions()` couldn't resolve it, returning `undefined`. This caused `isArray = false`, which later caused the code generator to treat array access as bit extraction.

### Fix Quality

The fix is minimal and correct:

```typescript
// New (fixed):
const isArray = arrayDim !== null && arrayDim.length > 0;
const arrayDimensions = isArray
  ? this._evaluateArrayDimensions(arrayDim)
  : undefined;
```

This matches the pattern in `trackVariableType()` - check array syntax presence first, then try to resolve dimensions.

### Test Coverage

Excellent test coverage with 5 scenarios:
1. ✅ Enum-sized array with enum index (original bug)
2. ✅ Enum-sized array with integer index
3. ✅ Literal-sized array (regression check)
4. ✅ Multiple arrays in same scope
5. ✅ Return with index arithmetic

The integration test has 18 validation points, plus a dedicated unit test.

### Minor Notes

- The `package-lock.json` changes appear to be dependency cleanup (removing `"peer": true` from several packages). These are unrelated to the fix but harmless.

---

### Impact

This unblocks:
- OSSM `SensorValues.get(EValueId id)` function
- Any scope that returns array elements indexed by enum parameters

**Ready to merge.** 🚀

PR #688, opened 2026-02-05
test: close branch coverage gaps in analysis/ (round 2)

jlaustill:
## Summary
- Adds 15 unit tests across 7 analysis test files to close remaining branch coverage gaps
- ScopeStack and CommentExtractor reach **100% branch coverage**
- InitializationAnalyzer branch coverage jumps from 43.75% to **64.37%** (+21pp)
- Overall analysis/ branch coverage improves from 80.99% to **83.57%**

## Coverage Changes

| File | Branch Before | Branch After | Change |
|------|:---:|:---:|:---:|
| ScopeStack.ts | 93.75% | **100%** | +6.25 |
| CommentExtractor.ts | 94.59% | **100%** | +5.41 |
| DivisionByZeroAnalyzer.ts | 82.35% | **88.23%** | +5.88 |
| FloatModuloAnalyzer.ts | 70% | **73.33%** | +3.33 |
| FunctionCallAnalyzer.ts | 90.76% | **93.84%** | +3.08 |
| InitializationAnalyzer.ts | 43.75% | **64.37%** | +20.62 |

## Tests Added
- **ScopeStack**: `cloneState` with shadowed variable in nested scopes
- **CommentExtractor**: whitespace hidden tokens filtered in `getCommentsBefore`/`getCommentsAfter`
- **DivisionByZeroAnalyzer**: unary prefix divisor, parenthesized expression divisor
- **FloatModuloAnalyzer**: complex parenthesized expression in modulo operand
- **FunctionCallAnalyzer**: non-scope member access early return, C-Next symbol not matching C/C++
- **NullCheckAnalyzer**: nested scope null check with outer-scope c_ variable
- **InitializationAnalyzer**: if/else branch analysis, if-without-else restore, deterministic for-loop with struct field merge, write context tracking, declareParameter without active scope

## Test plan
- [x] All 7 modified test files pass individually
- [x] Full unit suite passes (3724 tests, 139 files)
- [x] Full integration suite passes (904/904)
- [x] All pre-push quality checks pass (prettier, oxlint, types, tests)

### DIFF
--- a/src/transpiler/logic/analysis/__tests__/FloatModuloAnalyzer.test.ts
+++ b/src/transpiler/logic/analysis/__tests__/FloatModuloAnalyzer.test.ts
@@ -361,7 +361,8 @@ describe("FloatModuloAnalyzer", () => {
       const analyzer = new FloatModuloAnalyzer();
       const errors = analyzer.analyze(tree);
 
-      // Parenthesized expression hits isFloatOperand fall-through
+      // Integer operands with parenthesized expression: isFloatOperand
+      // falls through to return false (neither float literal nor float identifier)
       expect(errors).toHaveLength(0);
     });
   });
--- a/src/transpiler/logic/analysis/__tests__/FunctionCallAnalyzer.test.ts
+++ b/src/transpiler/logic/analysis/__tests__/FunctionCallAnalyzer.test.ts
@@ -536,7 +536,7 @@ describe("FunctionCallAnalyzer", () => {
       expect(doSomethingErrors).toHaveLength(0);
     });
 
-    it("should return false from isExternalFunction when symbol has non-C/C++ language", () => {
+    it("should still flag E0422 for C-Next functions in symbol table", () => {
       const code = `
         void main() {
           cnextFunc();
--- a/src/transpiler/logic/analysis/__tests__/InitializationAnalyzer.test.ts
+++ b/src/transpiler/logic/analysis/__tests__/InitializationAnalyzer.test.ts
@@ -350,6 +350,22 @@ describe("InitializationAnalyzer", () => {
       analyzer.setWriteContext(false);
       expect(analyzer.isInWriteContext()).toBe(false);
     });
+
+    it("should not flag assignment target as uninitialized use", () => {
+      const code = `
+        void main() {
+          u32 x;
+          x <- 5;
+          u32 y <- x;
+        }
+      `;
+      const tree = parse(code);
+      const analyzer = new InitializationAnalyzer();
+      const errors = analyzer.analyze(tree);
+
+      // x is assigned (write context) then read - no error
+      expect(errors).toHaveLength(0);
+    });
   });
 
   // ========================================================================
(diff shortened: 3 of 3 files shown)


jlaustill (commented):
## Code Review: test: close branch coverage gaps in analysis/ (round 2)

Overall this is a solid coverage PR — the tests are well-structured, comments are clear, and each test targets a specific uncovered branch. A few observations:

### DivisionByZeroAnalyzer — parenthesized zero is a false negative

The test at `DivisionByZeroAnalyzer.test.ts` correctly documents that `10 / (0)` is **not** detected as division by zero because `isZero()` doesn't unwrap parenthesized expressions. The test is valid for branch coverage purposes, but this is a real false negative worth tracking as a separate issue. A user writing `x / (0)` or `x / (SOME_CONST)` wrapped in parens would silently bypass the check.

### InitializationAnalyzer — `declareParameter` test could be stronger

The `declareParameter` edge case test only asserts `not.toThrow()`. It would be more meaningful to also verify observable state — e.g., that the parameter is actually usable afterward without triggering E0381:

```typescript
it("should create implicit scope when declaring parameter without active scope", () => {
  const analyzer = new InitializationAnalyzer();
  
  // Should not throw - creates implicit scope
  expect(() => {
    analyzer.declareParameter("param1", 1, 0, "u32");
  }).not.toThrow();

  // Verify parameter is actually declared and initialized
  expect(analyzer.isInWriteContext()).toBe(false); // sanity check
  // Could also verify param1 is accessible in the scope
});
```

Without verifying post-conditions, this test proves the guard doesn't crash but doesn't prove it *works*.

### InitializationAnalyzer — `writeContext` tests are thin

Testing `setWriteContext`/`isInWriteContext` as a pure getter/setter round-trip is fine for line/branch coverage, but doesn't exercise how write context actually affects analysis behavior. Consider adding a test that demonstrates the *effect* of write context — e.g., that an assignment target in write context is not flagged as uninitialized use:

```cnx
void main() {
  u32 x;
  x <- 5;   // write context: x is the target, not a "use"
  u32 y <- x; // read context: x is now initialized
}
```

This would exercise the write context tracking through the actual analysis pipeline.

### FloatModuloAnalyzer — test comment accuracy

The comment says "Parenthesized expression hits isFloatOperand fall-through" which is technically correct, but it's worth noting the test doesn't actually test a *float* inside parentheses — it tests integer `u32` operands, so it falls through because neither operand has a float type. The branch being hit is the final `return false` in `isFloatOperand`, which is the "not a float literal and not a known float identifier" path. The test is valid, just the comment could be more precise.

### Nits

- **FunctionCallAnalyzer `isExternalFunction` test**: Good test. The name "should return false from isExternalFunction when symbol has non-C/C++ language" references a private method name, which couples the test description to implementation details. Something like "should still flag E0422 for C-Next functions in symbol table" focuses on the observable behavior instead. Minor nit though.

### Summary

The tests are well-targeted and the coverage gains are meaningful — especially the +21pp jump on InitializationAnalyzer. The ScopeStack and CommentExtractor reaching 100% branch coverage is a nice milestone. The parenthesized-zero false negative is the most interesting finding and probably deserves its own issue.

LGTM with the minor suggestions above. None are blocking.

PR #693, opened 2026-02-05
test: close branch coverage gaps in codegen/analysis (round 4)

jlaustill:
## Summary
- Add 16 tests across 5 files targeting uncovered branches in codegen and analysis layers
- **LiteralGenerator**: 7 new tests covering boolean literals (stdbool effect), f32/f64 float suffixes, and u64/i64/u8 integer suffix transforms → **100% branch coverage**
- **EnumAssignmentValidator**: 3 new tests for `this.WrongEnum` throw, `variable.field` non-enum throw, and known enum allow → **100% branch coverage**
- **MemberChainAnalyzer**: 4 new tests for null children early return, nested struct type tracking, non-integer nested field, and array-of-structs bit access
- **StructCollector**: 1 new test for const-resolved string array dimensions (constValues path in string array branch)
- **DivisionByZeroAnalyzer**: 1 new test for `getErrors()` accessor method

## Test plan
- [x] `npm run unit` — 3769 passed
- [x] `npm run test:q` — 904/904 passed
- [x] Coverage verified on all 5 target files

### DIFF
--- a/src/transpiler/logic/analysis/__tests__/DivisionByZeroAnalyzer.test.ts
+++ b/src/transpiler/logic/analysis/__tests__/DivisionByZeroAnalyzer.test.ts
@@ -372,8 +372,9 @@ describe("DivisionByZeroAnalyzer", () => {
       const analyzer = new DivisionByZeroAnalyzer();
       analyzer.analyze(tree);
 
-      expect(analyzer.getErrors()).toHaveLength(1);
-      expect(analyzer.getErrors()[0].code).toBe("E0800");
+      const errors = analyzer.getErrors();
+      expect(errors).toHaveLength(1);
+      expect(errors[0].code).toBe("E0800");
     });
   });
 });
--- a/src/transpiler/output/codegen/generators/expressions/__tests__/LiteralGenerator.test.ts
+++ b/src/transpiler/output/codegen/generators/expressions/__tests__/LiteralGenerator.test.ts
@@ -123,5 +123,46 @@ describe("LiteralGenerator", () => {
       expect(result.code).toBe("42");
       expect(result.effects).toEqual([]);
     });
+
+    it("should transform uppercase U64 suffix to ULL", () => {
+      const node = createMockLiteral("0xFFU64");
+      const result = generateLiteral(
+        node,
+        mockInput,
+        mockState,
+        mockOrchestrator,
+      );
+
+      expect(result.code).toBe("0xFFULL");
+      expect(result.effects).toEqual([]);
+    });
+  });
+
+  describe("passthrough literals", () => {
+    it("should pass through plain integer without effects", () => {
+      const node = createMockLiteral("42");
+      const result = generateLiteral(
+        node,
+        mockInput,
+        mockState,
+        mockOrchestrator,
+      );
+
+      expect(result.code).toBe("42");
+      expect(result.effects).toEqual([]);
+    });
+
+    it("should pass through string literal without effects", () => {
+      const node = createMockLiteral('"hello"');
+      const result = generateLiteral(
+        node,
+        mockInput,
+        mockState,
+        mockOrchestrator,
+      );
+
+      expect(result.code).toBe('"hello"');
+      expect(result.effects).toEqual([]);
+    });
   });
 });


jlaustill (commented):
## Code Review: test: close branch coverage gaps in codegen/analysis (round 4)

Test-only PR: 364 additions, 0 deletions across 5 files (1 new, 4 modified). 16 tests targeting codegen and analysis layers. Clean and well-organized.

---

### LiteralGenerator — New Test File (7 tests)

Good first unit test file for this generator. The tests correctly verify:

- Boolean `true`/`false` → stdbool include effect
- `f32` suffix → `f`, `f64` suffix → stripped
- `u64` → `ULL`, `i64` → `LL`, `u8` → stripped

The mock strategy is appropriate — `generateLiteral` only calls `node.getText()` so the minimal mock is sufficient, and the unused `mockInput`/`mockState`/`mockOrchestrator` are correctly typed as empty objects.

**Missing edge cases** that would strengthen coverage:

- **Hex/binary literals with suffixes**: `0xFFu64` → `0xFFULL`, `0b1010u8` → `0b1010`. The regex uses `$` anchor so these should work, but a test would confirm.
- **Uppercase variants**: `3.14F32`, `42U64` — the regex uses `[fF]32` and `[uU]64` but no test verifies the uppercase path.
- **Plain literals without suffixes**: `42`, `3.14`, `"hello"` — verifying these pass through unchanged with no effects would document the "no-op" path.

None of these are blocking, but the uppercase variant test in particular would be cheap to add and exercises a distinct regex branch.

### EnumAssignmentValidator — 3 Tests

These three tests neatly close the remaining branches in `validateNonEnumExpression`:

1. **`this.WrongEnum.MEMBER`** — Correctly exercises the `scopedEnumName !== typeName` throw path (line ~93). The test sets `getCurrentScope: () => "MyScope"` and uses `"this.Wrong.ACTIVE"` where `Wrong` doesn't match `State`.

2. **`variable.field` non-enum** — Exercises the `parts.length === 2 && !knownEnums.has(parts[0])` throw path. Uses the default `validator` from the `describe` block which has `knownEnums: new Set(["Color", "Status"])`, so `someVar` is not a known enum.

3. **`Status.OK` allowed** — Exercises the `knownEnums.has(parts[0])` allow path. `Status` IS a known enum, so it passes through even though we're assigning to `Color`. This is correct behavior — the earlier `getExpressionEnumType` check would catch actual type mismatches; `validateNonEnumExpression` only validates the expression *shape*.

These are tight, well-targeted tests.

### MemberChainAnalyzer — 4 Tests

These use manual mock objects rather than parsing real C-Next source. This is the right approach for MemberChainAnalyzer since it operates on AST node shapes rather than full programs.

**Null children early return** — Simple and correct. Exercises `if (!children || ...)` guard at line ~78.

**Nested struct type tracking** — This is the most valuable test in the PR. It sets up a two-level struct chain (`Outer.child: Inner`, `Inner.flags: u8`) and verifies that `obj.child.flags[0]` is correctly identified as bit access with `baseTarget: "obj.child.flags"`. This exercises the `currentStructType` tracking loop where the type transitions from `Outer` → `Inner` → `undefined` (since `u8` isn't a struct).

**Non-integer nested field** — Good complement to the previous test. Same struct chain shape but with `f32 value` instead of `u8 flags`, verifying that `isPrimitiveInt` is false so `isBitAccess` returns false. Confirms that subscript on a float field is treated as array access, not bit access.

**Array-of-structs bit access** — Exercises the `firstTypeInfo?.isArray && exprIndex === 1` branch (line ~153). The mock correctly has two expressions (index 0 for array subscript, index 1 for bit index) and the children array includes both `[0]` and `[1]` subscript patterns. Verifies `grid[0].flags[1]` → `isBitAccess: true, baseTarget: "grid[0].flags", bitIndex: "1"`.

One observation: the mock objects use `as never` type assertions, which silences TypeScript completely. This is pragmatic for test mocks but means the tests won't catch if the `AssignmentTargetContext` interface changes shape. A slightly safer approach would be `as unknown as Parser.AssignmentTargetContext`, which at least documents the intended type. Minor nit.

### StructCollector — 1 Test

This test exercises the `constValues?.has(dimText)` branch specifically within the string array dimension path (line ~64 in the source). The setup is clean: a `string<16>` field with `MAX_NAMES` dimension, and `constValues = new Map([["MAX_NAMES", 3]])`.

The assertion `expect(field?.dimensions).toEqual([3, 17])` correctly verifies both the const resolution (`3`) and the string capacity calculation (`16 + 1 = 17` for null terminator).

Note that the *non-string* array path (line ~84) has the same `constValues` branch but is not tested here. That branch was likely covered in an earlier round, but if not, it would be worth adding.

### DivisionByZeroAnalyzer — 1 Test

This test covers `getErrors()` — a one-line accessor that returns `this.errors`. While the coverage gain is real (the method wasn't called directly in previous tests, only `analyze()` which returns the same array), the test provides minimal value since `analyze()` already returns `this.errors` and is extensively tested. The assertion calling `getErrors()` twice without storing the result is also slightly wasteful:

```typescript
expect(analyzer.getErrors()).toHaveLength(1);
expect(analyzer.getErrors()[0].code).toBe("E0800");
```

Could be:
```typescript
const errors = analyzer.getErrors();
expect(errors).toHaveLength(1);
expect(errors[0].code).toBe("E0800");
```

Trivial nit, not blocking.

---

### Summary

16 well-targeted tests. The MemberChainAnalyzer nested struct and array-of-structs tests are the standout additions — they exercise complex type-tracking logic through multi-level member chains. The LiteralGenerator file is a good foundation that could be expanded with hex/binary/uppercase edge cases. The EnumAssignmentValidator tests cleanly close 3 specific branches to reach 100%.

LGTM.

PR #694, opened 2026-02-05
refactor: decompose PostfixExpressionGenerator to reduce cognitive complexity

jlaustill:
## Summary
- Extract 3 high-complexity functions (`generatePostfixExpression`: 79, `generateMemberAccess`: 67, `generateSubscriptAccess`: 59) into focused helpers each under the SonarCloud limit of 15
- Create `BitmapAccessHelper` to deduplicate 3 copies of bitmap field access pattern
- Add `MemberAccessValidator.validateGlobalEntityAccess` to deduplicate scope-requires-global error pattern for enums and registers
- Decompose `generateMemberAccess` into 9 null-coalescing handler functions
- Extract `handleSingleSubscript`, `handleBitRangeSubscript`, `handleFloatBitRange` from `generateSubscriptAccess`
- Extract `handleMemberOp`, `handleGlobalPrefix`, `handleThisScopeLength`, `tryPropertyAccess`, `initializeTrackingState` from `generatePostfixExpression`
- Bundle 11 mutable tracking variables into `ITrackingState` interface

## Test plan
- [x] All 65 existing PostfixExpressionGenerator unit tests pass
- [x] 6 new BitmapAccessHelper unit tests pass
- [x] 5 new MemberAccessValidator.validateGlobalEntityAccess tests pass
- [x] Full unit suite (3652 tests) passes
- [x] Full integration suite (904 tests) passes
- [x] oxlint, prettier, TypeScript type check all clean

### DIFF
--- a/src/transpiler/output/headers/generators/__tests__/generateStructHeader.test.ts
+++ b/src/transpiler/output/headers/generators/__tests__/generateStructHeader.test.ts
@@ -178,4 +178,45 @@ describe("generateStructHeader", () => {
       expect(result).toContain("uint32_t size;");
     });
   });
+
+  describe("string field dimensions", () => {
+    it("should not double the string capacity dimension for plain string fields", () => {
+      const fields = new Map<string, string>([
+        ["name", "string<64>"],
+        ["age", "u32"],
+      ]);
+      const dimensions = new Map<string, Map<string, readonly number[]>>([
+        ["Person", new Map([["name", [65]]])],
+      ]);
+      const input = createInput(new Map([["Person", fields]]), dimensions);
+
+      const result = generateStructHeader("Person", input);
+
+      expect(result).toContain("char name[65];");
+      expect(result).not.toContain("name[65][65]");
+      expect(result).toContain("uint32_t age;");
+    });
+
+    it("should not double the string capacity for array-of-string fields", () => {
+      const fields = new Map<string, string>([["arr", "string<64>"]]);
+      const dimensions = new Map<string, Map<string, readonly number[]>>([
+        ["Container", new Map([["arr", [4, 65]]])],
+      ]);
+      const input = createInput(new Map([["Container", fields]]), dimensions);
+
+      const result = generateStructHeader("Container", input);
+
+      expect(result).toContain("char arr[4][65];");
+      expect(result).not.toContain("arr[4][65][65]");
+    });
+
+    it("should handle string field with no dimensions in map", () => {
+      const fields = new Map<string, string>([["label", "string<16>"]]);
+      const input = createInput(new Map([["Widget", fields]]));
+
+      const result = generateStructHeader("Widget", input);
+
+      expect(result).toContain("char label[17];");
+    });
+  });
 });
--- a/src/transpiler/output/headers/generators/generateStructHeader.ts
+++ b/src/transpiler/output/headers/generators/generateStructHeader.ts
@@ -54,11 +54,19 @@ function generateStructHeader(name: string, input: IHeaderTypeInput): string {
     // Issue #461: Handle string<N> types which map to char[N+1]
     // The embedded dimension must come after array dimensions in C syntax
     // Example: string<64> arr[4] -> char arr[4][65], not char[65] arr[4]
+    //
+    // Fix: When dims already include the string capacity (from StructCollector),
+    // use dimSuffix directly. Only extract embedded dim when no dims provided.
     const embeddedMatch = /^(\w+)\[(\d+)\]$/.exec(cType);
     if (embeddedMatch) {
       const baseType = embeddedMatch[1];
       const embeddedDim = embeddedMatch[2];
-      lines.push(`    ${baseType} ${fieldName}${dimSuffix}[${embeddedDim}];`);
+      // dims already includes string capacity, so just use dimSuffix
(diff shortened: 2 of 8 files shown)


jlaustill (commented):
## Code Review: PostfixExpressionGenerator Decomposition

**Overall**: This is a well-executed refactoring. The decomposition is clean, behavior-preserving, and the null-coalescing chain pattern for `generateMemberAccess` is a particularly readable approach. Good test coverage on the new utilities.

### Strengths

- **`BitmapAccessHelper`** correctly deduplicates 3 copies of the bitmap field lookup + access pattern. The `errorDescriptor` parameter preserves the context-specific error messages without introducing branching inside the helper.
- **`validateGlobalEntityAccess`** is a clean consolidation of the identical enum/register scope-validation logic. The 5 tests cover all branches well.
- **`ITrackingState`** bundles 11 loose `let` variables into a single typed object, making the mutation surface explicit and easier to reason about.
- **Null-coalescing dispatch chain** (`tryBitmapFieldAccess ?? tryScopeMemberAccess ?? ...`) is elegant and self-documenting — the order of the chain makes the priority clear at a glance.
- **`initializeMemberOutput`** eliminates repeated default-output construction across all 9 handlers.

### Observations

1. **`BitmapAccessHelper.generate` (line 50-51)** — Minor: `fields?.has(memberName)` is checked, then `fields.get(memberName)!` is called. The `!` is safe because of the preceding `has()`, but you could simplify to `const fieldInfo = fields?.get(memberName); if (fieldInfo) { ... }` to avoid the non-null assertion entirely.

2. **`tryPropertyAccess` for `.capacity` / `.size`** — These look up type info via `primaryId` rather than `tracking.currentIdentifier`. This matches the original code, so it's behavior-preserving, but it means `.capacity`/`.size` on chained member expressions (e.g., `a.b.capacity`) would resolve against `a`'s type, not `a.b`'s. Worth filing as a separate issue if that's actually incorrect behavior.

3. **`handleMemberOp` mutation pattern** — This function mutates `tracking` in place, while `generateMemberAccess` and `generateSubscriptAccess` return result objects that the caller copies back. The asymmetry is documented and intentional (it avoids a second level of result-unpacking), but worth noting for future maintainers.

4. **Net -498 lines** across a 1058/498 add/delete — the new code is ~560 lines longer, which is expected given the additional function signatures and JSDoc for 9+ new helpers. The per-function cognitive complexity should be well under the SonarCloud limit of 15 now.

Looks good overall. All CI checks green, including SonarCloud quality gate.

jlaustill (commented):
## Code Review (Updated — 4 commits)

The PR has grown from the original refactoring into 4 well-structured commits that address the prior review feedback and fix a real bug. All CI checks pass.

### Commit-by-commit assessment

**Commit 1** — Original decomposition (reviewed previously). Looks good.

**Commit 2** — `IPostfixContext` and `IFloatBitRangeContext` to fix SonarCloud S107 (too many parameters). Clean solution. Also addressed the `fields?.has() + get()!` → `fields?.get()` suggestion.

**Commit 3** — Bug fix for `.capacity`/`.size` on struct member strings. The `resolveStringTypeInfo()` fallback chain (resolvedIdentifier → rootIdentifier → struct field lookup) is the right approach. Good that it mirrors how `.length` already handles this case. The integration test (`string-struct-capacity.test.cnx`) with 8 validations is thorough.

**Commit 4** — `primaryId` → `rootIdentifier`, `currentIdentifier` → `resolvedIdentifier`. The new names are much clearer about semantics (immutable root vs. evolving resolved state).

### Issues

1. **Destructuring in `handleMemberOp`** — Line ~473: `const { rootIdentifier, isStructParam, input, state, orchestrator, effects } = ctx;` violates the project's "No destructuring" coding standard (CLAUDE.md). Should use `ctx.rootIdentifier`, `ctx.input`, etc. throughout the function body instead.

2. **Header generation bug in test snapshot** — The committed `.expected.h` and `.test.h` files show struct string members doubled:
   ```c
   char name[65][65];  // should be char name[65]
   char bio[129][129];  // should be char bio[129]
   ```
   The `.expected.c` is correct (`char name[65]`), so this is a pre-existing header generator bug for string fields in structs. The test file is validating the current (buggy) header output. This should probably be tracked as a separate issue rather than blocking this PR, but it's worth noting since these snapshots will need updating when the header bug is fixed.

3. **`IPostfixContext` JSDoc says "Immutable context" but contains `effects: TGeneratorEffect[]`** — The `effects` array reference is stable but its contents are mutated via `push()`. Minor accuracy issue in the doc comment — the interface itself is immutable (fields don't change), but one field references a mutable collection.

### Minor observations

- `resolveStringTypeInfo` uses `as TTypeInfo` cast for the synthetic type info object. This is safe here since all the required fields are provided, but an alternative would be to construct a full `TTypeInfo` without the cast (if the type has optional fields that aren't set).
- The regex `/^string<(\d+)>$/` in `resolveStringTypeInfo` is clean and focused. Using `regex.exec(string)` instead of `string.match(regex)` aligns with the SonarCloud preference noted in the project conventions.

### Verdict

The refactoring is solid, the bug fix is a real improvement, and the renames add clarity. The destructuring issue (#1) is the only coding standards violation. The header bug (#2) is pre-existing and shouldn't block this PR.

PR #695, opened 2026-02-05
feat: improve hover provider symbol resolution with scope hierarchy

jlaustill:
## Summary
- **Scope hierarchy in symbols**: Added `parent` field to `IBaseSymbol` and propagated through all 6 collectors (Function, Variable, Enum, Bitmap, Struct, Register) so scope membership is tracked from the logic layer, not inferred later
- **Enum member hover support**: `TSymbolAdapter.convertEnum` now creates `EnumMember` symbols (like `OIL_PRES`) for hover/autocomplete — previously only the enum type itself was visible
- **Proper dot notation display**: Hover provider resolves parent chains to C-Next dot notation (`SensorProcessor.getAtmosphericPressurekPa.appData`) instead of C-style underscores, with cross-file resolution via workspace index
- **`this`/`global` keyword resolution**: `this.appData` now correctly resolves to the enclosing scope's member (e.g., `Ossm.appData`) instead of finding a random `appData` from another file
- **C/C++ fallback cleanup**: Replaces `::` with `.` in C/C++ extension hover fallback content for consistent C-Next notation
- **New symbol kinds**: Added `enum`, `enumMember`, `bitmap`, `bitmapField` to `TSymbolKind` with proper mapping

## Test plan
- [x] All 3816 unit tests pass
- [x] All 904 integration tests pass
- [x] Pre-push quality checks all pass
- [ ] Verify hover on `this.X` shows correct scope member (not a random cross-file match)
- [ ] Verify hover on `global.EnumName.Member` shows enum member info
- [ ] Verify hover on struct fields shows dot notation (not `::`)
- [ ] Verify "Defined in" shows top-level scope name only

### DIFF
--- a/src/transpiler/logic/symbols/cnext/adapters/TSymbolAdapter.ts
+++ b/src/transpiler/logic/symbols/cnext/adapters/TSymbolAdapter.ts
@@ -132,6 +132,8 @@ class TSymbolAdapter {
     });
 
     // Create EnumMember symbols for hover/autocomplete
+    // Note: sourceLine points to the enum declaration — IEnumSymbol.members
+    // is Map<string, number> (name → value) with no per-member line info.
     for (const [memberName, memberValue] of enumSym.members) {
       result.push({
         name: memberName,
--- a/vscode-extension/src/hoverProvider.ts
+++ b/vscode-extension/src/hoverProvider.ts
@@ -305,11 +305,19 @@ function formatSourceFooter(
  * e.g., "Scope_method" → "Scope.method"
  * Checks local symbols first, then workspace index for cross-file resolution.
  */
+/** Max recursion depth for parent chain resolution (guard against cycles) */
+const MAX_PARENT_DEPTH = 10;
+
 function resolveDisplayParent(
   parent: string,
   symbols: ISymbolInfo[],
   workspaceIndex?: WorkspaceIndex,
+  depth: number = 0,
 ): string {
+  if (depth >= MAX_PARENT_DEPTH) {
+    return parent;
+  }
+
   // Search local symbols first, then workspace-wide
   let parentSymbol = symbols.find((s) => s.fullName === parent);
   if (!parentSymbol && workspaceIndex) {
@@ -323,6 +331,7 @@ function resolveDisplayParent(
       parentSymbol.parent,
       symbols,
       workspaceIndex,
+      depth + 1,
     );
     return `${resolvedGrandparent}.${parentSymbol.name}`;
   }
@@ -564,6 +573,9 @@ export default class CNextHoverProvider implements vscode.HoverProvider {
     // Resolve C-Next qualifiers: "this" → current scope, "global" → scope lookup
     if (parentName === "this") {
       // "this.X" means X is a member of the enclosing scope
+      // TODO: This picks the first scope in the file, not the one enclosing
+      // the cursor. Fine for single-scope files; needs position-aware lookup
+      // for multi-scope files.
       const enclosingScope = symbols.find(
         (s) => s.kind === "namespace" || s.kind === "scope",
       );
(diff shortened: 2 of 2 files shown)


jlaustill (commented):
## Code Review: feat: improve hover provider symbol resolution with scope hierarchy

Nice work overall — this PR delivers meaningful UX improvements for the hover experience: dot notation, `this`/`global` resolution, enum member hovers, and proper scope chains. The architecture (pushing `parent` down to `IBaseSymbol` and propagating through all collectors) is the right approach. Here are my findings:

---

### High Confidence Issues

**1. Potential infinite loop in `resolveDisplayParent` (hoverProvider.ts:308-333)**

The recursive `resolveDisplayParent` walks the parent chain, but there's no cycle detection. If a symbol's parent chain ever forms a cycle (e.g., due to a bug in a collector), this will stack overflow.

```typescript
function resolveDisplayParent(parent, symbols, workspaceIndex) {
  // No visited set — if parentSymbol.parent eventually loops back, infinite recursion
  const resolvedGrandparent = resolveDisplayParent(parentSymbol.parent, ...);
}
```

Suggestion: Add a `visited: Set<string>` parameter (or a max depth constant like 10) to guard against cycles. This is a defensive measure — it may never trigger today, but it's cheap insurance.

**2. \`this\` resolution picks the *first* scope, not the *enclosing* scope (hoverProvider.ts:565-572)**

```typescript
if (parentName === "this") {
  const enclosingScope = symbols.find(
    (s) => s.kind === "namespace" || s.kind === "scope",
  );
```

\`symbols.find()\` returns the **first** scope/namespace in the symbol list, not necessarily the one enclosing the cursor position. If a file has multiple scopes (e.g., \`scope A { ... }\` then \`scope B { ... }\`), hovering \`this.x\` inside scope B would resolve to scope A.

Since the hover provider doesn't have cursor-to-scope mapping today, this is a known limitation — but it would be worth a \`// TODO\` comment acknowledging it for future improvement.

**3. Enum member \`sourceLine\` points to the enum declaration, not the member (TSymbolAdapter.ts:134-145)**

```typescript
for (const [memberName, memberValue] of enumSym.members) {
  result.push({
    sourceLine: enumSym.sourceLine,  // always the enum's line, not the member's
```

All enum members share the same \`sourceLine\` as the parent enum. The hover "Source" link will jump to the enum declaration line for every member. If \`IEnumSymbol.members\` is a \`Map<string, number>\`, there's no per-member line info available, so this is a data model limitation — but worth a comment explaining why.

---

### Medium Confidence Observations

**4. \`global.X\` resolution clears parentName entirely (hoverProvider.ts:573-577)**

```typescript
} else if (parentName === "global") {
  parentName = undefined;
}
```

This means \`global.LED\` falls through to the "no parent" path, finding the first symbol named \`LED\` in the file. This works for \`global.ScopeName\` but \`global.ScopeName.member\` would need the dot-chain logic to already have resolved \`parentName\` to \`ScopeName\` (not \`global\`). The comment says "global.X.Y is handled by the dot-chain logic" — I'd verify this works for 3-segment chains like \`global.Scope.method\` with a manual test.

**5. \`::\` replacement in C/C++ fallback is broad (hoverProvider.ts:717-729)**

```typescript
content.value.replace(/::/g, ".")
```

This replaces *all* \`::\` occurrences with \`.\`, which is correct for C++ scope resolution operators but could theoretically affect other text in hover content (e.g., URLs like \`https://\`, time notation, or documentation text containing \`::\`). In practice this is probably fine since C/C++ hover content is typically code, but something to keep an eye on.

**6. Workspace-wide \`.find()\` for cross-file lookups (hoverProvider.ts:585-588)**

```typescript
symbol = this.workspaceIndex
  .getAllSymbols()
  .find((s) => s.name === word && s.parent === parentName);
```

Linear scan over all workspace symbols on every hover. For small-to-medium projects this is fine, but for large codebases it could become a performance concern. The workspace index likely already has a lookup by name — consider using that instead of \`getAllSymbols().find()\` if available.

---

### Positive Callouts

- Clean propagation of \`parent\` through all 6 collectors — no collector was missed
- The \`TSymbolAdapter\` test updates are well-structured, verifying both the enum symbol and its 3 members with value assertions
- The \`TSymbolKind\` additions (\`enum\`, \`enumMember\`, \`bitmap\`, \`bitmapField\`) properly align with \`ESymbolKind\` values
- The \`parseWithSymbols.ts\` mapping cleanup (removing the stale comment "C-Next doesn't have classes") is a nice touch
- The \`buildSymbolHover\` switch cases for new kinds are thorough and consistent with existing patterns

PR #698, opened 2026-02-06
refactor: reduce generateVariableDecl cognitive complexity from 52 to ~10

jlaustill:
## Summary
- Refactored `generateVariableDecl` in CodeGenerator to reduce cognitive complexity from 52 to ~10
- Extracted shared `VariableModifierBuilder` helper to eliminate duplication across:
  - `CodeGenerator.generateVariableDecl`
  - `ControlFlowGenerator.generateForVarDecl`
- Added `IntegerLiteralValidator` helper using existing `LiteralUtils`
- Created focused helper methods to flatten control flow with early returns

## New Helpers

**VariableModifierBuilder** (`helpers/VariableModifierBuilder.ts`)
- Extracts const/atomic/volatile/extern modifiers from parser context
- Validates mutual exclusion of atomic and volatile
- Shared across variable and for-loop declarations

**IntegerLiteralValidator** (`helpers/IntegerLiteralValidator.ts`)
- Uses `LiteralUtils.parseIntegerLiteral` for literal detection (eliminates regex duplication)
- Centralizes error formatting with line/column information

## Extracted Methods in CodeGenerator
- `_inferVariableType`: handles c_ prefix pointer inference
- `_trackLocalVariable`: type registry and const tracking  
- `_handleArrayDeclaration`: array dimension and init
- `_parseFirstArrayDimension`: dimension parsing
- `_generateVariableInitializer`: init with validation
- `_validateIntegerInitializer`: integer validation
- `_finalizeCppClassAssignments`: C++ class handling

## Test plan
- [x] All 905 integration tests pass
- [x] All 3855 unit tests pass (including 23 new tests for helpers)
- [x] TypeScript type check passes
- [x] Linting passes

### DIFF
--- a/src/transpiler/output/codegen/CodeGenerator.ts
+++ b/src/transpiler/output/codegen/CodeGenerator.ts
@@ -6179,31 +6179,23 @@ export default class CodeGenerator implements IOrchestrator {
     const exprText = ctx.expression()!.getText().trim();
     const line = ctx.start?.line ?? 0;
     const col = ctx.start?.column ?? 0;
+    const isLiteral = LiteralUtils.parseIntegerLiteral(exprText) !== undefined;
 
-    // Use LiteralUtils to check if it's a literal
-    if (LiteralUtils.parseIntegerLiteral(exprText) !== undefined) {
-      // Direct literal - validate it fits in the target type
-      try {
+    try {
+      if (isLiteral) {
+        // Direct literal - validate it fits in the target type
         this._validateLiteralFitsType(exprText, typeName);
-      } catch (validationError) {
-        const msg =
-          validationError instanceof Error
-            ? validationError.message
-            : String(validationError);
-        throw new Error(`${line}:${col} ${msg}`, { cause: validationError });
-      }
-    } else {
-      // Not a literal - check for narrowing/sign conversions
-      try {
+      } else {
+        // Not a literal - check for narrowing/sign conversions
         const sourceType = this.getExpressionType(ctx.expression()!);
         this._validateTypeConversion(typeName, sourceType);
-      } catch (validationError) {
-        const msg =
-          validationError instanceof Error
-            ? validationError.message
-            : String(validationError);
-        throw new Error(`${line}:${col} ${msg}`, { cause: validationError });
       }
+    } catch (validationError) {
+      const msg =
+        validationError instanceof Error
+          ? validationError.message
+          : String(validationError);
+      throw new Error(`${line}:${col} ${msg}`, { cause: validationError });
     }
   }
 


jlaustill (commented):
## Code Review: Reduce generateVariableDecl Cognitive Complexity

Overall, this is a solid refactoring that successfully reduces cognitive complexity through well-structured helper extraction. The pattern of returning `{ handled: boolean; code: string; ... }` objects for early returns is clean and idiomatic.

### 🐛 Bug: IntegerLiteralValidator is Created but Not Used

The `integerLiteralValidator` instance is initialized in the generator setup (lines 2323-2333) but **never used**. Instead, `_validateIntegerInitializer()` duplicates the logic by calling `LiteralUtils.parseIntegerLiteral()` directly:

```typescript
// CodeGenerator.ts - creates the helper
this.integerLiteralValidator = new IntegerLiteralValidator({...});

// _validateIntegerInitializer - duplicates the logic instead of using it
if (LiteralUtils.parseIntegerLiteral(exprText) !== undefined) {
  // ...validation logic duplicated here
}
```

**Recommendation:** Either:
1. Use `this.integerLiteralValidator!.validateIntegerAssignment(typeName, exprText, line, col)` and remove the duplicated logic in `_validateIntegerInitializer`
2. Remove `IntegerLiteralValidator` class entirely and keep the inline logic

Option 1 is cleaner since you've already written tests for the helper.

### ⚠️ Suggestion: buildSimple Missing Validation

`buildSimple()` doesn't validate the atomic+volatile conflict like `build()` does:

```typescript
// build() validates
if (ctx.atomicModifier() && ctx.volatileModifier()) {
  throw new Error(...);
}

// buildSimple() does not
static buildSimple(ctx: IModifierContext): Pick<IVariableModifiers, "atomic" | "volatile"> {
  return {
    atomic: ctx.atomicModifier() ? "volatile " : "",
    volatile: ctx.volatileModifier() ? "volatile " : "",
  };
}
```

If a for-loop var declaration has both `atomic` and `volatile`, it would silently pass. Is this intentional? If so, a comment explaining why would help future maintainers.

### 👍 What's Good

1. **Clean extraction pattern** - `_inferVariableType`, `_trackLocalVariable`, `_handleArrayDeclaration` are well-named with single responsibilities

2. **Static class pattern** - `VariableModifierBuilder` follows the project's static class convention correctly

3. **Shared code elimination** - Using the same `VariableModifierBuilder` in both `CodeGenerator` and `ControlFlowGenerator` eliminates duplication

4. **Comprehensive tests** - 23 tests covering edge cases like error preservation and missing `start` tokens

5. **Early return pattern** - Converting nested conditionals to early returns with result objects is much more readable

### Minor Suggestions

1. **IModifierContext interface** - `constModifier` is optional to support `ForVarDeclContext`, but since `buildSimple` exists specifically for that case, consider making the interface stricter or splitting into two interfaces.

2. **Comment on atomic transpilation** - The `atomic` → `"volatile "` mapping might be confusing. A brief comment explaining ADR-049's rationale would help readers understand why atomic becomes volatile.

---

Once the `integerLiteralValidator` usage issue is resolved, this PR is ready to merge. Nice refactoring work! 🎯

PR #701, opened 2026-02-06
fix: unify getMemberTypeInfo across expression and assignment paths

jlaustill:
## Summary

- Replace hand-rolled `getMemberTypeInfo` in `buildHandlerDeps()` with delegation to new `_getFullMemberTypeInfo()` canonical method, fixing the assignment path's inability to resolve C header struct fields
- Widen `IOrchestrator.getMemberTypeInfo` return type from `{ baseType, isArray }` to `TTypeInfo` so both interfaces (`IOrchestrator` and `IHandlerDeps`) agree on the contract
- Add cross-file scope tests to `DualCodePaths.test.ts` covering both function calls and variable assignments through `run()` and `transpileSource()` paths

## Test plan

- [x] `npm run unit` — 3834 unit tests pass
- [x] `npm run test:q` — 905 integration tests pass
- [x] TypeScript compilation clean (`npx tsc --noEmit`)
- [x] New DualCodePaths tests verify cross-file scope function calls produce underscore notation (`Provider_getValue()`)
- [x] New DualCodePaths tests verify cross-file scope variable assignments produce underscore notation (`Storage_value = 42`)
- [x] Both `run()` and `transpileSource()` paths verified for each test

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -61,6 +61,9 @@ SonarCloud runs as a required PR check. **PRs cannot merge until the SonarCloud 
 npm run unit:coverage
 ```
 
+- **SonarCloud coverage on new code**: The quality gate requires >= 80% on new lines. Even 1 uncovered line in a small PR can fail the gate. Run `npm run unit:coverage` and check new/modified methods before pushing.
+- **Widening interface return types**: When changing a return type (e.g., `{ baseType, isArray }` → `TTypeInfo`), search for ALL test mocks of that method and update them to include newly required fields. Use `Grep` for the method name across `__tests__/` directories.
+
 Check the terminal report for files you changed — any new code below 80% coverage will fail the quality gate.
 
 **Useful API queries:**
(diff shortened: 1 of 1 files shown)


jlaustill:
Thanks for the thorough review! Addressing your observations:

**1. `isConst` hardcoded to `false`** — Intentional. C-Next struct fields don't have per-field const modifiers in the grammar (confirmed in `StructCollector.ts:41` which also hardcodes `isConst = false`). This maintains exact parity with the original inline implementation.

**2. `bitWidth` default of 32** — For struct member type tracking, this default matches the original `buildHandlerDeps` behavior (`TYPE_WIDTH[fieldType] ?? 32`). String types like `string<32>` aren't in `TYPE_WIDTH` but the `bitWidth` isn't used for string field operations — callers check `isString`/`stringCapacity` instead. A more precise default isn't needed today.

**3. String dimension filtering** — Correct that C macro dimensions (strings) are lost in the `TTypeInfo` return. This is by design: `TTypeInfo.arrayDimensions` is `number[]`, and the string macro names (e.g., `DEVICE_COUNT`) are passed through to generated C headers via `_getStructFieldInfo()` in a separate path. No `IHandlerDeps` caller needs the string dimensions — they use `arrayDimensions` for bounds checking which requires resolved numeric values.

Also pushed a test coverage fix (array struct field test) to address the SonarCloud quality gate — was at 78.3%, should now be above 80%.

PR #755, opened 2026-02-10
fix: handle C-Next style array types in C++ mode codegen

jlaustill:
## Summary

- **Array parameter codegen**: C-Next style `u8[8] param` now correctly generates `uint8_t param[8]` instead of `uint8_t& param` (scalar reference) in C++ mode. Fixes 14+ functions across user projects.
- **C++ class array init**: `CppClass[4] items` now generates `= {}` instead of `= {0}` when the type is from a C++ header, preventing compilation failures for classes with constructors.
- Both bugs shared the same root cause: codegen only checked `arrayDimension()` (C-style `u8 param[8]`) but missed `arrayType()` (C-Next style `u8[8] param`).

## Changes

| File | Change |
|------|--------|
| `CodeGenerator.ts` | Fix `generateParameter()`, `_getFunctionSignature()`, `registerCallbackType()`, `_getArrayZeroInitializer()` to check `arrayType()` |
| `CodeGenerator.test.ts` | 6 new unit tests covering both fixes |
| `tests/functions/array-param-cpp.test.cnx` | Integration test for C-Next style array params in C++ mode |
| `tests/array-initializers/cpp-class-array-init.test.cnx` | Integration test for C++ class array initialization |

## Test plan

- [x] 912/912 integration tests pass
- [x] 4832/4832 unit tests pass (6 new)
- [x] Verified against original reproducers (`repro.cnx` files)
- [x] Coverage on CodeGenerator.ts: 81.27% statements, 82.55% lines (above 80% gate)
- [x] All pre-push quality checks pass (prettier, cspell, oxlint, tsc, tests)

### DIFF
--- a/scripts/test-utils.ts
+++ b/scripts/test-utils.ts
@@ -20,7 +20,20 @@ import IValidationResult from "./types/IValidationResult";
 import ITestResult from "./types/ITestResult";
 import detectCppSyntax from "../src/transpiler/logic/detectCppSyntax";
 
+// Shared patterns for distinguishing C++ constructors from C function prototypes
+const C_KEYWORDS =
+  "return|if|while|for|switch|case|else|do|break|continue|goto|sizeof|typeof|alignof";
+const C_TYPES =
+  "void|int|char|float|double|long|short|unsigned|signed|bool|enum|struct|union|static|extern|const|volatile|inline|u?int\\d+_t|size_t";
+
 class TestUtils {
+  // First word of a line that is NOT a C++ constructor (keywords + C types)
+  static readonly NON_CONSTRUCTOR_FIRST_WORD = new RegExp(
+    `^(${C_KEYWORDS}|${C_TYPES})$`,
+  );
+  // Type keywords appearing in function arguments indicate a prototype, not a constructor
+  static readonly C_TYPE_IN_ARGS = new RegExp(`\\b(${C_TYPES})\\b`);
+
   /**
    * Normalize output for comparison (trim trailing whitespace, normalize line endings)
    */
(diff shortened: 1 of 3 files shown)


tests/functions/array-struct-member-cpp.expected.c
```diff
@@ -14,7 +14,7 @@
 #include <stdbool.h>
 
 // Array of external C structs with bool/enum members
-SensorReading sensors[4] = {0};
+SensorReading sensors[4] = {};
```
jlaustill: Nit: This change from `{0}` to `{}` appears to be a side-effect of the C++ class array fix. The `SensorReading` struct comes from a C++ header (has typed enum), so it now correctly uses `{}`. This is the right behavior, just noting it's an intentional consequence of the fix.

scripts/test-utils.ts
```diff
@@ -107,16 +108,23 @@ class TestUtils {
...
     // Split into two patterns to reduce regex complexity (SonarCloud S5843)
-    const constructorMatch = /^\s*(\w+)\s+\w+\([^)]*\)\s*;/m.exec(cCode);
+    const constructorMatch = /^\s*(\w+)\s+\w+\(([^)]*)\)\s*;/m.exec(cCode);
     if (constructorMatch) {
       const firstWord = constructorMatch[1];
-      const isKeyword =
-        /^(return|if|while|for|switch|case|else|do|break|continue|goto|sizeof|typeof|alignof)$/.test(
+      const argsContent = constructorMatch[2];
+      const isKeywordOrCType =
+        /^(return|if|while|for|switch|case|else|do|break|continue|goto|sizeof|typeof|alignof|void|int|char|float|double|long|short|unsigned|signed|bool|enum|struct|union|static|extern|const|volatile|inline|u?int\d+_t|size_t)$/.test(
```
jlaustill: Nit: This regex is getting complex with all the C types added. Consider extracting the type list to a constant for reuse between `isKeywordOrCType` and `argsHaveTypes` patterns.

Example:
```typescript
const C_TYPES = 'void|int|char|float|double|long|short|unsigned|signed|bool|u?int\\d+_t|size_t';
const C_KEYWORDS = 'return|if|while|for|switch|case|else|do|break|continue|goto|sizeof|typeof|alignof';
```

PR #765, opened 2026-02-11
feat: dual-mode test infrastructure for C and C++ validation

jlaustill:
## Summary

Overhauls the test infrastructure to run tests in **BOTH C and C++ modes** by default, catching sync bugs between output types.

- **Default behavior**: Tests now run in both C and C++ modes
- **New markers**: `// test-c-only`, `// test-cpp-only`, `// test-no-exec`
- **Mode-specific expected files**: `.expected.c`/`.cpp`, `.expected.h`/`.hpp`
- **New script**: `generate-cpp-snapshots.ts` for migration

## Bug Fixes Caught by Dual-Mode Testing

- Fix `strcmp` for single-char indexing (string[i] returns char, not string)
- Fix callback typedef to use pointer (*) in C mode, reference (&) in C++
- Fix `_Static_assert` → `static_assert` for C++ output
- Rename C++ reserved keywords used as variables (and, or, xor, mutable)

## Tests Marked as C-Only

- ARM intrinsic tests (critical sections, atomic operations)
- C interop tests using C-style enum bitwise operations
- Tests relying on pointer semantics for parameter indexing
- Tests using `{0}` initialization with enum struct fields

## Test plan

- [x] All 912 integration tests pass
- [x] All 4864 unit tests pass
- [x] `npm run test:all` passes

### DIFF
--- a/src/transpiler/output/codegen/__tests__/CodeGenerator.coverage.test.ts
+++ b/src/transpiler/output/codegen/__tests__/CodeGenerator.coverage.test.ts
@@ -54,6 +54,142 @@ describe("CodeGenerator Coverage Tests", () => {
   });
 
   // ==========================================================================
+  // NEW CODE IN PR: _isArrayAccessStringExpression (lines 811-852)
+  // ==========================================================================
+  describe("_isArrayAccessStringExpression() - PR new code", () => {
+    it("should return false for string property access (.length)", () => {
+      const source = `
+        string<32> name <- "test";
+        void main() {
+          u32 len <- name.length;
+        }
+      `;
+      const { code } = setupGenerator(source);
+      // .length returns a number, not a string
+      expect(code).toContain("strlen(name)");
+    });
+
+    it("should return false for string property access (.capacity)", () => {
+      const source = `
+        string<32> name <- "test";
+        void main() {
+          u32 cap <- name.capacity;
+        }
+      `;
+      const { code } = setupGenerator(source);
+      // .capacity returns a number
+      expect(code).toContain("32");
+    });
+
+    it("should return true for array of strings indexing", () => {
+      const source = `
+        string<32>[5] names;
+        void puts(string<32> s) {}
+        void main() {
+          puts(names[0]);
+        }
+      `;
+      const { code } = setupGenerator(source);
+      // Array of strings generates 2D char array
+      expect(code).toContain("names[0]");
+      expect(code).toContain("puts(");
+    });
+
+    it("should return false for single string character indexing", () => {
+      const source = `
+        string<32> name <- "hello";
+        void main() {
+          u8 ch <- name[0];
+        }
+      `;
+      const { code } = setupGenerator(source);
+      // Single string indexing returns a char, not a string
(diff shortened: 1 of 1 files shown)


CLAUDE.md
```diff
@@ -442,6 +442,48 @@ u32 main() {
...
+
+The test infrastructure supports running tests in **both C and C++ modes** to validate both output types.
+
+#### Test Mode Markers
+
+| Marker                   | Meaning                                                    |
+| ------------------------ | ---------------------------------------------------------- |
+| _(no marker)_            | Run in C mode only (backward compatibility)                |
+| `// test-c-only`         | Explicitly run only in C mode (e.g., MISRA-specific tests) |
+| `// test-cpp-only`       | Run only in C++ mode (e.g., C++ template interop tests)    |
```
jlaustill: **[important]** design: The table says "_(no marker)_" means "Run in C mode only (backward compatibility)" but the code in `TestUtils.getTestModes()` returns `["c", "cpp"]` (both modes) by default.

The PR body also says "Default behavior: Tests now run in both C and C++ modes". This CLAUDE.md documentation should be corrected to match.

```suggestion
| _(no marker)_            | Run in BOTH C and C++ modes (default)                      |
```

scripts/generate-cpp-snapshots.ts
```diff
@@ -0,0 +1,261 @@
...
+    });
+
+    const result = await pipeline.transpileSource(source, {
+      workingDir: dirname(cnxFile),
+      sourcePath: cnxFile,
+    });
+
+    if (!result.success) {
+      const errors = result.errors
+        .map((e) => `${e.line}:${e.column} ${e.message}`)
```
jlaustill: **[nit]** quality: For tests that have cross-file dependencies via `#include`, this script transpiles only the main `.test.cnx` file but doesn't generate `.expected.cpp/.hpp` for helper `.cnx` files. If a test includes a helper and the helper generates different output in C++ mode, the test may fail unexpectedly.

Consider documenting this limitation, or extending the script to find and process helper files referenced via includes.

scripts/test-utils.ts
```diff
@@ -63,11 +64,77 @@ class TestUtils {
...
+    if (TestUtils.hasCOnlyMarker(source)) return ["c"];
+    if (TestUtils.hasCppOnlyMarker(source)) return ["cpp"];
+    // Legacy support: test-cpp-mode means C++ only
+    if (TestUtils.hasCppModeMarker(source)) return ["cpp"];
+    return ["c", "cpp"]; // Default: both modes
+  }
+
+  /**
+   * Get expected file paths for a given mode
+   */
```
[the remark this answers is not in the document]
jlaustill: Keeping test-cpp-mode for backward compatibility. It's documented as a legacy marker and removing it would break existing tests without benefit.

scripts/generate-cpp-snapshots.ts
jlaustill: Noted for future enhancement. The current scope focuses on main test files. Helper file C++ snapshot generation can be added in a follow-up PR.

PR #775, opened 2026-02-12
fix: allow enum access inside scopes without global. prefix when unambiguous

### LINKED ISSUE #774: Enum values require global. prefix inside scopes even without naming conflict
## Description

Enum values accessed inside a scope require the `global.` prefix even when there is no naming conflict. The error message says to use `global.EnumName.VALUE` but `EnumName.VALUE` should be sufficient when unambiguous.

## Reproducer

```cnx
enum EColor { RED, GREEN, BLUE }

scope Test {
    public bool isRed(EColor c) {
        if (c = EColor.RED) { return true; }
        return false;
    }
}
```

## Expected

Transpiles successfully — `EColor.RED` is unambiguous (no scope member named `EColor`).

## Actual

```
Error: Use 'global.EColor.RED' to access enum 'EColor' from inside scope 'Test'
```

## Version

cnext v0.1.67

jlaustill:
## Summary

Fixes #774

- Previously, accessing a global enum from inside a scope always required the `global.` prefix
- Now, `global.` is only required when there's an actual naming conflict (scope has a member with the same name as the enum)
- Unambiguous access like `EColor.RED` works directly inside scopes

## Changes

- `MemberAccessValidator.validateGlobalEntityAccess()` now accepts optional `scopeMembers` parameter to detect actual naming conflicts
- `PostfixExpressionGenerator` passes `state.scopeMembers` to the validator for both enum and register access
- Updated unit tests for conflict-based validation behavior
- Added integration test `scope-enum-access.test.cnx` for unambiguous enum access
- Converted `cross-scope-enum-bare-error` test from error case to success case

## Test plan

- [x] New integration test validates enum access without `global.` prefix inside scopes
- [x] Unit tests verify `global.` is still required when there IS a naming conflict
- [x] All 939 integration tests pass
- [x] All 5050 unit tests pass

### DIFF
--- a/src/transpiler/output/codegen/generators/expressions/PostfixExpressionGenerator.ts
+++ b/src/transpiler/output/codegen/generators/expressions/PostfixExpressionGenerator.ts
@@ -1432,36 +1432,6 @@ const initializeMemberOutput = (
 });
 
 /**
- * Check if a scope member is shadowing a global enum.
- * This produces invalid C code, so we must error early with a helpful message.
- */
-const checkEnumShadowingConflict = (
-  ctx: IMemberAccessContext,
-  input: IGeneratorInput,
-  state: IGeneratorState,
-): void => {
-  if (!state.currentScope || !ctx.rootIdentifier) {
-    return;
-  }
-  // Check if the original identifier is a known global enum
-  if (!input.symbols!.knownEnums.has(ctx.rootIdentifier)) {
-    return;
-  }
-  // Check if it was resolved to something different (scope member)
-  if (ctx.result === ctx.rootIdentifier) {
-    return;
-  }
-  // Check if the resolved name is NOT an enum (scope member shadows global enum)
-  if (input.symbols!.knownEnums.has(ctx.result)) {
-    return;
-  }
-  // Conflict: scope member shadows global enum
-  throw new Error(
-    `Error: Use 'global.${ctx.rootIdentifier}.${ctx.memberName}' to access enum '${ctx.rootIdentifier}' from inside scope '${state.currentScope}' (scope member '${ctx.rootIdentifier}' shadows the global enum)`,
-  );
-};
-
-/**
  * Generate member access (obj.field).
  * Dispatches to specialized handlers via null-coalescing chain.
  */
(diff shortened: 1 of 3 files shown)


jlaustill (commented):
## Review Summary

**PR:** #775 — fix: allow enum access inside scopes without global. prefix when unambiguous

### Overview

This PR fixes issue #774 by making the `global.` prefix optional for enum/register access inside scopes when there's no naming conflict. The implementation adds conflict detection via `scopeMembers` parameter and introduces a new `checkEnumShadowingConflict()` function. The approach is sound, with comprehensive test coverage.

### Findings

| Severity | Count |
|----------|-------|
| Critical | 0 |
| Important | 1 |
| Moderate | 1 |
| Nit | 1 |

### Key Issues

#### [Important] Two validation paths with different logic (PostfixExpressionGenerator.ts)

The new `checkEnumShadowingConflict()` function (around line 1434) uses `ctx.result !== ctx.rootIdentifier` to detect shadowing, while `validateGlobalEntityAccess()` (called in `tryEnumMemberAccess`) uses `scopeMembers?.has(entityName)`. These are different detection mechanisms.

The flow is:
1. `checkEnumShadowingConflict()` throws if result differs AND root is a known enum AND result is NOT a known enum
2. `tryEnumMemberAccess()` → `validateGlobalEntityAccess()` throws if `scopeMembers.has(entityName)`

In the conflict case, the first check throws with a more detailed message including "(scope member 'X' shadows the global enum)". This seems intentional for better error messages, but consider documenting why both checks exist.

#### [Moderate] Consider consolidating or documenting dual validation

`checkEnumShadowingConflict()` runs before the handler dispatch chain, and would throw before `validateGlobalEntityAccess()` ever runs in the conflict case. If this is intentional (to provide a better error message for the shadowing case specifically), a comment explaining this would help future maintainers.

#### [Nit] JSDoc outdated (MemberAccessValidator.ts:57-58)

The JSDoc for `validateGlobalEntityAccess()` still says "inside a scope that doesn't own the entity" but the behavior changed to "inside a scope that has a naming conflict". Consider updating the description.

### What Looks Good

- Comprehensive test coverage with both unit tests and integration tests
- Clear, actionable error messages that suggest the fix (`Use 'global.EnumName.Member'...`)
- Good test refactoring: renamed tests to reflect new "conflict-based" behavior
- Integration tests cover both conflict (`scope-enum-naming-conflict`) and no-conflict scenarios
- Properly converted the old error test case into a success test

---

PR #802, opened 2026-02-13
refactor: consolidate symbol types to TType-based system

jlaustill:
## Summary

- Consolidated duplicate symbol types from `logic/symbols/types/` to `transpiler/types/symbols/`
- Updated all C-Next collectors to use `IScopeSymbol` references instead of string-based scope names
- TSymbols now use bare names with scope references; mangled names computed on demand
- Fixed test isolation issue in CodeGenerator.test.ts by resetting SymbolRegistry
- Added documentation for SymbolRegistry test isolation and TSymbol naming conventions

## Test plan

- [x] All 5221 unit tests pass
- [x] No lint errors
- [x] Dead code check passes

### DIFF
--- a/src/transpiler/logic/symbols/cnext/adapters/TSymbolInfoAdapter.ts
+++ b/src/transpiler/logic/symbols/cnext/adapters/TSymbolInfoAdapter.ts
@@ -232,7 +232,7 @@ class TSymbolInfoAdapter {
   // === Private Processing Methods ===
 
   // Use shared utility for name mangling
-  private static getMangledName = SymbolNameUtils.getMangledName;
+  private static readonly getMangledName = SymbolNameUtils.getMangledName;
 
   private static processStruct(
     struct: IStructSymbol,
--- a/src/transpiler/logic/symbols/cnext/collectors/ScopeCollector.ts
+++ b/src/transpiler/logic/symbols/cnext/collectors/ScopeCollector.ts
@@ -48,7 +48,7 @@ class ScopeCollector {
     // Get or create the scope via SymbolRegistry
     const scope = SymbolRegistry.getOrCreateScope(scopeName);
 
-    // Update scope metadata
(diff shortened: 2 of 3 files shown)


src/transpiler/types/symbols/IScopeSymbol.ts
```diff
@@ -0,0 +1,32 @@
...
+  readonly members: string[];
+
+  /** Functions in this scope */
+  readonly functions: IFunctionSymbol[];
+
+  /** Variables in this scope */
+  readonly variables: unknown[];
+
+  /** Visibility of each member */
+  readonly memberVisibility: ReadonlyMap<string, TVisibility>;
```
jlaustill: Nit: The `parent` field on `IScopeSymbol` extends from `IBaseSymbol.scope`, but the interface also has its own `parent` field. This duplication could be confusing. Consider documenting the distinction:
- `scope` (from IBaseSymbol): The containing scope for name resolution
- `parent`: Same as scope for scopes, but exists for hierarchical traversal

src/transpiler/logic/symbols/cnext/__tests__/FunctionCollector.test.ts
```diff
@@ -13,16 +20,24 @@ describe("FunctionCollector", () => {
...
       const funcCtx = tree.declaration(0)!.functionDeclaration()!;
-      const symbol = FunctionCollector.collect(funcCtx, "test.cnx");
+      const globalScope = TestScopeUtils.getGlobalScope();
+      const body = funcCtx.block();
+      const symbol = FunctionCollector.collect(
+        funcCtx,
+        "test.cnx",
+        globalScope,
+        body,
+      );
```
jlaustill: Nit: Consider extracting the repeated `funcCtx.block()` call and scope creation into a helper to reduce test boilerplate. Many tests have very similar setup code.

PR #828, opened 2026-02-18
fix: honor cppRequired from config file when --cpp not specified

### LINKED ISSUE #827: cnext.config.json cppRequired: true not honored
## Bug

`cppRequired: true` in `cnext.config.json` is not applied. The transpiler finds the config file but reports `cppRequired: false`.

## Minimal Reproduction

```bash
mkdir /tmp/cnext-bug && cd /tmp/cnext-bug
echo '{ "cppRequired": true }' > cnext.config.json
echo 'void foo() {}' > test.cnx

# Config is found but cppRequired shows false
cnext --config
# Output includes:
#   Config file:    /tmp/cnext-bug/cnext.config.json
#   cppRequired:    false

# Generates .c instead of .cpp
cnext test.cnx
# Output: Generated 1 output files: /tmp/cnext-bug/test.c

# --cpp flag works correctly
rm test.c
cnext test.cnx --cpp
# Output: Generated 1 output files: /tmp/cnext-bug/test.cpp
```

## Expected

`cnext --config` should show `cppRequired: true` and `cnext test.cnx` should generate `test.cpp`.

## Environment

- cnext v0.2.1
- Node v24.13.1 (via volta)
- Linux 6.17.9

jlaustill:
## Summary

- Fixes config file `cppRequired: true` being ignored when `--cpp` flag is not specified on CLI
- Root cause: `??` (nullish coalescing) only falls through on `null`/`undefined`, not `false`
- Changed merge logic from `??` to `||` so config file value is used when CLI doesn't specify `--cpp`

## Test plan

- [x] Added unit test for issue #827
- [x] All 157 CLI unit tests pass
- [x] Manual verification: `cnext --config` now shows `cppRequired: true` from config file

Fixes #827

### DIFF
--- a/src/cli/Cli.ts
+++ b/src/cli/Cli.ts
@@ -100,7 +100,7 @@ class Cli {
       defines: args.defines,
       preprocess: args.preprocess,
       verbose: args.verbose,
-      cppRequired: args.cppRequired ?? fileConfig.cppRequired ?? false,
+      cppRequired: args.cppRequired || fileConfig.cppRequired || false,
       noCache: args.noCache || fileConfig.noCache === true,
       parseOnly: args.parseOnly,
       headerOutDir: args.headerOutDir ?? fileConfig.headerOut,
--- a/src/cli/__tests__/Cli.test.ts
+++ b/src/cli/__tests__/Cli.test.ts
@@ -180,40 +180,39 @@ describe("Cli", () => {
       expect(result.config?.includeDirs).toContain("lib/");
     });
 
-    it("uses file config cppRequired when CLI is undefined", () => {
-      // This test documents that with yargs, cppRequired is always defined,
-      // so file config cppRequired only applies if we explicitly set CLI to undefined
+    it("honors config file cppRequired when CLI does not specify --cpp (issue #827)", () => {
+      // Issue #827: When user doesn't pass --cpp, config file cppRequired should be used
+      // yargs returns cppRequired: false as the default, but config file should override
       const fileConfig: IFileConfig = {
         cppRequired: true,
       };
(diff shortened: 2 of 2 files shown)


jlaustill (commented):
## Review Summary

**PR:** #828 — fix: honor cppRequired from config file when --cpp not specified

### Overview

This PR fixes a bug where `cppRequired: true` in the config file was being ignored when the `--cpp` flag wasn't specified on the CLI. The fix correctly changes the merge strategy from nullish coalescing (`??`) to logical OR (`||`).

### Findings

| Severity | Count |
|----------|-------|
| Critical | 0 |
| Important | 0 |
| Moderate | 0 |
| Nit | 0 |

### Analysis

The fix is correct. The root cause was that yargs provides `cppRequired: false` as the default when `--cpp` isn't specified. With `??`, this `false` value was considered "defined" and used instead of falling through to the config file.

The new `||` semantics mean: "use C++ mode if **either** CLI or config file enables it" — which matches user expectations and is consistent with how `noCache` and `debugMode` already work on adjacent lines.

| Merge operator | Behavior |
|----------------|----------|
| `??` (nullish) | Falls through only on `null`/`undefined` |
| `\|\|` (logical OR) | Falls through on any falsy value (`false`, `0`, `""`, etc.) |

### What Looks Good

1. **Consistent merge strategy** — The fix aligns `cppRequired` with `noCache` and `debugMode` which already use `||`
2. **Clear test updates** — Tests were updated to reflect realistic yargs behavior (CLI returns `false` when `--cpp` not specified, not `undefined`)
3. **Good commit message** — Clearly explains the root cause and fix

### Verdict

Looks good! The fix is minimal, correct, and well-tested.

---

Issue #872, opened 2026-02-19
MISRA Rule 7.2: Remaining 93 violations in array element assignments


jlaustill:
## Background

Issue #842 fixed 68% of MISRA Rule 7.2 violations by adding U suffix to unsigned integer literals. This follow-up addresses the remaining 93 violations.

## Current State

The fix in #842 adds U suffix when `expectedType` is known:
- Variable declarations: `u32 x <- 5;` → `x = 5U;` ✅
- Function arguments: `add(1, 2)` → `add(1U, 2U)` ✅ 
- Array initializers: `[1, 2, 3]` → `{1U, 2U, 3U}` ✅

## Remaining Violations

The following patterns do NOT get U suffix yet:

### Array element assignments
\`\`\`c-next
u32[10] arr;
arr[5] <- 42;  // Generates: arr[5] = 42; (missing U)
\`\`\`

### Array indices in assignment context
\`\`\`c-next
arr[5] <- 42;  // Index 5 doesn't get U suffix
\`\`\`

### Comparison expressions
\`\`\`c-next
if (magic[0] != 0xDEAD)  // 0xDEAD doesn't get U suffix
\`\`\`

## Technical Details

The fix requires propagating `expectedType` in additional contexts:
- `AssignmentContextBuilder` for array element assignments
- `BinaryExprGenerator` for comparisons where LHS type is unsigned
- Array subscript handling for the index literal

## Files to Modify

- `src/transpiler/output/codegen/assignment/handlers/` - Assignment handlers
- `src/transpiler/output/codegen/helpers/AssignmentContextBuilder.ts`
- `src/transpiler/output/codegen/generators/expressions/BinaryExprGenerator.ts`

## Related

- Parent issue: #839
- Fixed in: #842

PR #873, opened 2026-02-19
fix(misra): add U suffix to unsigned integer literals per Rule 7.2

### LINKED ISSUE #842: MISRA C:2012 Rule 7.2: 'U' suffix required on unsigned constants (287 violations)
## MISRA Rule

**Rule 7.2 (Mandatory)**: A 'u' or 'U' suffix shall be applied to all integer constants that are represented in an unsigned type.

## Violation Count

287 violations across test files.

## Description

Unsigned integer constants must have explicit `U` suffix to prevent implicit type conversion issues.

## Example Violation

```c
uint32_t mask = 0x80000000;  // MISRA 7.2 violation - needs U suffix
uint32_t mask = 0x80000000U; // Compliant
```

## Fix Strategy

Update CodeGenerator to:
1. Add `U` suffix to all unsigned literal values
2. Add `UL`/`ULL` suffix for larger unsigned types as appropriate

## Related

- Part of #839 breakdown
- MISRA C:2012 Mandatory rule
- Related to ADR-052 (Safe Numeric Literal Generation)

jlaustill:
## Summary
- Implements MISRA C:2012 Rule 7.2 compliance for unsigned integer literals
- Reduces violations from 287 to ~93 (68% reduction)
- Adds 13 unit tests for the new behavior

## Test plan
- [x] All 5299 unit tests pass
- [x] All 952 integration tests pass
- [x] Pre-push quality checks pass

## Implementation Details

The `LiteralGenerator` now checks `state.expectedType` and adds:
- `U` suffix for u8, u16, u32, uint8_t, uint16_t, uint32_t types
- `ULL` suffix for u64, uint64_t types

## Remaining Work

~93 violations remain in contexts where `expectedType` isn't propagated:
- Array element assignments
- Comparison expressions

These are tracked in follow-up issue #872.

Fixes #842


jlaustill:
**Re: Automated review feedback**

Thanks for the thorough review! A quick note on the two items raised:

1. **Octal handling**: The regex `\^\d` matches all numeric literals starting with a digit (0-9), which includes octal literals like `0777`. The second pattern `^0[xXbBoO]` specifically handles prefix notation. So octal IS covered by the first regex.

2. **Hex floats**: C-Next's parser doesn't support hex float syntax (`0x1.0p10`), so this edge case isn't applicable. The grammar rejects it at parse time.

Both items turn out to be non-issues in practice, but the analysis was correct about what *would* be a concern if these cases were possible.

Also fixed the SonarCloud suggestion to use `\d` instead of `[0-9]` in the regex.

PR #894, opened 2026-02-22
perf: parallelize C static analysis with matrix strategy

jlaustill:
## Summary

- Add `--tool` CLI flag to `batch-validate.mjs` to run individual analysis tools
- Update CI workflow to use GitHub Actions matrix strategy for parallel execution
- All 4 tools (cppcheck, clang-tidy, misra, flawfinder) now run concurrently instead of sequentially

## Test plan

- [ ] Verify `npm run validate:c -- --tool cppcheck` runs only cppcheck
- [ ] Verify `npm run validate:c -- --tool invalid` shows error message
- [ ] Verify CI matrix jobs run in parallel and complete successfully

### DIFF
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -351,6 +351,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
+        # Note: 'misra' is a cppcheck addon, not a standalone tool
+        # batch-validate.mjs handles this dependency automatically
         tool: [cppcheck, clang-tidy, misra, flawfinder]
 
     steps:
--- a/scripts/test-utils.ts
+++ b/scripts/test-utils.ts
@@ -427,6 +427,8 @@ class TestUtils {
     // Pattern 2: Type& paramName in function param context
     // Requires: type starts with letter, & follows, param name starts with letter, ends with , or )
     // Uses [a-zA-Z] instead of \w to exclude patterns like "0xFFU & value)" which are bitwise ops
+    // NOTE: Intentionally permissive — may match in typedef/macro contexts, which is acceptable
+    // since the goal is C++ compiler selection (false positives are harmless, false negatives break builds)
     if (/[a-zA-Z_]\w*\s*&\s*[a-zA-Z_]\w*\s*[,)]/.test(cCode)) {
       return true;
     }
@@ -739,6 +741,9 @@ class TestUtils {
       result.headerMatch = true;
 
       // Find existing helper implementation files
+      // NOTE: Missing helpers are silently skipped — if transpile phase completed successfully,
+      // helpers should exist; if they don't, it's likely the test doesn't use helpers in this mode
+      // (e.g., a helper only needed in C++ mode but we're running C mode)
       for (const helperCnx of helperCnxFiles) {
         const helperBaseName = basename(helperCnx, ".cnx");
         const implExt = mode === "cpp" ? "cpp" : "c";
@@ -932,9 +937,10 @@ class TestUtils {
 
     // Execute if requested and not ARM-only code
     if (shouldExec && /^\s*\/\/\s*test-execution\s*$/m.test(source)) {
-      // Read generated code to check for ARM runtime requirements
-      const generatedCode = readFileSync(expectedImplPath, "utf-8");
-      if (TestUtils.requiresArmRuntime(generatedCode)) {
+      // Read existing code to check for ARM runtime requirements
+      // (may be freshly generated or pre-existing in executeOnly mode)
+      const existingCode = readFileSync(expectedImplPath, "utf-8");
+      if (TestUtils.requiresArmRuntime(existingCode)) {
         result.execSuccess = true;
         result.skippedExec = true;
         // No cleanup needed for helper files


jlaustill (commented):
## Review Summary

**PR:** #894 — perf: parallelize C static analysis with matrix strategy

### Overview
This PR optimizes CI by splitting integration tests into transpile and execute phases, enabling C static analysis tools to run in parallel via GitHub Actions matrix strategy. The architecture is sound and the implementation is well-structured.

### Findings

| Severity | Count |
|----------|-------|
| Critical | 0 |
| Important | 1 |
| Moderate | 2 |
| Nit | 1 |

### Key Issues

1. **[important]** In `test-utils.ts` line 932, `executeOnly` mode reads the generated code file but `transpileResult` is not in scope — this variable is referenced on the old line 884 but the new code correctly reads from `expectedImplPath` instead. However, this highlights a potential future maintenance issue since the code paths diverge significantly.

2. **[moderate]** The C++ reference parameter detection regex could have false positives — see inline comment.

3. **[moderate]** In `executeOnly` mode, helper files that don't exist are silently skipped rather than failing — this may mask configuration issues.

### What Looks Good

- Clean separation of transpile and execute phases with clear options interface (`ITestOptions`)
- Matrix strategy implementation in GitHub Actions is correct and uses `fail-fast: false` appropriately
- The `--tool` flag in `batch-validate.mjs` has proper validation with helpful error messages
- Good handling of edge cases like MISRA requiring cppcheck availability

---

scripts/test-utils.ts
```diff
@@ -418,6 +419,18 @@ class TestUtils {
       return true;
     }
 
+    // C++ reference parameters in function declarations
+    // Pattern 1: const Type& paramName (const reference)
+    if (/\bconst\s+\w+\s*&\s*\w+/.test(cCode)) {
+      return true;
+    }
+    // Pattern 2: Type& paramName in function param context
```
jlaustill: **[moderate]** quality: The second regex `[a-zA-Z_]\w*\s*&\s*[a-zA-Z_]\w*\s*[,)]` could match type declarations in other contexts, not just function parameters.

For example, it would match inside a `typedef` or macro context like:
```c
typedef int (*Callback)(int& ref);
```

While this is likely fine for the current use case (detecting C++ features to switch compilers), consider adding a comment noting this is intentionally permissive rather than precise, to avoid future maintainers thinking it's a bug.

scripts/test-utils.ts
```diff
@@ -703,108 +718,144 @@ class TestUtils {
...
 
-    // Transpile helper files via CLI
-    // NOTE: Don't use -o flag here. The CLI's -o flag causes a rename operation
-    // that would move tracked helper files to temp locations. Instead, let
-    // helpers generate in place (they're tracked in git anyway).
-    const helperImplFiles: string[] = [];
+      // Mark transpile/snapshot phases as passed (already validated in transpile phase)
+      result.transpileSuccess = true;
+      result.snapshotMatch = true;
+      result.headerMatch = true;
```
jlaustill: **[moderate]** design: In `executeOnly` mode, helper files that don't exist are silently skipped (line 739: `if (existsSync(helperImplFile))`). This differs from normal mode where helper transpilation failures are still processed.

Consider whether missing helper files in `executeOnly` mode should:
1. Be a hard error (file should exist if transpile phase completed)
2. Log a warning
3. Remain silent (current behavior)

If intentional, a comment explaining why would help future maintainers.

scripts/test-utils.ts
```diff
@@ -881,7 +932,9 @@ class TestUtils {
 
     // Execute if requested and not ARM-only code
     if (shouldExec && /^\s*\/\/\s*test-execution\s*$/m.test(source)) {
```
jlaustill: **[nit]** quality: Good fix reading from `expectedImplPath` instead of `transpileResult.code` — this makes the code work correctly in `executeOnly` mode where `transpileResult` wouldn't be available.

The variable name `generatedCode` is accurate, though `existingCode` might be slightly clearer since in `executeOnly` mode it's pre-existing rather than freshly generated.

.github/workflows/pr-checks.yml
```diff
@@ -304,11 +343,15 @@ jobs:
...
+    needs: [integration-transpile]
     # Advisory: pre-existing MISRA violations exist; don't block PR merges
     continue-on-error: true
+    strategy:
+      fail-fast: false
+      matrix:
+        tool: [cppcheck, clang-tidy, misra, flawfinder]
 
     steps:
       - name: Checkout code
```
jlaustill: **[important]** design: The matrix includes `misra` as a separate tool, but `batch-validate.mjs` treats MISRA as a cppcheck addon (it runs via `runMisra()` which depends on cppcheck). The code handles this correctly with the `if (selectedTool === "misra" && !hasCppcheck)` check, but the matrix implies these are independent tools.

Consider either:
1. Documenting in a comment that `misra` is a cppcheck addon, or
2. Using `cppcheck-misra` as the matrix value to make the relationship explicit

This would help future maintainers understand the dependency.

scripts/test-utils.ts
jlaustill: Fixed. Renamed to `existingCode` with a comment clarifying it may be freshly generated or pre-existing in executeOnly mode.

.github/workflows/pr-checks.yml
jlaustill: Fixed. Added comment in the matrix section noting that MISRA is a cppcheck addon and that batch-validate.mjs handles the dependency.

PR #953, opened 2026-02-24
fix: scope variables with forward-declared struct types generate as pointers (#948)

### LINKED ISSUE #948: Scope variables with forward-declared struct types generate as value instead of pointer
## Reproduction

```bash
cnext test.cnx
```

**fake_lib.h:**
```c
#ifndef FAKE_LIB_H
#define FAKE_LIB_H
#include <stdint.h>
typedef struct _widget_t widget_t;
widget_t * widget_create(void);
void widget_set_value(widget_t * w, int32_t val);
#endif
```

**test.cnx:**
```cnx
#include "fake_lib.h"

scope MyScope {
  widget_t w;

  public void init() {
    this.w <- global.widget_create();
    global.widget_set_value(this.w, 42);
  }
}
```

## Expected output (test.c)

```c
static widget_t* MyScope_w = NULL;

void MyScope_init(void) {
    MyScope_w = widget_create();
    widget_set_value(MyScope_w, 42);
}
```

## Actual output (test.c)

```c
static widget_t MyScope_w = {0};

void MyScope_init(void) {
    MyScope_w = widget_create();
    widget_set_value(&MyScope_w, 42);
}
```

## Problem

When a scope variable has a forward-declared (incomplete) struct type, the transpiler:
1. Generates it as a value type with `= {0}` — this fails at compile time for incomplete types
2. Adds `&` when passing to `global.` functions — creates `widget_t**` instead of `widget_t*`

## Expected behavior

Forward-declared struct types should be treated as opaque pointer types (like `esp_lcd_panel_handle_t`), generating pointer scope variables.

## Context

Blocks LVGL integration. `lv_obj_t` is forward-declared in LVGL headers. Scope variables like `lv_obj_t scale;` fail to compile.

Minimal repro at: `/tmp/cnext-bugs/scope-var-incomplete-type/`

jlaustill:
## Summary
Fixes scope variables with forward-declared (opaque) struct types generating as values instead of pointers.

When a scope variable uses a forward-declared struct type like `typedef struct _widget_t widget_t;`, it must be generated as a pointer with NULL initialization instead of a value with {0} initialization.

**Before (wrong):**
```c
static widget_t MyScope_w = {0};  // Fails - incomplete type can't be instantiated
```

**After (correct):**
```c
static widget_t* MyScope_w = NULL;  // Valid - pointer to incomplete type
```

## Test plan
- [x] Added `bugs/issue-948-forward-decl-scope-var/` - main test case
- [x] Added `bugs/issue-948-forward-then-define/` - edge case test (forward decl then full def)
- [x] Added unit tests in `CResolver.integration.test.ts` for opaque type detection
- [x] All 942 integration tests pass
- [x] All 5522 unit tests pass

## Changes
- **SymbolTable**: Add opaque type tracking (markOpaqueType, isOpaqueType, getAllOpaqueTypes, restoreOpaqueTypes, registerStructTagAlias)
- **StructCollector (C)**: Detect forward declarations and mark as opaque; unmark when full definition found
- **CResolver**: Extract typedef name from init-declarator-list for `typedef struct _foo foo;` patterns
- **DeclaratorUtils**: Add extractFirstDeclaratorName helper
- **ScopeGenerator**: Generate pointer with NULL for opaque types
- **CallExprGenerator**: Skip address-of for opaque scope variables (already pointers)
- **CacheManager**: Store/restore opaque types (cache version bumped to 5)
- **ICodeGenSymbols**: Add opaqueTypes field with test mock updates

Closes #948

### DIFF
--- a/src/transpiler/logic/symbols/c/index.ts
+++ b/src/transpiler/logic/symbols/c/index.ts
@@ -114,29 +114,7 @@ class CResolver {
     const isTypedef = DeclaratorUtils.hasStorageClass(declSpecs, "typedef");
     const isExtern = DeclaratorUtils.hasStorageClass(declSpecs, "extern");
 
-    // Check for struct/union
-    const structSpec = DeclaratorUtils.findStructOrUnionSpecifier(declSpecs);
-    if (structSpec) {
-      CResolver.collectStructSymbol(
-        structSpec,
-        decl,
-        declSpecs,
-        sourceFile,
-        line,
-        symbolTable,
-        isTypedef,
-        symbols,
-        warnings,
-      );
-    }
-
-    // Check for enum
-    const enumSpec = DeclaratorUtils.findEnumSpecifier(declSpecs);
-    if (enumSpec) {
-      CResolver.collectEnumSymbols(enumSpec, sourceFile, line, symbols);
-    }
-
-    // Collect declarators (variables, function prototypes, typedefs)
+    // Build context early for reuse across helper methods
     const ctx: IDeclarationContext = {
       sourceFile,
       line,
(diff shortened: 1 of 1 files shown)


debug-symbols.ts
```diff
@@ -0,0 +1,15 @@
+import CParser from "./src/transpiler/logic/parser/c/CParser";
```
jlaustill: **[important]** quality: This debug script is committed at the project root. The CLAUDE.md says "No root-level analysis artifacts". This appears to be a development artifact used during implementation and should be removed from the PR.

PR #972, opened 2026-02-25
feat: single entry point file replaces directory input

jlaustill:
## Summary

- Replace directory-based file discovery with single `.cnx` entry point (`cnext src/main.cnx` instead of `cnext src/`)
- All other files discovered automatically via `#include` directives (already worked, directory scanning was redundant)
- Fix self-include bug: generated `.cpp` files now use headerOut-relative paths (`#include "Data/Foo.h"` instead of flat `#include "Foo.h"`), eliminating need for per-subdirectory `-I` flags
- Infer `basePath` from entry file's parent directory (overridable via config)
- Remove `FileDiscovery.discover()`, `InputExpansion.findCNextFiles()`, `fast-glob` dependency, and `IDiscoveryOptions` type
- CLI validates single file input with clear error messages for directories, multiple files, and nonexistent files

## Test plan

- [x] 5523 unit tests pass
- [x] 950 integration tests pass
- [x] 28 CLI integration tests pass
- [x] All pre-push quality checks pass (TypeScript, oxlint, prettier, knip, cspell)
- [ ] Test against OSSM build (`cnext src/main.cnx` instead of `cnext src`)
- [ ] Test against OGauge build (already uses `cnext src/main.cnx`)
- [ ] Verify OSSM can remove extra `-I include/Data` etc. flags from platformio.ini

### DIFF
--- a/tests/integration/issue-349-angle-include-paths.test.ts
+++ b/tests/integration/issue-349-angle-include-paths.test.ts
@@ -99,7 +99,7 @@ async function testSiblingAngleBracketInclude() {
   writeFileSync(join(displayDir, "main.cnx"), mainSource, "utf-8");
 
   const pipeline = new Transpiler({
-    input: sourceDir,
+    input: join(displayDir, "main.cnx"),
     outDir: codeOutDir,
     headerOutDir: headerOutDir,
     includeDirs: [sourceDir],
@@ -156,7 +156,7 @@ i32 main() {
   if (existsSync(headerOutDir)) rmSync(headerOutDir, { recursive: true });
 
   const pipeline = new Transpiler({
-    input: sourceDir,
+    input: join(sourceDir, "main.cnx"),
     outDir: codeOutDir,
     headerOutDir: headerOutDir,
     includeDirs: [sourceDir],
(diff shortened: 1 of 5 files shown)


tests/integration/issue-349-angle-include-paths.test.ts
```diff
@@ -99,7 +99,7 @@ async function testSiblingAngleBracketInclude() {
   writeFileSync(join(displayDir, "main.cnx"), mainSource, "utf-8");
 
   const pipeline = new Transpiler({
-    inputs: [sourceDir],
+    input: sourceDir,
```
jlaustill: **[critical]** bug: This test passes a **directory** to `input`, but the PR changes the API to only accept entry point files.

The CLI now rejects directories with: "Error: Directory input not supported. Specify an entry point file."

This applies to multiple tests in this file (lines 102, 159, 202, 256, 319) and several other integration tests:
- `issue-339-include-paths.test.ts` (lines 113, 195, 255, 313)
- `issue-294-cross-scope-bare-error.test.ts` (line 73)
- `determineProjectRoot.test.ts` (line 205)

These tests will pass now because `Transpiler` doesn't validate the input internally, but they no longer test realistic usage since the CLI rejects this pattern.

PR #976, opened 2026-02-26
fix: correct .hpp include extensions and update PIO/docs for entry-point CLI

jlaustill:
## Summary
- **Fix .hpp include bug**: `IncludeResolver` was created before `cppDetected` was determined, producing `.h` directives for `.cnx` includes in C++ mode instead of `.hpp`. This caused duplicate includes (`AppData.hpp` + `AppData.h`) in generated headers. Added `cppMode` parameter to `IncludeResolver` and stem-based dedup in `HeaderGeneratorUtils` as defense-in-depth.
- **Fix `--pio-install` script**: Generated PlatformIO build script still used old per-file glob pattern. Updated to entry-point approach (`cnext src/main.cnx`).
- **Update docs**: CLI help text, ADR-010, and ADR-048 now reflect single entry-point usage.

## Test plan
- [x] 5564 unit tests pass
- [x] 950/950 integration tests pass
- [x] issue-328 (false symbol conflict) passes with correct single `.hpp` include
- [x] chain-forward (transitive includes) passes with correct `.hpp` extension
- [x] All pre-push quality checks pass (prettier, oxlint, spelling, dead code, types)
- [ ] Verify ossm project builds clean with these fixes

### DIFF
--- a/src/cli/PlatformIOCommand.ts
+++ b/src/cli/PlatformIOCommand.ts
@@ -46,7 +46,7 @@ class PlatformIOCommand {
     // Create cnext_build.py script
     // Issue #833: Run transpilation at import time (before compilation),
     // not as a pre-action on buildprog (which runs after compilation)
-    const buildScript = `Import("env")
+    const buildScript = String.raw`Import("env")
 import subprocess
 import sys
 from pathlib import Path
@@ -67,7 +67,7 @@ def transpile_cnext():
             text=True
         )
         if result.stdout:
-            lines = result.stdout.strip().split("\\n")
+            lines = result.stdout.strip().split("\n")
             for line in lines:
                 if line.startswith(("Compiled", "Collected", "Generated")):
                     print(f"  {line}")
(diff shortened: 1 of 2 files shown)


src/transpiler/Transpiler.ts
```diff
@@ -491,7 +491,11 @@ class Transpiler {
     );
 
     // Resolve includes from source content
-    const resolver = new IncludeResolver(searchPaths, this.fs);
+    const resolver = new IncludeResolver(
+      searchPaths,
+      this.fs,
+      this.cppDetected,
```
jlaustill: **[moderate]** design: The `cppDetected` value passed here will only be `true` if `--cpp` flag was explicitly used. Auto-detection from `.hpp` files or C++ syntax in headers happens later during `_collectAllHeaderSymbols()` (Stage 2), after include discovery (Stage 1) has already completed.

The stem-based dedup in `HeaderGeneratorUtils.generateIncludes()` handles this timing mismatch gracefully, but this creates an implicit dependency between the two mechanisms. Consider adding a brief comment here explaining this intentional design:

```suggestion
    const resolver = new IncludeResolver(
      searchPaths,
      this.fs,
      this.cppDetected, // Note: Only true if --cpp flag was used; auto-detection happens later. HeaderGeneratorUtils handles extension mismatch via stem-based dedup.
    );
```

Issue #981, opened 2026-02-27
Macro-sized array field on local struct generates bit extraction instead of array access


jlaustill:
## Summary

When a C struct has an array field whose size is defined by a `#define` macro (e.g., `uint8_t data[BUF_SIZE]`), array indexing on a **local variable** of that struct type generates bit extraction (`(msg.data >> i) & 1`) instead of array access (`msg.data[i]`).

The same code works correctly when the struct is a **function parameter**.

Replacing the macro with a literal (`uint8_t data[8]`) also produces correct output.

## Trigger

All three conditions required:
1. Array field size is a `#define` macro (not a literal)
2. Struct is a local variable (not a function parameter)
3. Array indexing with `msg.data[i]`

## Reproduction

**fake_lib.h**
\`\`\`c
#ifndef FAKE_LIB_H
#define FAKE_LIB_H

#include <stdint.h>

#define BUF_SIZE 8

typedef struct {
    uint8_t data[BUF_SIZE];
} msg_t;

void fill_msg(msg_t* m);
void use_byte(uint8_t b);

#endif
\`\`\`

**test.cnx**
\`\`\`c-next
#include "fake_lib.h"

scope Test {
  // WORKS: msg is a parameter → generates msg.data[i]
  void from_param(msg_t msg) {
    u8 i <- 3;
    u8 b <- msg.data[i];
    global.use_byte(b);
  }

  // BUG: msg is a local → generates (msg.data >> i) & 1
  public void from_local() {
    msg_t msg;
    global.fill_msg(msg);
    u8 i <- 3;
    u8 b <- msg.data[i];
    global.use_byte(b);
  }
}
\`\`\`

Run: `cnext test.cnx`

## Expected output (test.c)

Both functions should generate array access:
```c
// from_param — correct
uint8_t b = msg->data[i];

// from_local — should also be array access
uint8_t b = msg.data[i];
```

## Actual output (test.c)

```c
// from_param — correct
uint8_t b = msg->data[i];

// from_local — WRONG: bit extraction
uint8_t b = (uint8_t)((msg.data >> i) & 1);
```

## Affects

Both C and C++ output modes. Tested on cnext v0.2.15.

## Workaround

Pass the struct to a helper function where it becomes a parameter:
```c-next
void helper(msg_t msg) {
    u8 i <- 3;
    u8 b <- msg.data[i];  // correct — msg is a parameter
}
```

## Discovered in

[...]

PR #982, opened 2026-02-28
fix: parse macro-sized array dimensions in C struct fields (#981)

jlaustill:
## Summary

- Fixed C struct fields with macro-sized arrays (e.g., `uint8_t data[BUF_SIZE]`) incorrectly generating bit extraction code instead of array access
- Updated `parseArrayDimensions()` to capture macro names as strings alongside numeric dimensions
- Extended array dimension types to `(number | string)[]` across C and C++ symbol resolution

## Test plan

- [x] Added regression test in `tests/bugs/issue-981-macro-array-local/`
- [x] All 5576 unit tests pass
- [x] TypeScript compiles without errors

### DIFF
--- a/src/transpiler/logic/symbols/SymbolTable.ts
+++ b/src/transpiler/logic/symbols/SymbolTable.ts
@@ -772,14 +772,7 @@ class SymbolTable {
   // ========================================================================
 
   /**
-   * Add struct field information
-   * @param structName Name of the struct
-   * @param fieldName Name of the field
-   * @param fieldType Type of the field (e.g., "uint32_t")
-   * @param arrayDimensions Optional array dimensions if field is an array
-   */
-  /**
-   * Add struct field information
+   * Add struct field information.
    * Issue #981: Accept (number | string)[] for arrayDimensions to support macro-sized arrays.
    * @param structName Name of the struct
    * @param fieldName Name of the field


src/transpiler/logic/symbols/SymbolTable.ts
```diff
@@ -753,21 +778,30 @@ class SymbolTable {
...
    * @param arrayDimensions Optional array dimensions if field is an array
    */
+  /**
+   * Add struct field information
+   * Issue #981: Accept (number | string)[] for arrayDimensions to support macro-sized arrays.
+   * @param structName Name of the struct
+   * @param fieldName Name of the field
+   * @param fieldType Type of the field (e.g., "uint32_t")
+   * @param arrayDimensions Optional array dimensions - numbers for resolved, strings for macros
+   */
```
jlaustill: **[moderate]** quality: Duplicate JSDoc comment block.

There are two consecutive JSDoc comments before `addStructField` (lines 775-780 and 781-788). Only the second one is used by TypeScript, making the first one dead documentation.

```suggestion
  /**
   * Add struct field information.
   * Issue #981: Accept (number | string)[] for arrayDimensions to support macro-sized arrays.
   * @param structName Name of the struct
   * @param fieldName Name of the field
   * @param fieldType Type of the field (e.g., "uint32_t")
   * @param arrayDimensions Optional array dimensions - numbers for resolved, strings for macros
   */
```

The old JSDoc at lines 750-755 should be removed, keeping only this updated version.

PR #987, opened 2026-03-02
fix: array params no longer get auto-const per ADR-006

### LINKED ISSUE #986: Array params transpile as const, inconsistent with ADR-006 mutable pass-by-reference
## Summary

ADR-006 says all params are pass-by-reference as mutable. Scalar params correctly transpile as mutable pointers (`uint8_t*`), but array params transpile as `const` (`const uint8_t data[8]`). This prevents passing array params to C APIs that expect mutable pointers.

## Reproduction

**test.cnx:**
```c-next
#include "fake_lib.h"

scope Test {
  void handle(u8[8] data) {
    global.process_data(data);  // C API expects mutable uint8_t*
  }
}
```

**fake_lib.h:**
```c
void process_data(uint8_t data[8]);
```

```bash
cnext test.cnx
```

**Generated test.c:**
```c
void Test_handle(const uint8_t data[8]) {  // BUG: const
    process_data(data);  // error: const uint8_t* to uint8_t* conversion
}
```

## Expected

```c
void Test_handle(uint8_t data[8]) {  // mutable, consistent with scalars
    process_data(data);
}
```

## Comparison: scalar params ARE mutable

```c-next
scope Test {
  void handle(u8 val) { val <- 42; }
}
```

Generates:
```c
void Test_handle(uint8_t* val) {  // mutable — correct per ADR-006
    (*val) = 42U;
}
```

Scalar `u8 val` → `uint8_t*` (mutable). Array `u8[8] data` → `const uint8_t data[8]` (const). Inconsistent.

## Context

Found in ogauge: `J1939_setData(msg, data)` expects `uint8_t*` but C-Next transpiles `u8[8] data` param as `const uint8_t*`, causing a C++ compile error.

## Repro files

`/tmp/cnext-bugs/const-array-param/`

jlaustill:
## Summary

Fixes #986: Array parameters were incorrectly getting `const` applied when passed to C APIs expecting mutable pointers.

**Problem:**
- ADR-006 specifies all params are pass-by-reference as mutable
- Scalar params correctly transpiled as mutable pointers (`uint8_t*`)
- Array params incorrectly got `const` when not directly modified in the function body
- This broke C API compatibility:

```c-next
void handle(u8[8] data) {
  global.process_data(data);  // C API expects mutable uint8_t*
}
```

Generated: `void handle(const uint8_t data[8])` - compile error
Expected: `void handle(uint8_t data[8])` - correct

**Fix:**
Arrays now never get auto-const. Only explicit const from source code (`const u8[8] data`) adds the qualifier.

## Changes

- `ParameterInputAdapter`: Set `isAutoConst=false` for all array params
- `Transpiler`: Remove arrays from auto-const logic in header generation
- Add unit test for new array auto-const behavior
- Add regression test in `bugs/issue-986-const-array-param/`
- Update affected test snapshots (arrays that were read-only now don't have const)

## Test plan

- [x] New regression test compiles and executes successfully
- [x] All 952 integration tests pass
- [x] All 5579 unit tests pass
- [x] Build passes with no type errors


tests/const-params/auto-const-pointer.expected.c
```diff
@@ -28,8 +28,9 @@ void arrayMod(uint8_t arr[4]) {
     arr[0] = 99U;
 }
 
-// Array parameter read-only
-uint8_t arrayReadOnly(const uint8_t arr[4]) {
+// Array parameter read-only - arrays never get auto-const per ADR-006
+// (arrays are mutable by default; use explicit const if needed)
```
jlaustill, line 32: **[important]** design: The `.expected.c` and `.expected.cpp` snapshots were correctly updated here to remove `const` from `arrayReadOnly`, but the corresponding `auto-const-pointer.expected.h` file was **not** updated in this PR.

On `main`, that file still contains:
```c
uint8_t arrayReadOnly(const uint8_t arr[4]);
```

Since `Transpiler.convertToHeaderSymbols()` no longer sets `isAutoConst: true` for array params, the generated header should now be:
```c
uint8_t arrayReadOnly(uint8_t arr[4]);
```

This is the only `.expected.h` file in the repo affected (I checked all others — none have `const` on array params in header prototypes). If the test framework compares `.expected.h` files, this would cause a test failure. If it doesn't, the file is still stale and misleading.

**Fix:** Update `tests/const-params/auto-const-pointer.expected.h` line 21:
```suggestion
uint8_t arrayReadOnly(uint8_t arr[4]);
```

jlaustill (commented):
## Review Summary

**PR:** #987 — fix: array params no longer get auto-const per ADR-006

### Overview

Clean, well-scoped bug fix that correctly removes auto-const inference from array parameters. The change aligns with ADR-006's principle that all params are pass-by-reference and mutable. The fix addresses both code paths (CodeGenerator via `ParameterInputAdapter` and header generation via `Transpiler.convertToHeaderSymbols`), preventing `const` from being applied to array params that are only read — which broke compatibility with C APIs expecting mutable pointers.

### Findings

| Severity | Count |
|----------|-------|
| Critical | 0 |
| Important | 1 |
| Moderate | 0 |
| Nit | 0 |

### Key Issues

1. **[important]** The `tests/const-params/auto-const-pointer.expected.h` file was NOT updated in this PR, but it contains `arrayReadOnly(const uint8_t arr[4])` which should now be `arrayReadOnly(uint8_t arr[4])` per the behavioral change. See inline comment.

### What Looks Good

- **Both code paths fixed**: The `ParameterInputAdapter._buildArrayInputFromAST()` and `Transpiler.convertToHeaderSymbols()` changes together eliminate all paths where arrays could get auto-const. The cross-reference agent confirmed no other paths exist.
- **Excellent regression test**: The `fake_lib.c`/`.h` approach directly exercises the real-world scenario (passing an array to a C API that mutates it), with execution validation via return codes.
- **Minimal, focused change**: The actual logic changes are ~6 lines across 2 files, with the rest being consistent snapshot updates. No scope creep.
- **Good comment hygiene**: Updated the `auto-const-pointer.test.cnx` comments to explain the new behavior, and the `ParameterSignatureBuilder.test.ts` comments clarify why existing builder tests still pass.

---

PR #1000, opened 2026-06-15
fix: add volatile modifier to atomic scope variables (#998)

### LINKED ISSUE #998: Scope-level `atomic` variable definition drops `volatile` (global `atomic` keeps it)
**Version:** cnext v0.2.16
**Related:** #468 (closed — was about `extern` decls in the *header*; this is the variable *definition* in the `.cpp`), #997 (use case)

## Summary

An `atomic` variable declared **inside a scope** generates a plain (non-`volatile`) definition, while the **same** `atomic` declaration at **global** scope correctly generates `volatile`. Since `atomic` exists for ISR safety, the missing `volatile` on scope atomics is an ISR-visibility hazard (the compiler may cache the value across an ISR/main-loop boundary).

## Minimal repro

`atomic_scope_volatile.cnx`:
```cnx
atomic u8 globalCounter <- 0;

scope Foo {
    atomic u8 scopeCounter <- 0;

    void bump() {
        this.scopeCounter <- 1;
    }
}
```

Transpile (`cppRequired: true`):
```
cnext atomic_scope_volatile.cnx
```

## Actual generated output
```cpp
volatile uint8_t globalCounter = 0U;     // ✓ global atomic -> volatile
static uint8_t   Foo_scopeCounter = 0U;  // ✗ scope atomic  -> NO volatile
```

## Expected
```cpp
static volatile uint8_t Foo_scopeCounter = 0U;   // scope atomic should also be volatile
```

## Note: the existing snapshot test enshrines the bug
`tests/scope/scope-atomic-modifier.expected.cpp` currently expects `static uint8_t AtomicTest_counterU8 = 0U;` (no `volatile`) for `atomic u8 counterU8` — so the snapshot suite passes despite the missing qualifier. That expected output should gain `volatile` as part of the fix.

## Real-world impact
Hit while building an ISR-safe SPSC CAN-TX ring buffer (`scope CanTxQueue` with `atomic u8 head/tail`) for the OVGT turbo controller — the timer ISR produces, the main loop consumes. Without `volatile` on the shared indices, the producer/consumer handoff isn't guaranteed visible across the ISR boundary. (De-facto mitigated today only because function-call boundaries force reloads.)

jlaustill:
## Summary

- Fixes #998: Scope-level `atomic` variable definition drops `volatile` (global `atomic` keeps it)

Scope-level atomic variables were missing the `volatile` qualifier in generated C/C++ code, while global atomic variables correctly included it. Since `atomic` exists for ISR safety, the missing `volatile` on scope atomics was an ISR-visibility hazard where the compiler could cache values across ISR/main-loop boundaries.

## Changes

- Add `atomicModifier()` check in `ScopeGenerator.generateRegularVariable()`
- Generate `"volatile "` prefix when atomic modifier is present
- Update unit test mocks to include `atomicModifier` method
- Add 2 unit tests for atomic scope variable generation
- Update expected outputs in 4 test files to include `volatile`

## Before/After

**Before (incorrect):**
```cpp
static uint8_t Foo_scopeCounter = 0U;  // scope atomic → NO volatile
```

**After (correct):**
```cpp
static volatile uint8_t Foo_scopeCounter = 0U;  // scope atomic → volatile ✓
```

## Test plan

- [x] All 958 integration tests pass
- [x] All 5615 unit tests pass (including 2 new tests)
- [x] Verified with exact repro from issue
- [x] Pre-push quality checks pass

### DIFF
--- a/src/transpiler/output/codegen/generators/declarationGenerators/ScopeGenerator.ts
+++ b/src/transpiler/output/codegen/generators/declarationGenerators/ScopeGenerator.ts
@@ -26,6 +26,7 @@ import ArrayDimensionUtils from "./ArrayDimensionUtils";
 import QualifiedNameGenerator from "../../utils/QualifiedNameGenerator";
 import CodeGenState from "../../../../state/CodeGenState";
 import ScopeUtils from "../../../../../utils/ScopeUtils";
+import VariableModifierBuilder from "../../helpers/VariableModifierBuilder";
 
 /**
  * Generate initializer expression for a variable declaration.
@@ -206,15 +207,23 @@ function generateRegularVariable(
     orchestrator.markOpaqueScopeVariable(fullName);
   }
 
-  // Issue #282: Add 'const' modifier for const variables
+  // Issue #998: Use VariableModifierBuilder for consistent modifier handling
+  // This handles const, atomic, volatile, and validates mutual exclusion
+  // Scope variables are always at file scope (not in function body), no initializer for modifier purposes
+  const modifiers = VariableModifierBuilder.build(
+    varDecl,
+    false, // inFunctionBody - scope vars are file scope
+    false, // hasInitializer - doesn't affect volatile/atomic handling
+    false, // cppMode - doesn't affect volatile/atomic handling
+  );
+  // For scope variables: static for private, no modifier for public
+  // Then add volatile (from atomic or volatile keyword), then const
+  const staticPrefix = isPrivate ? "static " : "";
+  const volatilePrefix = modifiers.atomic || modifiers.volatile;
   const constPrefix = isConst ? "const " : "";
-  // Issue #998: Add 'volatile' modifier for atomic variables (ISR-safety)
-  const isAtomic = varDecl.atomicModifier() !== null;
-  const atomicPrefix = isAtomic ? "volatile " : "";
-  const prefix = isPrivate ? "static " : "";
 
   // Build declaration with all dimensions
-  let decl = `${prefix}${atomicPrefix}${constPrefix}${type} ${fullName}`;
+  let decl = `${staticPrefix}${volatilePrefix}${constPrefix}${type} ${fullName}`;
   decl += ArrayDimensionUtils.generateArrayTypeDimension(
     arrayTypeCtx,
     orchestrator,
--- a/src/transpiler/output/codegen/generators/declarationGenerators/ScopeGenerator.ts
+++ b/src/transpiler/output/codegen/generators/declarationGenerators/ScopeGenerator.ts
@@ -208,10 +208,13 @@ function generateRegularVariable(
 
   // Issue #282: Add 'const' modifier for const variables
   const constPrefix = isConst ? "const " : "";
+  // Issue #998: Add 'volatile' modifier for atomic variables (ISR-safety)
+  const isAtomic = varDecl.atomicModifier() !== null;
+  const atomicPrefix = isAtomic ? "volatile " : "";
   const prefix = isPrivate ? "static " : "";
 
   // Build declaration with all dimensions
-  let decl = `${prefix}${constPrefix}${type} ${fullName}`;
+  let decl = `${prefix}${atomicPrefix}${constPrefix}${type} ${fullName}`;
   decl += ArrayDimensionUtils.generateArrayTypeDimension(
     arrayTypeCtx,
     orchestrator,
--- a/src/transpiler/output/codegen/generators/declarationGenerators/__tests__/ScopeGenerator.test.ts
+++ b/src/transpiler/output/codegen/generators/declarationGenerators/__tests__/ScopeGenerator.test.ts
@@ -127,6 +127,7 @@ function createMockVariableDecl(options: {
   type: string;
   isConst?: boolean;
   isAtomic?: boolean;
+  isVolatile?: boolean;
   initialValue?: string;
   arrayDims?: string[];
   hasStringType?: boolean;
@@ -143,8 +144,10 @@ function createMockVariableDecl(options: {
         options.arrayTypeSize,
       ),
     constModifier: () => createMockConstModifier(options.isConst ?? false),
-    // Issue #998: atomic modifier for scope variables
+    // Issue #998: atomic and volatile modifiers for scope variables
+    // Uses VariableModifierBuilder for consistent handling
     atomicModifier: () => (options.isAtomic ? ({} as unknown) : null),
+    volatileModifier: () => (options.isVolatile ? ({} as unknown) : null),
     expression: () =>
       options.initialValue ? createMockExpression(options.initialValue) : null,
     arrayDimension: () =>
@@ -816,6 +819,44 @@ describe("ScopeGenerator", () => {
 
       expect(result.code).toContain("volatile bool Shared_flag = false;");
       expect(result.code).not.toContain("static volatile bool Shared_flag");
+    });
+
+    it("generates volatile-keyword variable with volatile modifier (Issue #998)", () => {
+      const varDecl = createMockVariableDecl({
+        name: "reg",
+        type: "u8",
+        isVolatile: true,
+        initialValue: "0",
+      });
+      const member = createMockScopeMember({ variableDecl: varDecl });
+      const ctx = createMockScopeContext("HW", [member]);
+      const input = createMockInput();
+      const state = createMockState();
+      const orchestrator = createMockOrchestrator();
+
+      const result = generateScope(ctx, input, state, orchestrator);
+
+      expect(result.code).toContain("static volatile uint8_t HW_reg = 0;");
+    });
+
+    it("throws error when both atomic and volatile are specified (Issue #998)", () => {
+      const varDecl = createMockVariableDecl({
+        name: "bad",
+        type: "u8",
+        isAtomic: true,
+        isVolatile: true,
+        initialValue: "0",
+        startLine: 10,
+      });
+      const member = createMockScopeMember({ variableDecl: varDecl });
+      const ctx = createMockScopeContext("Test", [member]);
+      const input = createMockInput();
+      const state = createMockState();
+      const orchestrator = createMockOrchestrator();
+
+      expect(() => generateScope(ctx, input, state, orchestrator)).toThrow(
+        /Cannot use both 'atomic' and 'volatile' modifiers/,
+      );
     });
   });
 
--- a/src/transpiler/output/codegen/generators/declarationGenerators/__tests__/ScopeGenerator.test.ts
+++ b/src/transpiler/output/codegen/generators/declarationGenerators/__tests__/ScopeGenerator.test.ts
@@ -126,6 +126,7 @@ function createMockVariableDecl(options: {
   name: string;
   type: string;
   isConst?: boolean;
+  isAtomic?: boolean;
   initialValue?: string;
   arrayDims?: string[];
   hasStringType?: boolean;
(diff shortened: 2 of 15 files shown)


src/transpiler/output/codegen/generators/declarationGenerators/ScopeGenerator.ts
```diff
@@ -208,10 +208,13 @@ function generateRegularVariable(
 
   // Issue #282: Add 'const' modifier for const variables
   const constPrefix = isConst ? "const " : "";
+  // Issue #998: Add 'volatile' modifier for atomic variables (ISR-safety)
+  const isAtomic = varDecl.atomicModifier() !== null;
```
jlaustill: **[important]** design: This re-implements modifier logic that already lives in `VariableModifierBuilder` (`src/transpiler/output/codegen/helpers/VariableModifierBuilder.ts`). That helper was introduced in Issue #696 *to eliminate this exact duplication* across `generateVariableDecl`, `generateParameter`, and `ControlFlowGenerator.generateForVarDecl`. Per the project's "No Duplicate Code Paths — ZERO EXCEPTIONS" rule, the scope path should reuse it rather than fork the logic.

**[important]** bug: Because the check only looks at `atomicModifier()`, a scope variable declared with the **`volatile`** keyword still loses its qualifier. I verified this against the transpiler:

```cnx
scope Foo {
    public volatile u8 reg <- 0;   // → `uint8_t Foo_reg = 0U;`   (volatile DROPPED, both before and after this fix)
    public atomic   u8 counter <- 0; // → `volatile uint8_t Foo_counter = 0U;` (fixed by this PR)
}
```

The global path emits `volatile` for *both* keywords, so this leaves the identical scope/global inconsistency #998 reports — just for `volatile`. Reusing `VariableModifierBuilder.build(...)` + `toPrefix(...)` resolves both keywords here and also enforces the atomic+volatile mutual-exclusion check, which the inline version silently skips. For example:

```ts
const modifiers = VariableModifierBuilder.build(varDecl, /*inFunctionBody*/ false, hasInitializer, orchestrator.cppMode);
// then fold modifiers into the prefix instead of constPrefix/atomicPrefix
```

If you'd rather not widen the scope of this PR, the `volatile`-keyword case should at least be tracked as its own issue (per the Bug Tracking rule) rather than left silently broken.

src/transpiler/output/codegen/generators/declarationGenerators/ScopeGenerator.ts
```diff
@@ -208,10 +208,13 @@ function generateRegularVariable(
...
+  const atomicPrefix = isAtomic ? "volatile " : "";
   const prefix = isPrivate ? "static " : "";
 
   // Build declaration with all dimensions
-  let decl = `${prefix}${constPrefix}${type} ${fullName}`;
+  let decl = `${prefix}${atomicPrefix}${constPrefix}${type} ${fullName}`;
```
jlaustill: **[moderate]** quality: `${prefix}${atomicPrefix}${constPrefix}` emits `volatile const`, but `VariableModifierBuilder.toPrefix()` orders it `const volatile` (`${extern}${const}${atomic}${volatile}`). Both are valid C, but the codebase now has two orderings for the same construct. Reusing the shared builder (see the comment above) makes the ordering consistent with the global path automatically — another reason to route through it rather than assemble the prefix by hand here.

jlaustill (commented):
## Review Summary

**PR:** #1000 — fix: add volatile modifier to atomic scope variables (#998)

### Overview

Correctly fixes the reported #998 symptom — scope-level `atomic` variables now emit `volatile`, matching the global path — and the snapshot updates are complete and accurate (verified `enum-all-scenarios` has no `.test.cpp`, `arduino-portable` is `test-c-only`, so nothing was left stale). The two concerns below are about *how* the fix is wired in: it reimplements modifier logic inline instead of reusing the existing `VariableModifierBuilder`, and as a direct consequence it leaves a sibling case (`volatile`-keyword scope variables) still broken.

### Findings

| Severity | Count |
|----------|-------|
| Critical | 0 |
| Important | 2 |
| Moderate | 1 |
| Nit | 0 |

### Key Issues

- **[important] design — duplicate code path:** The fix hand-rolls `atomicModifier() ? "volatile "` inline, but `VariableModifierBuilder` (created in Issue #696 *specifically to eliminate this exact duplication* across `generateVariableDecl`/`generateParameter`/`generateForVarDecl`) already owns this logic. Project CLAUDE.md calls duplicate code paths "the WORST anti-pattern in this project."
- **[important] bug — `volatile`-keyword scope variables are still dropped:** I verified that a scope variable declared with the `volatile` keyword (e.g. `public volatile u8 reg`) generates a **non-volatile** definition both before *and after* this fix, because the new check only looks at `atomicModifier()`, never `volatileModifier()`. The global path emits `volatile` for both keywords, so this leaves the exact same scope/global inconsistency #998 describes — just for `volatile` instead of `atomic`. Routing through `VariableModifierBuilder` would fix both at once (and add its atomic+volatile mutual-exclusion validation, which the inline path skips).

### What Looks Good

- The root-cause analysis in the PR/issue is accurate, and the before/after is exactly right for the `atomic` case.
- Snapshot/expected-output updates are thorough and consistent across C and C++ variants — and correctly omitted where a `.test.cpp` doesn't exist or the test is `test-c-only`.
- Two focused unit tests cover both the `static` (private) and non-`static` (public) atomic cases.

---

jlaustill:
Addressed review feedback with commit ef929db4:

**Finding 1: Duplicate code path** — Refactored to use `VariableModifierBuilder` instead of inline modifier logic. This aligns with the CLAUDE.md guidance against duplicate code paths.

**Finding 2: volatile-keyword scope variables** — Fixed. Now both `atomic` and `volatile` keywords correctly generate `volatile` in output. Added validation that rejects using both modifiers together (via `VariableModifierBuilder.build()`).

Added 2 new unit tests:
- `volatile`-keyword scope variables generate correctly
- `atomic` + `volatile` together throws the expected error

All 958 integration tests and 5617 unit tests pass.

jlaustill (commented):
## Review Summary

**PR:** #1000 — fix: add volatile modifier to atomic scope variables (#998)

### Overview
The fix is correct and well-targeted: scope-level `atomic` variables now emit `volatile`, closing the real ISR-visibility hazard described in #998. Routing modifier handling through `VariableModifierBuilder` is the right call — it reuses the existing validation (atomic/volatile mutual exclusion) instead of reinventing it, and the snapshot updates correctly un-enshrine the bug. CI is fully green. My comments below are about path consistency and minor cleanup, not correctness of the core fix.

### Findings

| Severity | Count |
|----------|-------|
| Critical | 0 |
| Important | 0 |
| Moderate | 1 |
| Nit | 2 |

### Key Issues
- **Moderate (design/consistency):** The scope path manually concatenates `static volatile const`, whereas the global path's `VariableModifierBuilder.toPrefix()` emits `const` before `volatile`. For a `volatile const` scope variable (a legitimate read-only-hardware-register case the grammar allows), the two paths now produce different qualifier ordering — exactly the kind of dual-path divergence the project guards against. Functionally identical in C, but worth aligning. See inline comment.
- **Nit:** `modifiers.const` and `modifiers.extern` are computed by `build()` but discarded; `isConst`/`constPrefix` re-derives const-ness that `modifiers.const` already holds.
- **Nit / please document:** This PR also makes the bare `volatile` keyword take effect on scope variables (it was silently dropped before) — a good latent-bug fix beyond the atomic title, and it's tested. Worth a line in the PR description so it's not a surprise.

### What Looks Good
- Reuses `VariableModifierBuilder` rather than duplicating modifier/validation logic — aligns with the project's no-duplicate-paths rule.
- Snapshot updates (`.expected.c`/`.cpp` + committed `.test.c`/`.test.cpp`) correctly replace the bug-enshrining expectations, matching the issue's note about the stale snapshot.
- New unit tests cover private, public, `volatile`-keyword, and the atomic+volatile error path.

---

src/transpiler/output/codegen/generators/declarationGenerators/ScopeGenerator.ts
```diff
@@ -206,12 +207,23 @@ function generateRegularVariable(
...
+  // For scope variables: static for private, no modifier for public
+  // Then add volatile (from atomic or volatile keyword), then const
+  const staticPrefix = isPrivate ? "static " : "";
+  const volatilePrefix = modifiers.atomic || modifiers.volatile;
   const constPrefix = isConst ? "const " : "";
-  const prefix = isPrivate ? "static " : "";
 
   // Build declaration with all dimensions
-  let decl = `${prefix}${constPrefix}${type} ${fullName}`;
+  let decl = `${staticPrefix}${volatilePrefix}${constPrefix}${type} ${fullName}`;
```
jlaustill, line 226: **[moderate]** design: This concatenates qualifiers as `static volatile const`, but the global/normal-variable path emits them via `VariableModifierBuilder.toPrefix()` which orders them `const volatile` (see `VariableDeclHelper.ts:659` → `toPrefix` = `${extern}${const}${atomic}${volatile}`).

For a scope variable that is both `volatile` and `const` — a legitimate read-only memory-mapped register (`volatile const u8 statusRegister`, which the grammar `atomicModifier? volatileModifier? constModifier?` permits) — the scope path now emits `static volatile const uint8_t` while a global declaration emits `const volatile uint8_t`. Same semantics in C, but it's two code paths producing different output for the same construct, which this project explicitly tries to avoid.

Simplest alignment is to flip the order so const precedes volatile, matching `toPrefix`:

```suggestion
  let decl = `${staticPrefix}${constPrefix}${volatilePrefix}${type} ${fullName}`;
```

src/transpiler/output/codegen/generators/declarationGenerators/ScopeGenerator.ts
```diff
@@ -206,12 +207,23 @@ function generateRegularVariable(
     orchestrator.markOpaqueScopeVariable(fullName);
   }
 
-  // Issue #282: Add 'const' modifier for const variables
+  // Issue #998: Use VariableModifierBuilder for consistent modifier handling
+  // This handles const, atomic, volatile, and validates mutual exclusion
+  // Scope variables are always at file scope (not in function body), no initializer for modifier purposes
+  const modifiers = VariableModifierBuilder.build(
```
jlaustill, line 213: **[nit]** quality: `build()` returns `const` and `extern` too, but both are discarded here — and `constPrefix` below re-derives const-ness from `isConst` (line 187, `varDecl.constModifier() !== null`), duplicating `modifiers.const`. You can drop the separate `isConst`/`constPrefix` and use `modifiers.const` directly so there's a single source of truth for the const qualifier. (`extern` is correctly irrelevant for scope vars since they're `static`/file-scope, so ignoring it is fine — a one-line note saying so would help the next reader.)

src/transpiler/output/codegen/generators/declarationGenerators/ScopeGenerator.ts
```diff
@@ -206,12 +207,23 @@ function generateRegularVariable(
...
+    varDecl,
+    false, // inFunctionBody - scope vars are file scope
+    false, // hasInitializer - doesn't affect volatile/atomic handling
+    false, // cppMode - doesn't affect volatile/atomic handling
+  );
+  // For scope variables: static for private, no modifier for public
+  // Then add volatile (from atomic or volatile keyword), then const
+  const staticPrefix = isPrivate ? "static " : "";
+  const volatilePrefix = modifiers.atomic || modifiers.volatile;
```
jlaustill, line 222: **[nit]** Heads up that this line also changes behavior for the bare `volatile` keyword on scope variables, not just `atomic`: the previous code only handled `static`/`const`, so a scope-level `volatile u8 reg` was silently emitted as `static uint8_t reg` (no `volatile`). After this change it correctly becomes `static volatile uint8_t reg`. That's a genuine additional fix (and you've added a test for it — nice), but it's not mentioned in the PR summary, which only describes the atomic case. Worth calling out in the description so reviewers know the `volatile`-keyword behavior changed too.

Issue #914, opened 2026-02-23
bug(codegen): header prototype missing pointer on primitive callback params


jlaustill:
## Summary

When a scope function matches a callback typedef, the `.c` implementation correctly generates pointer signatures (e.g., `uint8_t* buf`), but the `.h` header prototype generates the parameter as a value (e.g., `uint8_t buf`). This header/implementation mismatch is undefined behavior in C and will cause compile errors or silent corruption when another translation unit includes the header.

**cnext version:** v0.2.6

### `widget.h`

```c
#ifndef WIDGET_H
#define WIDGET_H

#include <stdint.h>

typedef struct widget_t widget_t;

typedef struct {
    int32_t x1;
    int32_t y1;
} rect_t;

typedef void (*flush_cb_t)(widget_t *w, const rect_t *area, uint8_t *buf);

widget_t *widget_create(void);
void widget_set_flush_cb(widget_t *w, flush_cb_t cb);
void widget_flush_ready(widget_t *w);

#endif
```

### `lib.cnx`

```cnx
#include "widget.h"

scope Renderer {
    public void flush(widget_t w, const rect_t area, u8 buf) {
        global.widget_flush_ready(w);
    }

    public void init() {
        widget_t w <- global.widget_create();
        global.widget_set_flush_cb(w, this.flush);
    }
}
```

### `test.cnx`

```cnx
#include "widget.h"
#include "lib.cnx"

i32 main() {
    Renderer.init();
    return 0;
}
```

### Run

```bash
cnext test.cnx
```

### Expected `lib.h`

```c
void Renderer_flush(widget_t* w, const rect_t* area, uint8_t* buf);
```

### Actual `lib.h`

```c
void Renderer_flush(widget_t* w, const rect_t* area, uint8_t buf);
//                                                    ^^^^^^^^
//                        missing pointer — doesn't match implementation
```

### Actual `lib.c` (correct)

```c
void Renderer_flush(widget_t* w, const rect_t* area, uint8_t* buf) {
//                                                    ^^^^^^^^^
//                               correctly matches flush_cb_t typedef
```

## Analysis

The callback-aware signature fix from PR #897 and subsequent PRs correctly transforms the `.c` implementation to match the callback typedef. However, the `.h` header generation doesn't apply the same transform to primitive params like `uint8_t`.

Note: `widget_t*` and `const rect_t*` are correct in both header and implementation — only primitive types (`uint8_t`) are affected.

## Impact

[...]

PR #1001, opened 2026-06-16
fix: skip auto-const for opaque handle function parameters

### LINKED ISSUE #995: Opaque-handle function parameters generate `const T*`, breaking calls to mutating C APIs
## Summary

When an opaque handle type (an incomplete `struct`, e.g. LVGL's `lv_obj_t`) is passed as a **C-Next function parameter**, the transpiler emits the parameter as `const T*`. This breaks handing the handle to any C API that takes a non-const `T*` — which is the normal case for these libraries (`lv_obj_align`, `lv_obj_set_style_*`, etc.). Opaque-handle parameters should be generated as non-const `T*`.

Discovered factoring shared LVGL widget styling into a helper in a real project (ogauge). It forced inlining the styling at every call site because no helper can accept an `lv_obj_t`.

**C-Next version:** v0.2.16

## Minimal repro

`fake_lib.h`:
```c
#ifndef FAKE_LIB_H
#define FAKE_LIB_H
/* Opaque handle type + a function that MUTATES through a non-const pointer. */
typedef struct _widget widget_t;
widget_t* widget_create(void);
void widget_move(widget_t* w, int x);   /* non-const widget_t* */
#endif
```

`test.cnx`:
```
#include "fake_lib.h"

scope UI {
  // Pass an opaque handle as a parameter, then hand it to a mutating C API.
  private void reposition(widget_t w, i32 x) {
    global.widget_move(w, x);
  }

  void run() {
    widget_t w <- global.widget_create();
    this.reposition(w, 5);
  }
}
```

Run: `cnext test.cnx`

## Expected (generated `test.c`)

```c
static void UI_reposition(widget_t* w, int32_t x) {
    widget_move(w, x);
}
```

## Actual (generated `test.c`)

```c
static void UI_reposition(const widget_t* w, int32_t x) {
    widget_move(w, x);
}
```

The parameter is `const widget_t*`, but it's passed straight to `widget_move(widget_t*)`.

## Compiler result

`gcc -c test.c`:
```
warning: passing argument 1 of 'widget_move' discards 'const' qualifier from pointer target type [-Wdiscarded-qualifiers]
note: expected 'widget_t *' but argument is of type 'const widget_t *'
```

In C-Next's **C++ output mode** (what real projects build, generating `.cpp`) this is a hard error:
`invalid conversion from 'const lv_obj_t*' to 'lv_obj_t*' [-fpermissive]`.

## Impact

Opaque handles can't be passed to helper functions at all if those helpers then call mutating C APIs — every such call must be inlined. Common for LVGL UI code.

jlaustill:
## Summary
- Fixes #995: Opaque-handle function parameters generate `const T*`, breaking calls to mutating C APIs

## Problem
When an opaque handle type (an incomplete struct typedef like LVGL's `lv_obj_t`) is passed as a C-Next function parameter, the transpiler incorrectly emitted `const T*` instead of `T*`. This broke calls to C APIs that expect non-const pointers.

**Before:**
```c
static void UI_reposition(const widget_t* w, int32_t x) {
    widget_move(w, x);  // ERROR: discards 'const' qualifier
}
```

**After:**
```c
static void UI_reposition(widget_t* w, int32_t x) {
    widget_move(w, x);  // OK
}
```

## Root Cause
`ParameterInputAdapter.fromAST()` computed `isAutoConst = true` for all unmodified non-callback params without checking if the type was an opaque handle.

## Changes
- Add `isOpaqueType?` dependency to `IFromASTDeps` interface in `ParameterInputAdapter.ts`
- Exclude opaque types from auto-const logic (same reasoning as arrays per ADR-006)
- Pass `CodeGenState.isOpaqueType` callback to adapter from `CodeGenerator.generateParameter()`
- Add 3 unit tests for opaque type parameter handling
- Add integration test with mock C library that exercises the fix

## Test plan
- [x] New failing test created first (TDD)
- [x] Fix applied and test passes
- [x] All 959 integration tests pass
- [x] All 5620 unit tests pass
- [x] Existing opaque-related tests (issue-895) still pass


jlaustill (commented):
## Review Summary

**PR:** #1001 — fix: skip auto-const for opaque handle function parameters

### Overview
This PR correctly fixes issue #995 by excluding opaque handle types from auto-const logic in `ParameterInputAdapter.fromAST()`. The fix is well-motivated, well-tested with both unit and integration tests, and follows TDD practices. However, there's a potential dual code path issue that should be addressed.

### Findings

| Severity | Count |
|----------|-------|
| Critical | 0 |
| Important | 1 |
| Moderate | 0 |
| Nit | 0 |

### Key Issues
- **Dual code path**: `Transpiler.convertToHeaderSymbols()` has its own auto-const logic that doesn't check for opaque types. While the current test uses a private function (not exported to header), this creates a latent bug for public functions with opaque parameters.

### What Looks Good
- Clean fix targeting the root cause in `ParameterInputAdapter.fromAST()`
- Comprehensive unit tests covering positive and negative cases
- Full integration test with a real C library exercising the fix
- Good documentation with issue references in comments

---

src/transpiler/output/codegen/helpers/ParameterInputAdapter.ts
```diff
@@ -142,10 +149,14 @@ class ParameterInputAdapter {
     const isKnownPrimitive = !!deps.typeMap[typeName];
     // Issue #958: C-header typedef struct types need pointer semantics
     const isTypedefStruct = deps.isTypedefStructType(typeName);
+    // Issue #995: Opaque handles (incomplete struct typedefs) should not get auto-const
+    const isOpaque = deps.isOpaqueType?.(typeName) ?? false;
     // Issue #895: Don't add auto-const for callback-compatible functions
     // because it would change the signature and break typedef compatibility
+    // Issue #995: Don't add auto-const for opaque handles because they must be
```
jlaustill: **[important]** design: This fix adds the opaque type check to `ParameterInputAdapter.fromAST()`, which is used for `.c` file generation. However, there's a **parallel code path** in `Transpiler.convertToHeaderSymbols()` (lines 1520-1541) that also applies auto-const logic without the opaque type check.

The test case uses a `private void reposition(widget_t w, ...)` function, which won't appear in the header. But if someone writes a **public** function with an opaque handle parameter, the header generation would incorrectly emit `const widget_t*` even though the `.c` file correctly emits `widget_t*`.

To avoid this dual code path issue, consider also adding the opaque type check to `Transpiler.convertToHeaderSymbols()` at line 1527:

```suggestion
// Issue #995: Opaque handles should not get auto-const
const isOpaque = CodeGenState.isOpaqueType(param.type ?? "");
```

And update line 1535:
```typescript
if (isPointerParam && !isOpaque && unmodified.has(param.name)) {
```

Alternatively, add an integration test with a **public** function that takes an opaque parameter to verify header generation is also correct.

jlaustill: Fixed. Added isOpaque check to Transpiler.convertToHeaderSymbols() and updated the test to use a public function to cover both .c and .h generation paths.

src/transpiler/Transpiler.ts
```diff
@@ -1524,9 +1524,13 @@ class Transpiler {
           // ADR-006: Only non-array pointer params get auto-const.
           // Arrays are pass-by-reference and mutable by default - auto-const would
           // break compatibility with C APIs expecting mutable pointers (issue #986).
+          // Issue #995: Opaque handles should not get auto-const because they must
+          // be passed to C APIs that expect non-const pointers.
+          const isOpaque = CodeGenState.isOpaqueType(param.type ?? "");
```
jlaustill: **[important]** design: This is the second, independent place that computes auto-const eligibility — `ParameterInputAdapter.fromAST()` does the same with a *different* expression (`!isCallbackCompatible && !isModified && !isConst && !isOpaque`), while here it's `!isConst && !isArray && !isOpaque && type !== f32/f64/ISR && !knownEnums`. Having to add the `isOpaque` exclusion in both spots (your two commits) is exactly the dual-code-path anti-pattern `CLAUDE.md` forbids; Issue #914 set the precedent of resolving such values **once** onto `IParameterSymbol`.

Resolving `isAutoConst` once (including the opaque check) onto the parameter symbol would let both the body and header paths read a single field — and would have prevented the C++ header gap flagged on `opaque-param-const.expected.hpp`.

jlaustill (commented):
## Review Summary

**PR:** #1001 — fix: skip auto-const for opaque handle function parameters

### Overview
This correctly fixes #995: opaque-handle parameters no longer get auto-`const`, so generated code can call mutating C APIs. The TDD approach, the linked-library integration test, and the unit coverage of the new dependency are all solid. Two design/correctness concerns below, both rooted in the header generation path being a separate code path from the body path.

### Findings

| Severity | Count |
|----------|-------|
| Critical | 0 |
| Important | 2 |
| Moderate | 0 |
| Nit | 0 |

### Key Issues

1. **Duplicate auto-const logic (dual code path).** The opaque exclusion had to be added in *two* places with *two different* boolean expressions — `ParameterInputAdapter.fromAST()` (body path) and `Transpiler.convertToHeaderSymbols()` (header path). The two commits on this branch are direct evidence: the first fixed the body, the second ("add opaque type check to header generation path") patched the header after the fact. This is the anti-pattern `CLAUDE.md` calls "the WORST anti-pattern in this project," and Issue #914 already set the precedent — resolve the value **once** onto `IParameterSymbol` so both `.c` and `.h`/C++ paths read one field.

2. **C++ header/body signature mismatch.** `opaque-param-const.expected.hpp` declares `void UI_reposition(widget_t& w, int32_t x);` but `opaque-param-const.expected.cpp` defines `void UI_reposition(widget_t* w, int32_t x)`. The C path is consistent (`widget_t*` in both); only the C++ header diverges, because the header path (`fromSymbol`) never applies opaque→pointer syntax the way `fromAST` does. The self-contained execution test masks this (the body's own re-declaration shadows the internal call), but an external C++ consumer including the header would link against the wrong/undefined symbol. This is the concrete consequence of finding #1 — and per the project rule, code that's touched here needs to be correct regardless of whether the gap pre-existed.

### What Looks Good
- Clean TDD: failing test first, then the fix.
- The `fake_lib.c`/`.h` opaque-handle harness with a genuinely non-const mutating API (`widget_move`) is an excellent, realistic reproduction.
- Unit tests cover all three branches of the new dependency (opaque true, non-opaque false, dependency absent).

---

tests/bugs/issue-995-opaque-param-const/opaque-param-const.expected.hpp
```diff
@@ -0,0 +1,25 @@
...
+#include <stdint.h>
+#include <stdbool.h>
+#include "fake_lib.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/* Function prototypes */
+void UI_reposition(widget_t& w, int32_t x);
```
jlaustill: **[important]** bug: The C++ header declares the opaque param as a reference (`widget_t& w`), but `opaque-param-const.expected.cpp` defines it as a pointer (`widget_t* w`). The C snapshots are consistent (`widget_t*` in both `.expected.h` and `.expected.c`); only the C++ header diverges, because the header path (`fromSymbol`) doesn't force pointer syntax for opaque types the way `fromAST` does.

The execution test passes only because the `.cpp` re-declares the function with the pointer signature, so the internal `UI_reposition(w, 42)` call resolves to the pointer overload — the `extern "C"` reference declaration is never linked against. An external C++ TU including this header would call the reference signature and hit an undefined symbol.

```suggestion
void UI_reposition(widget_t* w, int32_t x);
```
The fix belongs in the header generation path (so it's emitted, not just snapshot-edited), ideally via the unified `IParameterSymbol` field from the other comment.

src/transpiler/Transpiler.ts
```diff
@@ -1517,31 +1517,42 @@ class Transpiler {
...
-            return { ...param, isAutoConst: true };
-          }
-          return param;
-        });
-
-        return { ...headerSymbol, parameters: updatedParams };
-      }
+      const updatedParams = headerSymbol.parameters.map((param) => {
+        // Issue #995: Resolve opaque type info ONCE onto the symbol.
+        // This is the single source of truth for both body (.c/.cpp) and header (.h/.hpp).
```
jlaustill, line 1527: **[important]** design: This comment claims `isOpaqueHandle` is "the single source of truth for both body (.c/.cpp) and header (.h/.hpp)", but the body path never reads this field. `CodeGenerator.generateParameter()` → `ParameterInputAdapter.fromAST()` recomputes the opaque consequences from scratch (via the `isOpaqueType` callback); only the header path (`fromSymbol`) consumes `isOpaqueHandle`.

So the *rule* "opaque ⇒ no auto-const + force pointer + pass-by-reference" lives in two adapters, not one. Per the project's no-duplicate-paths rule, changing the rule later would require editing both. The detection (`CodeGenState.isOpaqueType`) is correctly single-sourced — it's the *application* that's duplicated.

Suggested unification: add `isOpaqueHandle` to `IParameterInput`, have both `fromAST` and `fromSymbol` set it as a pure pass-through, and apply the suppress-const/force-pointer rule once inside `ParameterSignatureBuilder` (whose stated contract is already "all decisions pre-computed in the input").

src/transpiler/output/codegen/helpers/ParameterInputAdapter.ts
```diff
@@ -222,19 +233,24 @@ class ParameterInputAdapter {
...
+      // Issue #995: Opaque handles never get auto-const (resolved value may still be true from symbol)
+      isAutoConst: isOpaqueHandle ? false : (param.isAutoConst ?? false),
       isArray: false,
       isCallback: false,
       isString: false,
-      isPassByValue: isCallbackPointer ? false : deps.isPassByValue,
-      isPassByReference: isCallbackPointer ? true : !deps.isPassByValue,
-      forcePointerSyntax: isCallbackPointer || undefined,
+      isPassByValue: needsPointerSemantics ? false : deps.isPassByValue,
+      isPassByReference: needsPointerSemantics ? true : !deps.isPassByValue,
```
jlaustill: **[important]** design: This is the other half of the duplicated opaque rule. Here (`fromSymbol`, header path) opaqueness *explicitly* forces `isPassByValue: false` / `isPassByReference: true` / `forcePointerSyntax: true`.

But in `fromAST` (body path, lines ~147-171) opaqueness only adds `|| isOpaque` to `forcePointerSyntax` and `&& !isOpaque` to `isAutoConst` — it never explicitly forces pass-by-reference. It gets there only because `isTypedefStructType(typeName)` returns true for forward-declared structs. If that ever stops holding, the two paths silently diverge: header stays a pointer, body becomes pass-by-value.

This is exactly why the rule should be expressed once. Carry `isOpaqueHandle` on `IParameterInput` and resolve pass-by-reference/pointer/const for opaque handles in a single place (the builder) rather than re-deriving it per adapter.

jlaustill (commented):
## Review Summary

**PR:** #1001 — fix: skip auto-const for opaque handle function parameters

### Overview
Fixes #995: opaque-handle parameters (incomplete struct typedefs like LVGL's `lv_obj_t`) were emitted as `const T*`, breaking calls to mutating C APIs. The output is correct, the integration test exercises a real non-const C API, and all CI is green. However, the fix encodes the opaque *behavior rule* in two places, which is a dual code path that should be unified before merge.

### Findings

| Severity | Count |
|----------|-------|
| Critical | 0 |
| Important | 1 |
| Moderate | 0 |
| Nit | 0 |

### Key Issues
- **[important] design — dual code path for the opaque rule.** Detection (`CodeGenState.isOpaqueType`) is correctly single-sourced, but the *consequence* of opaqueness (suppress auto-const + force pointer + pass-by-reference) is computed independently in both `ParameterInputAdapter.fromAST` (body) and `fromAST`/`fromSymbol` (header). They don't even derive it the same way — `fromSymbol` explicitly sets `isPassByReference: true`, while `fromAST` relies on `isTypedefStructType()` happening to be true for forward-declared structs. They agree today by coincidence, not by construction. The `ParameterSignatureBuilder` contract ("all decisions are pre-computed in the input") is the natural home for this: carry a single `isOpaqueHandle` flag on `IParameterInput`, have both adapters pass it through unchanged, and apply the rule once. See inline comments.

### What Looks Good
- **Detection genuinely single-sourced:** both paths route through `CodeGenState.isOpaqueType()`, reading the pre-resolved opaque set (filtered at `Transpiler.ts:375` via body-resolution), so a completed struct is never misclassified as opaque.
- **Strong test coverage:** real compile+execute integration test with a mock C library exposing a non-const mutating API, plus targeted unit tests for both adapters including negative cases.
- **Defense in depth on the const suppression itself** (both `Transpiler.ts` and `fromSymbol`).

---

src/transpiler/Transpiler.ts
jlaustill: Addressed in 68bcab42. Consolidated the opaque rule into `ParameterSignatureBuilder`:

1. Added `isOpaqueHandle` to `IParameterInput` as a pass-through flag
2. Both `fromAST` and `fromSymbol` now just set `isOpaqueHandle` without applying the rule
3. `ParameterSignatureBuilder` applies the rule once:
   - `_getConstPrefix()`: suppresses auto-const for opaque handles
   - `_buildRefParam()`: forces pointer syntax
   - `build()`: routes opaque handles to `_buildRefParam()`

The Transpiler comment was updated to reflect that the rule is applied in the builder, not just resolved onto the symbol.

src/transpiler/output/codegen/helpers/ParameterInputAdapter.ts
jlaustill: Addressed in 68bcab42. The rule application is now consolidated in `ParameterSignatureBuilder`:

- `fromSymbol` now just sets `isOpaqueHandle: param.isOpaqueHandle || undefined` (pass-through)
- No longer manipulates `isPassByValue`, `isPassByReference`, `forcePointerSyntax`, or `isAutoConst` for opaque handles
- Builder's `build()` routes `isOpaqueHandle` params to `_buildRefParam()`
- Builder's `_buildRefParam()` forces pointer syntax for `isOpaqueHandle`
- Builder's `_getConstPrefix()` suppresses auto-const for `isOpaqueHandle`

The same pattern applies to `fromAST` — it just detects and passes through `isOpaqueHandle`, letting the builder apply the rule.

src/transpiler/Transpiler.ts
```diff
@@ -1517,31 +1517,41 @@ class Transpiler {
...
+        const isOpaque = CodeGenState.isOpaqueType(param.type ?? "");
```
jlaustill, line 1528: **[moderate]** design: This opaque resolution runs *after* the callback-compatible early-return above (the `if (callbackTypedefType)` branch that returns `TypedefParamParser.resolveCallbackParams(...)`). So on the **header** path, callback-compatible functions never get `isOpaqueHandle` resolved — while the **body** path (`ParameterInputAdapter.fromAST`) sets `isOpaqueHandle` unconditionally, regardless of callback status.

I traced this and it produces **no output difference today**, because for callback-compatible params the typedef signature is authoritative for both const-ness and pointer syntax (`forceConst`/`forcePointerSyntax` win), so the `isOpaqueHandle` flag is inert there. So this is not a bug.

But it is exactly the kind of body-vs-header asymmetry this project treats as a smell: if a future change ever makes `isOpaqueHandle` affect output independently of `forceConst`, the `.h` would silently diverge from the `.c` for callback functions. Worth a one-line comment here noting that callback-compatible functions intentionally skip opaque resolution because the typedef is authoritative — so the next reader doesn't have to re-derive that the asymmetry is safe.

jlaustill (commented):
## Review Summary

**PR:** #1001 — fix: skip auto-const for opaque handle function parameters

### Overview
Fixes #995, where opaque-handle parameters (incomplete struct typedefs like LVGL's `lv_obj_t`) were emitted as `const T*`, breaking calls to non-const C APIs and producing a hard error in C++ output mode. The fix is clean, well-tested (TDD), and — importantly for this project — consolidates the rule into a single source of truth rather than patching two paths independently. Detection (`isOpaqueHandle`) is resolved once and threaded through `IParameterSymbol` / `IParameterInput`, while the actual rule (suppress auto-const, force pointer syntax) lives only in `ParameterSignatureBuilder`. This mirrors the #914 callback approach and respects the No-Duplicate-Code-Paths rule.

I traced both generation paths: the C++ header path was previously emitting `const widget_t& w` (a reference) and is now correctly `widget_t* w`; the integration test exercises both `.c`/`.cpp` and `.h`/`.hpp` via a public scope function and validates runtime behavior. The expected snapshots match.

### Findings

| Severity | Count |
|----------|-------|
| Critical | 0 |
| Important | 0 |
| Moderate | 1 |
| Nit | 0 |

### Key Issues
- **[moderate]** One inline note on `Transpiler.ts`: the new opaque resolution sits *after* the callback-compatible early-return, so the header path never sets `isOpaqueHandle` for callback-compatible functions while the body path always does. No output divergence today (the typedef is authoritative for those params), but it's a latent asymmetry worth a clarifying comment given how sensitive this project is to dual paths.

### What Looks Good
- Genuine single-source-of-truth consolidation — the rule lives only in `ParameterSignatureBuilder._getConstPrefix`/`_buildRefParam`; adapters merely pass the flag through. The 4-commit history shows the author noticed and fixed the dual-path version (commit `997ce38`/`68bcab4`).
- Strong test coverage: unit tests assert exact signature strings (`widget_t* widget`, including the C++ `&`→`*` case and the explicit-const-preserved case), plus an execution integration test with a real mock C library and `test-link`.
- Behavior is consistent with the ADR-006 array reasoning (#986) — opaque handles, like arrays, are mutable-by-default because the transpiler can't see into the incomplete type.

---

src/transpiler/Transpiler.ts
jlaustill: Addressed in ad6f0666. Added a clarifying comment explaining that callback-compatible functions intentionally skip `isOpaqueHandle` resolution because their parameters get pointer/const semantics from the typedef signature (`isCallbackPointer`/`isCallbackConst`), which takes precedence over opaque handling in the builder.

PR #1003, opened 2026-06-18
fix: arrays of opaque handles mis-generate init and element address-of (#996)

### LINKED ISSUE #996: Arrays of opaque handles: invalid `= NULL` initializer and spurious `&` on element access
## Summary

A fixed-size array of opaque handles (e.g. `lv_obj_t[3]`) mis-generates in two ways:

1. **Invalid initializer:** the array is emitted as `static T* arr[3] = NULL;` — a scalar `NULL` initializer for an array (and `NULL` isn't even in scope without `<stddef.h>`). It should be a brace initializer, e.g. `= {0}`.
2. **Spurious address-of on element access:** passing an element `arr[i]` to a function generates `&arr[i]` (`T**`) instead of `arr[i]` (`T*`).

Discovered trying to store LVGL widgets in an array field and style them in a loop (ogauge).

**C-Next version:** v0.2.16

## Minimal repro

`fake_lib.h`:
```c
#ifndef FAKE_LIB_H
#define FAKE_LIB_H
typedef struct _widget widget_t;
widget_t* widget_create(void);
void widget_move(widget_t* w, int x);
#endif
```

`test.cnx`:
```
#include "fake_lib.h"

scope UI {
  // A fixed-size array of opaque handles, stored and then used by index.
  widget_t[3] widgets;

  void run() {
    u8 i <- 0;
    while (i < 3) {
      this.widgets[i] <- global.widget_create();
      global.widget_move(this.widgets[i], 5);
      i +<- 1;
    }
  }
}
```

Run: `cnext test.cnx`

## Expected (generated `test.c`)

```c
static widget_t* UI_widgets[3] = {0};
...
        UI_widgets[i] = widget_create();
        widget_move(UI_widgets[i], 5);
```

## Actual (generated `test.c`)

```c
static widget_t* UI_widgets[3] = NULL;     // (1) invalid array init
...
        UI_widgets[i] = widget_create();
        widget_move(&UI_widgets[i], 5);    // (2) spurious & -> widget_t**
```

## Compiler result

`gcc -c test.c`:
```
error: 'NULL' undeclared here (not in a function)            // facet 1
note: expected 'widget_t *' but argument is of type 'widget_t **'   // facet 2
```

## Impact

Arrays of handles can't be used to drive repeated UI work in a loop (the natural way to style N similar widgets). Both facets must be avoided by falling back to individually-named fields.

jlaustill:
## Summary

Fixes #996. A fixed-size array of opaque handles (e.g. `widget_t[3]`, where `widget_t` comes from `typedef struct _widget widget_t;`) mis-generated in two ways that both broke compilation:

1. **Invalid initializer** — the array was emitted as `static widget_t* arr[3] = NULL;` (a scalar `NULL` for an array; `NULL` also isn't in scope without `<stddef.h>`). It should be a brace initializer.
2. **Spurious address-of** — passing an element `arr[i]` (already `widget_t*`) to a function expecting `widget_t*` generated `&arr[i]` (`widget_t**`).

```
gcc -c test.c
  error: 'NULL' undeclared here (not in a function)
  note: expected 'widget_t *' but argument is of type 'widget_t **'
```

## Fix

| Facet | Root cause | Fix |
|-------|-----------|-----|
| `= NULL` for array | `ScopeGenerator` forced `= NULL` for **every** opaque/external-struct field, ignoring `isArray` | Only scalars (single pointers) get `= NULL`; arrays route through the existing `generateInitializer` → `getZeroInitializer(type, isArray)` — the single source of truth for zero-init (ADR-015). Emits `{0}` in C, `{}` in C++. |
| `&arr[i]` | The "opaque handle is already a pointer → skip `&`" check matched only the exact base name, not subscript access | `isOpaqueScopeVariable` → **`isOpaqueScopeVariableAccess`**, which also recognizes an element of an opaque-handle array as already-a-pointer. |

### No duplicate path to unify
Both argument-generation paths were checked. The C-interop path (`_generateCFunctionArg`) decides address-of via the opaque predicate; the C-Next path (`ArgumentGenerator`) **never** takes the address of an opaque handle — `_shouldPassByValue` short-circuits pointer-typed params to pass-by-value. So there is no duplicated opaque-address-of decision. A C-Next-call regression test is included alongside the C-interop repro to lock in that the two paths stay in parity.

## Tests
- `bugs/issue-996-array-opaque-handles/` — C-interop repro + C-Next-call parity repro (both C and C++ snapshots)
- Unit coverage for `isOpaqueScopeVariableAccess` (scalar, array-element, prefix-collision, non-opaque)
- Generated C **and** C++ compile clean (`gcc`/`g++`)
- Full suite green: 5632 unit + 959 integration tests pass; scalar opaque handles (#948/#957/#958) unchanged


jlaustill (commented):
## Review Summary

**PR:** #1003 — fix: arrays of opaque handles mis-generate init and element address-of (#996)

### Overview

This PR fixes Issue #996 where fixed-size arrays of opaque handles (e.g., `widget_t[3]`) generated incorrect C code: `= NULL` (scalar) instead of `{0}` for initialization, and `&arr[i]` (pointer-to-pointer) instead of `arr[i]` for element access. The fix is well-targeted and properly extends the existing opaque handle machinery without duplicating code paths.

### Findings

| Severity | Count |
|----------|-------|
| Critical | 0 |
| Important | 0 |
| Moderate | 0 |
| Nit | 0 |

### What Looks Good

- **Single source of truth preserved** — The array initialization now routes through `generateInitializer()` → `getZeroInitializer(type, isArray)`, maintaining ADR-015 compliance
- **Clean method rename** — `isOpaqueScopeVariable` → `isOpaqueScopeVariableAccess` accurately reflects the expanded semantics (handles both direct and subscript access)
- **Comprehensive test coverage** — Unit tests cover scalar access, array-element access, prefix collision rejection ("UI_widgetsExtra" vs "UI_widgets[i]"), and non-opaque arrays
- **Dual-path parity verified** — The PR includes both C-interop and C-Next call path tests, with clear documentation explaining why there's no actual code duplication between the paths
- **Both C and C++ modes covered** — The `{0}` vs `{}` difference is handled by the existing `getZeroInitializer` logic

### No Issues Found

The implementation is clean and well-documented. The fix correctly:
1. Adds `&& !isArray` guard in `ScopeGenerator.ts:247` to route arrays through the standard initializer path
2. Renames and extends `isOpaqueScopeVariableAccess()` to extract the base name before `[` for array element matching
3. Updates all call sites and tests consistently

---

Issue #1006, opened 2026-06-18
issue-986 execution test can't link its C helper (process_data) — the closed-#986 fix is unverified


jlaustill:
## Summary

The `bugs/issue-986-const-array-param/` regression test is marked `// test-execution` and calls `process_data()` from a sibling helper `fake_lib.c`. When run, it fails at **link time** because `fake_lib.c` is not compiled/linked into the executable, so the `#986` fix is never actually exercised. Undetected because `bugs/` isn't in CI (see #929).

## Reproduction

```
npm test -- bugs/issue-986-const-array-param
```

```
C compile for execution failed: /usr/bin/ld: undefined reference to `process_data'
collect2: error: ld returned 1 exit status
```

## Analysis

- The `#986` codegen fix itself is intact — the generated C is correct (no spurious `const`):
  ```c
  void Handler_process(uint8_t data[8]) {
      process_data(data);
  }
  ```
- `bugs/issue-986-const-array-param/fake_lib.c` **does** define `void process_data(uint8_t data[8])`.
- The test harness does not link the sibling `fake_lib.c` when building/running the execution test, so linking fails.

## Impact

Execution tests that depend on an external C helper implementation (`fake_lib.c`) silently cannot run, so the behavior they assert (here, the closed #986 fix) is not verified.

## Fix direction

Teach the execution-test runner to compile and link sibling helper `.c` files in a `bugs/` test directory (or document/standardize a helper-linking convention). Best addressed alongside #929 (automated `bugs/` regression).

Issue #1005, opened 2026-06-18
Stale bug-repro test: bugs/switch-array-element uses obsolete `u8 data[4]` syntax and lacks expected snapshots


jlaustill:
## Summary

The `bugs/switch-array-element/` regression test no longer transpiles. Its `.cnx` uses the obsolete C-style postfix array declaration `u8 data[4]`, which the transpiler now rejects **by design** (it requires the prefix form `u8[4] data`). The test also has **no `.expected.*` snapshot files**, so it could never have been a passing snapshot test. It went undetected because `bugs/` isn't run in CI (see #929).

## Reproduction

`bugs/switch-array-element/switch-array.test.cnx` (line 12):

```
u8 data[4] <- [1, 2, 3, 4];
```

Run: `cnext switch-array.test.cnx`

```
Error: ...:3:8 Code generation failed: C-style array declaration is not allowed.
Use 'u8[4] data' instead of 'u8 data[4]'
```

## Analysis

- The rejection is intentional — the prefix form transpiles fine:
  ```
  u8[4] data <- [1, 2, 3, 4];   // OK
  ```
- The original bug this test guarded ("switch on array element rejected as boolean type") is **fixed**: `switch (data[0]) { ... }` transpiles correctly with the prefix syntax.
- So this is test maintenance, not a transpiler bug: update the `.cnx` to `u8[4] data` and generate the missing `.expected.c/.cpp/.h/.hpp` snapshots.

## Fix

1. Update `switch-array.test.cnx` to use `u8[4] data <- [1, 2, 3, 4];`.
2. Generate the missing expected snapshots (`npm test -- bugs/switch-array-element --update`).
3. Ideally fold into #929 so `bugs/` is verified in CI and can't rot again.

Issue #958, opened 2026-02-24
Complete struct types generate by-value instead of by-pointer (ADR-006 violation)


jlaustill:
## Summary

When a struct has its full definition visible (not just a forward declaration), the transpiler treats it as a concrete value type. Scope variables, function parameters, and local variables all generate by-value instead of by-pointer, violating ADR-006.

#948 fixed this for forward-declared (incomplete) structs. This bug is the same issue but for structs whose full body is visible through transitive includes.

## Reproduce

Minimal repro in `/tmp/cnext-bugs/958-forward-decl-params-locals/`

```c
// fake_lib.h
typedef struct _widget_t widget_t;

// Full definition visible (like LVGL's _private.h included transitively)
struct _widget_t {
    int32_t value;
    uint8_t flags;
};

// Functions behind #if guard transpiler can't evaluate (#945)
#if USE_WIDGET != 0
widget_t * widget_create(void);
int32_t widget_get_value(widget_t * w);
typedef void (*widget_cb_t)(widget_t * w);
void widget_set_callback(widget_t * w, widget_cb_t cb);
#endif
```

```cnx
#include "fake_lib.h"

scope App {
  widget_t root;

  void on_change(widget_t w) {
    i32 val <- global.widget_get_value(w);
  }

  public void init() {
    widget_t child <- global.widget_create();
    global.widget_set_callback(child, this.on_change);
  }
}
```

## Expected

```c
static widget_t* App_root = NULL;
static void App_on_change(widget_t* w) { ... }
void App_init(void) { widget_t* child = widget_create(); ... }
```

## Actual

```c
static widget_t App_root = {0};           // by-value
static void App_on_change(widget_t w) {}  // by-value
void App_init(void) { widget_t child = widget_create(); }  // by-value
```

## Root cause

When only a forward declaration is visible, the transpiler correctly generates pointers (#948 fix). But when the full `struct _X { ... }` is visible, it treats the type as concrete. Per ADR-006, ALL struct types should be pass-by-reference regardless of completeness.

## Impact

LVGL v9's `lvgl.h` transitively includes `lvgl_private.h`, which pulls in full struct definitions. Combined with #945 (function declarations hidden behind `#if LV_USE_*`), every LVGL type used in C-Next generates by-value.

Affected types: `lv_obj_t`, `lv_event_t`, `lv_display_t`, `lv_indev_t`, `lv_area_t`, `lv_indev_data_t`

## Related

[...]

PR #1020, opened 2026-06-20
ci: add automated regression testing for bugs/ directory

### LINKED ISSUE #929: Automated regression testing for bug category patterns
## Summary

Automate verification that all bugs in the `bugs/` directory remain fixed in both C and C++ modes.

## Background

The `bugs/` directory contains reproduction cases for past bugs. These should be automatically verified in both modes to prevent regressions, as many bugs were mode-specific.

## Current Bug Categories

| Category | Example Issues | Mode Sensitivity |
|----------|---------------|------------------|
| Const linkage | #525, #852 | C++ specific |
| Constructor detection | #517 | C++ specific |
| Parameter passing | #558 | Both modes |
| Struct initialization | #517, #882 | C++ specific |
| Member access | #314, #556 | Both modes |
| Namespace handling | #314, #388 | Both modes |
| Header generation | #369, #522 | Both modes |
| Callback interop | #895 | Both modes |

## Proposed Implementation

1. **Regression test script**
   ```bash
   for dir in bugs/issue-*; do
     # Run in C mode
     cnext "${dir}"/*.cnx --mode c
     gcc -o test_c *.c && ./test_c
     
     # Run in C++ mode  
     cnext "${dir}"/*.cnx --mode cpp
     g++ -o test_cpp *.cpp && ./test_cpp
   done
   ```

2. **CI Integration**
   - Run regression suite on every PR
   - Fail if any bug reproduction case regresses

3. **Mode-aware handling**
   - Respect `// test-c-only` and `// test-cpp-only` markers in bug reproductions
   - Some bugs are inherently mode-specific

## Acceptance Criteria

- [ ] Script runs all bug reproductions
- [ ] Both C and C++ modes tested where applicable
- [ ] CI integration catches regressions
- [ ] Clear reporting of which bug regressed
- [ ] Mode markers respected
- [ ] New bugs automatically included

## Directory Structure Expected

```
bugs/
  issue-517-cpp-class-init/
    repro.test.cnx
    repro.expected.c
    repro.expected.cpp  # If C++ specific
  issue-525-const-linkage/
    ...
```

## Labels

testing, c-cpp-interop, ci, regression

jlaustill:
## Summary

Closes #929

- Add CI pipeline stages to verify bug fixes remain fixed in `bugs/` directory
- Add npm scripts for local testing of bugs/ regression suite
- Update bugs/ snapshots to sync with current transpiler output

### CI Workflow (`.github/workflows/pr-checks.yml`)
- **bugs-transpile**: Transpile bugs/ tests (parallel with integration-transpile)
- **bugs-execute**: Compile and run bugs/ tests (parallel with integration-execute)  
- **bugs-verify-clean**: Verify no uncommitted changes in bugs/

### npm Scripts (`package.json`)
- `test:bugs` / `test:bugs:q`: Run all bugs/ tests
- `integration:bugs:transpile` / `integration:bugs:execute`: CI split stages

### Snapshot Updates
Updated 6 bugs/ test directories to sync with current transpiler output (source filename was added to header comment).

## Known Failing Tests

The following tests fail and are tracked in separate issues:
- `switch-array-element` (#1005): Uses obsolete C-style array syntax
- `issue-986-const-array-param` (#1006): Linker issue with extern function

These failures are expected and the CI will expose them (which is the desired behavior).

## Test plan

- [x] `npm run test:bugs` runs locally
- [x] `npm run integration:bugs:transpile` works
- [x] `npm run integration:bugs:execute` works
- [ ] CI runs the new bugs regression stages


bugs/issue-948-forward-then-define/test.expected.c
```diff
@@ -11,8 +11,8 @@
 #include "lib_with_late_def.h"
 
 /* Scope: Draw */
-static point_t Draw_origin = {0};
+static point_t* Draw_origin = NULL;
```
jlaustill: **[critical]** bug: This snapshot update blesses a transpiler regression rather than catching it.

The old snapshot was `static point_t Draw_origin = {0};` + `point_init(&Draw_origin);` — correct **value** semantics. The new output treats `point_t` as **opaque**: `static point_t* Draw_origin = NULL;` + `point_init(Draw_origin);` (passing NULL).

But `point_t` is fully defined in `lib_with_late_def.h` (`typedef struct _point_t point_t;` followed by `struct _point_t { ... };`), and this test's own comment states: *"The type should NOT be opaque since it has a full definition."* I confirmed with the live transpiler — it now emits the pointer/NULL form, so the bug #948 was guarding has **regressed**. On `main` this test currently *fails* (committed value snapshot ≠ regressed output); updating the snapshot here converts a regression-catching failure into a green test that hides the bug.

This is exactly what #929 exists to prevent. Please **don't bless this snapshot** — file the #948 regression (forward-decl-then-defined-in-same-header type wrongly treated as opaque), fix it in the transpiler, and regenerate to the value form. The same applies to `test.expected.cpp` / `test.test.c` / `test.test.cpp` in this dir.

bugs/issue-958-complete-struct-by-value/test.expected.c
```diff
@@ -13,7 +13,7 @@
 /* Scope: App */
 static widget_t* App_root = NULL;
 
-static void App_on_change(widget_t* w) {
+void App_on_change(widget_t* w) {
```
jlaustill, line 16: **[important]** bug: please verify this isn't a second blessed regression. `on_change` is a **non-public** scope member, yet it lost `static` (`static void App_on_change` → `void App_on_change`) and is now exported in the header along with `typedef struct widget_t widget_t;`. `widget_t` is a complete type here (`fake_lib.h` pulls in `widget_private.h` with the full definition), so the "complete-struct-by-value" intent should hold.

Two concerns: (1) a private callback shouldn't need external linkage just to have its address taken within the same TU; (2) emitting `typedef struct widget_t widget_t;` may conflict with the library's real tag (e.g. `struct _widget_t`) if both are visible. If this is intended behavior, a one-line note in the test would help; if not, it's a regression to file rather than snapshot.

.github/workflows/pr-checks.yml
```diff
@@ -166,6 +166,46 @@ jobs:
...
+
+      - name: Upload generated bugs test files
+        uses: actions/upload-artifact@v4
+        with:
+          name: generated-bugs-files
+          retention-days: 1
+          path: |
+            bugs/**/*.test.c
+            bugs/**/*.test.cpp
+            bugs/**/*.test.h
```
jlaustill, line 207: **[moderate]** testing: these upload globs (and the `bugs-execute`/`bugs-verify-clean` downloads) list `*.test.c/.cpp/.h` but omit `*.test.hpp`. Generated `.test.hpp` files exist for every bug dir (e.g. `issue-948-forward-decl-scope-var/test.test.hpp`), so `bugs-verify-clean` re-downloads everything *except* the C++ headers — meaning drift in a generated `.test.hpp` would never fail the clean check. Add `bugs/**/*.test.hpp` here (the pre-existing integration jobs have the same omission for `tests/**`, worth fixing there too).

jlaustill (commented):
## Review Summary

**PR:** #1020 — ci: add automated regression testing for bugs/ directory (closes #929)

### Overview
The CI wiring is solid and the goal is exactly right — `bugs/` rotted precisely because it wasn't gated. The three-stage layout (transpile → execute, plus verify-clean) mirrors the existing integration jobs cleanly, and the npm scripts work (I confirmed `tsx scripts/test.ts -- bugs ...` parses correctly — the literal `--` is skipped by the `!arg.startsWith('-')` filter). I ran the full suite on this branch: **12/12 pass** (the branch has merged #1021 and #1023, so the "known-failing" tests are now green).

**However, the snapshot sync blesses a real regression.** The PR description frames the snapshot changes as "source filename was added to header comment," but at least two dirs have substantive semantic codegen changes — and one of them is the very kind of bug this PR exists to catch.

### Findings

| Severity | Count |
|----------|-------|
| Critical | 1 |
| Important | 1 |
| Moderate | 3 |
| Nit | 0 |

### Key Issues
- **Critical (#948-forward-then-define):** The snapshot was updated from correct value semantics (`point_t Draw_origin = {0}` / `point_init(&Draw_origin)`) to opaque-pointer semantics (`point_t* Draw_origin = NULL` / `point_init(Draw_origin)` — passing NULL). The test's own comment says *"The type should NOT be opaque since it has a full definition,"* and `point_t` IS fully defined in `lib_with_late_def.h`. The transpiler has regressed; this PR makes the test green by blessing the buggy output instead of filing/fixing the regression. That's the opposite of #929's purpose and conflicts with "Correctness Over Convenience." See inline.
- **Important (#958):** Private `on_change` lost `static` and is now exported in the header (with a `typedef struct widget_t widget_t;` forward decl) — `widget_t` is a complete type here. Please confirm this isn't a second regression. See inline.
- **Moderate:** description mischaracterizes the snapshot diffs; the "Known Failing Tests" section is now stale (12/12 pass); and the upload globs omit `*.test.hpp`, so `bugs-verify-clean` can't catch C++ header drift. See inline + below.

### What Looks Good
- Correct, well-structured CI jobs that close a real coverage gap (the root cause behind #1005/#1006/#1021/#1023).
- Dependency handled correctly by merging main — #1021/#1023 are in, so the gate goes green rather than red.
- npm scripts verified to parse and run; suite runs 12/12 locally.

---

bugs/issue-948-forward-then-define/test.expected.c
jlaustill: Fixed. The regression was in `isTypedefStructType()` which was not updated to perform query-time resolution when commit c0d4cb4c changed to additive-only struct collection. Added the same resolution pattern that `isOpaqueType()` uses - checking if the underlying struct tag has a body definition. Snapshots now correctly show value semantics: `static point_t Draw_origin = {0}` with `point_init(&Draw_origin)`.

bugs/issue-958-complete-struct-by-value/test.expected.c
jlaustill: Investigated. The `static` loss is intentional behavior per ADR-016 (commit a0075c96): scope function visibility defaults changed from private to public. Since `on_change` has no visibility modifier, it's now public by default (no `static`). The original expected file was created before ADR-016. The scope variable is now also fixed to use value semantics: `static widget_t App_root = {0}` (same fix as #948).

.github/workflows/pr-checks.yml
jlaustill: Fixed. Added `bugs/**/*.test.hpp` to the upload glob.

PR #1021, opened 2026-06-20
fix: update switch-array-element test to use prefix array syntax

### LINKED ISSUE #1005: Stale bug-repro test: bugs/switch-array-element uses obsolete `u8 data[4]` syntax and lacks expected snapshots

jlaustill:
## Summary

Closes #1005

Updates the `bugs/switch-array-element` regression test to use the current C-Next prefix array syntax and generates the missing expected snapshots.

## Problem

The test used the obsolete C-style postfix array declaration:
```c-next
u8 data[4] <- [1, 2, 3, 4];  // Rejected by design
```

This syntax is now rejected in favor of the prefix form which the transpiler requires.

## Solution

Updated to use the prefix syntax:
```c-next
u8[4] data <- [1, 2, 3, 4];  // Correct
```

Generated the missing `.expected.c`, `.expected.cpp`, `.expected.h`, `.expected.hpp` snapshot files.

## Verification

The original bug (switch on array element rejected as "boolean type") remains fixed with the updated syntax:
```c-next
switch (data[0]) {
    case 1 { }
    default { }
}
```

This now transpiles correctly to:
```c
switch (data[0U]) {
    case 1: { break; }
    default: { break; }
}
```

## Test plan

- [x] `npm test -- bugs/switch-array-element` passes
- [ ] CI passes


bugs/switch-array-element/switch-array.test.cnx
```diff
@@ -9,7 +9,7 @@ scope Test {
     }
 
     public void failing() {
-        u8 data[4] <- [1, 2, 3, 4];
+        u8[4] data <- [1, 2, 3, 4];
```
jlaustill, line 12: **[moderate]** testing: The syntax fix and snapshots are correct (I verified the regenerated output is byte-identical to the committed `.expected.*` in both modes). But note this test still isn't exercised by CI — `scripts/test.ts:450` defaults to `join(rootDir, "tests")`, so `npm test` only runs `bugs/switch-array-element` when explicitly pointed at it. That's exactly why it rotted (#1005). Worth folding `bugs/` into the CI test sweep under #929 so this can't silently break again; otherwise the unchecked "CI passes" box in the PR will stay unchecked for this path indefinitely.

Nit: the `.cnx` has no test marker (`// test-no-exec` / `// test-transpile-only`). Since there's no `main()`, an explicit marker would make the intended validation (compile + snapshot, no execution) self-documenting.

jlaustill (commented):
## Review Summary

**PR:** #1021 — fix: update switch-array-element test to use prefix array syntax

### Overview
Clean, correct test maintenance. The `.cnx` switches from the obsolete postfix `u8 data[4]` to the prefix `u8[4] data` form, which matches the dominant repo convention (123 prefix uses vs. 3 postfix, all of which are `const` arrays). I regenerated the snapshots locally and they are **byte-identical** to the committed `.expected.*` in both C and C++ modes — verified accurate.

### Findings

| Severity | Count |
|----------|-------|
| Critical | 0 |
| Important | 0 |
| Moderate | 1 |
| Nit | 1 |

### Key Issues
- **Moderate (testing):** This regression test still won't run in CI. `scripts/test.ts` defaults to scanning `tests/`, so `npm test` skips `bugs/` unless explicitly targeted — the same root cause that let this test rot (#1005 / #929). The fix is correct, but without folding `bugs/` into CI the test can silently break again. See inline.

### What Looks Good
- Snapshots verified: fresh transpiler output matches all four committed `.expected.*` files exactly (C + C++), and `expected == test` for every extension.
- Syntax change aligns with the established prefix-array convention.
- Scope is tight and matches the issue's stated fix.

---

PR #1022, opened 2026-06-20
fix(E0381): extend use-before-init check to compound assignment

### LINKED ISSUE #1012: Compound assignment bypasses use-before-init check (E0381)
## Summary

Reading an uninitialized local in an expression is correctly rejected with `E0381`, but a **compound assignment** — which reads its LHS — is **not**. It compiles instead of erroring. The use-before-init analysis should treat the compound-assign target as a read.

**Bypasses the check** (compiles, exit 0):

```cnx
void demo() {
    u32 sum;    // zero-initialized storage, never explicitly assigned
    sum +<- 5;  // sum = sum + 5 — reads `sum` before any explicit assignment
}
```

→ emits (note `sum` is read by the overflow helper before it is assigned):

```c
void demo(void) {
    uint32_t sum = 0;
    sum = cnx_clamp_add_u32(sum, 5U);
}
```

**Contrast** — a plain read of the same uninitialized local is correctly rejected:

```cnx
void demo() {
    u32 y;
    u32 z <- y + 1;
}
```

```
error[E0381]: use of uninitialized variable 'y'
```

## Expected

`sum +<- 5;` on an uninitialized `sum` should error `E0381` (it reads `sum`), consistent with the plain-read case. Applies to every compound operator: `+<- -<- *<- /<- %<- &<- |<- ^<- <<<- >><-`.

## Notes

- Storage is still zero-initialized (`uint32_t sum = 0;`) — this is purely about the use-before-init *safety* analysis, which currently does not see the compound-assign read.

jlaustill:
## Summary

- Fixes compound assignment (`+<-`, `-<-`, `*<-`, etc.) bypassing the use-before-init check
- These operators read their LHS before writing, but the analyzer wasn't treating them as reads

## The Bug

```c-next
void demo() {
    u32 sum;      // declared but not initialized
    sum +<- 5;    // was accepted - should error E0381
}
```

## The Fix

Added a check in `InitializationListener.enterAssignmentStatement` to detect compound assignment operators (any operator other than `<-`) and call `checkRead()` on the LHS before recording the assignment.

## Test plan

- [x] New test: `compound-assign-use-before-init.test.cnx` — basic bug reproduction
- [x] New test: `compound-assign-all-ops.test.cnx` — all 10 compound operators trigger E0381
- [x] New test: `compound-assign-struct-field.test.cnx` — struct field case
- [x] New test: `compound-assign-initialized.test.cnx` — happy path (initialized vars still work)
- [x] 8 new unit tests for comprehensive coverage
- [x] All 971 integration tests pass
- [x] All 5644 unit tests pass

Fixes #1012


tests/initialization/compound-assign-use-before-init.test.c
```diff
@@ -0,0 +1,24 @@
...
+    uint32_t result;
+    if (__builtin_add_overflow(a, (uint32_t)b, &result)) return UINT32_MAX;
+    return result;
+}
+
+// test-error
+// Test: Compound assignment should check use-before-init on LHS
+// Issue #1012: Compound assignment bypasses E0381 check
+void demo(void) {
+    uint32_t sum = 0;
```
jlaustill: **[important]** testing: This generated `.test.c` shouldn't exist for a `// test-error` case. The transpiler stops at the E0381 error and emits no C — which is why the other two test-error files added in this PR (`compound-assign-all-ops`, `compound-assign-struct-field`) have only `.expected.error`, and 56 of 59 test-error `.cnx` files in the repo have no `.test.c` sibling.

Worse, the content here contradicts the test itself: the source declares `u32 sum;` (no initializer), but this output shows `uint32_t sum = 0;` and a *successful* compile — i.e. it's a stale artifact representing exactly the bug this PR fixes, even though `.expected.error` says compilation must fail with E0381.

Please remove it:
```
git rm tests/initialization/compound-assign-use-before-init.test.c
```

src/transpiler/logic/analysis/InitializationAnalyzer.ts
```diff
@@ -178,6 +185,36 @@ class InitializationListener extends CNextListener {
     }
   };
 
+  /**
+   * Check that the LHS of a compound assignment is initialized.
+   * Issue #1012: Compound assignment (e.g., sum +<- 5) reads the LHS.
+   */
+  private _checkCompoundAssignmentRead(
```
jlaustill: **[important]** design: `_checkCompoundAssignmentRead` re-implements the exact target-classification dispatch already in `enterAssignmentStatement` above (`postfixOps.length === 0` → base; else `analyzePostfixOps(...)` + `identifiers.length >= 2 && !hasSubscript` → struct field, else base). The *decision* of "is this a bare var / struct field / array element" now lives in two places that must stay in lockstep — the dual-code-path anti-pattern CLAUDE.md calls out ("single source of truth means the decision, not just the data"). If the classification ever changes (e.g. nested `a.b.c`, or subscript-then-field), both copies must be edited together or they silently diverge.

Consider resolving the target once into a small descriptor (`{ kind: 'var'|'field'|'base', varName, fieldName }`), then applying `checkRead` (when compound) and `recordAssignment` to that same resolved target — one classification, two consumers.

jlaustill (commented):
## Review Summary

**PR:** #1022 — fix(E0381): extend use-before-init check to compound assignment

### Overview
The fix is correct and well-targeted: compound assignment operators (`+<-`, `-<-`, …) read their LHS before writing, so the analyzer now runs `checkRead()` on the LHS before recording the assignment. Test coverage is excellent — all 10 operators, struct fields, array elements, the happy path, plus 8 unit tests. Two issues below: a stale generated artifact that contradicts the test's own expectation, and a duplicated classification path.

### Findings

| Severity | Count |
|----------|-------|
| Critical | 0 |
| Important | 2 |
| Moderate | 0 |
| Nit | 0 |

### Key Issues
- **Stale `.test.c` for a test-error case** (`compound-assign-use-before-init.test.c`): a `// test-error` test emits no C, so this file shouldn't exist — and its content (`uint32_t sum = 0;`) doesn't match the source (`u32 sum;`) and represents a *successful* compile of code the PR's own `.expected.error` says must fail. See inline.
- **Duplicate classification path**: `_checkCompoundAssignmentRead` re-implements the same bare-var / struct-field / array-element dispatch already in `enterAssignmentStatement`. Per CLAUDE.md this is the flagged dual-path anti-pattern. See inline.

### What Looks Good
- Clean, minimal core fix with a clear before-write ordering.
- Thorough, well-organized tests (every operator + struct + array + happy path + unit tests).
- Good inline comments and issue references throughout.

---

jlaustill (commented):
## ⚠️ Blocking concern — duplicate classification path (finding #2)

_(Intended as Request Changes — GitHub won't let me set that state on your own PR, so posting as a comment. Please treat as a must-fix before merge.)_

This is the dual-code-path issue, which CLAUDE.md treats as the worst anti-pattern in this project ("single source of truth means the _decision_, not just the data").

`_checkCompoundAssignmentRead` re-implements the exact target-classification dispatch already in `enterAssignmentStatement`:

```
postfixOps.length === 0                                    -> base identifier
analyzePostfixOps(...) + identifiers >= 2 && !hasSubscript -> struct field
else                                                      -> base identifier
```

That decision now lives in two places. The two copies call different leaves (`checkRead` vs `recordAssignment`), so it doesn't read like copy-paste — but the *classification* is duplicated, and the two must stay in lockstep. If the rule ever changes (nested `a.b.c`, subscript-then-field, etc.) and only one copy is updated, the read-check and the assignment-record silently disagree.

**Requested change:** classify the target once, then feed both consumers. For example:

```ts
private _resolveTarget(baseId, postfixOps) {
  if (postfixOps.length === 0) return { varName: baseId };
  const { identifiers, hasSubscript } = analyzePostfixOps(baseId, postfixOps);
  if (identifiers.length >= 2 && !hasSubscript) {
    return { varName: identifiers[0], fieldName: identifiers[1] };
  }
  return { varName: baseId };
}
```

Then `enterAssignmentStatement` resolves once and does `checkRead(...)` (when compound) followed by `recordAssignment(...)` on that same resolved target — one classification, two consumers.

(The stale `.test.c` from the earlier review still stands too, but this is the blocker.)

---

src/transpiler/logic/analysis/InitializationAnalyzer.ts
jlaustill: **[important]** design (blocking): this method duplicates the classification dispatch in `enterAssignmentStatement`. Resolve the target once into a `{ varName, fieldName? }` descriptor and have both `checkRead` (compound) and `recordAssignment` consume it — see the review summary for the suggested `_resolveTarget` helper.

jlaustill:
Addressed review feedback in commit 4a054d43:

1. **Stale .test.c removed** — deleted `compound-assign-use-before-init.test.c` (test-error cases don't generate C output)

2. **Unified classification path** — extracted `_resolveAssignmentTarget()` that returns `{ varName, fieldName? }`. Now the classification happens once and both `checkRead()` and `recordAssignment()` consume the same resolved target:

```ts
const target = this._resolveAssignmentTarget(baseId, postfixOps);
if (isCompoundAssignment) {
  this.analyzer.checkRead(target.varName, line, column, target.fieldName);
}
this.analyzer.recordAssignment(target.varName, target.fieldName);
```

All 971 integration tests and 45 InitializationAnalyzer unit tests pass.

PR #1023, opened 2026-06-20
fix: add test-link marker for issue-986 helper file

### LINKED ISSUE #1006: issue-986 execution test can't link its C helper (process_data) — the closed-#986 fix is unverified

jlaustill:
## Summary

Closes #1006

Fixes the linker error in the `issue-986-const-array-param` regression test by using the test framework's existing `// test-link:` marker.

## Problem

The test calls `process_data()` which is implemented in a sibling `fake_lib.c` file, but the test framework wasn't linking that file:

```
/usr/bin/ld: undefined reference to `process_data'
collect2: error: ld returned 1 exit status
```

## Solution

The test framework already supports linking additional C files via a `// test-link:` marker (see `TestUtils.findLinkedSourceFiles()` in test-utils.ts:588-601). This marker tells the runner to compile and link the specified files during test execution.

Added the marker to the test:
```c-next
// test-execution
// test-link: fake_lib.c
```

## Verification

The #986 fix (array params not marked as const when passed to C APIs) is now fully verified at runtime:
- `Handler_process()` receives `uint8_t data[8]` (not `const uint8_t data[8]`)
- Calls `process_data(data)` which modifies the buffer
- Test verifies the modifications persist

## Test plan

- [x] `npm test -- bugs/issue-986-const-array-param` passes with [C+C++ PARITY]
- [ ] CI passes

### DIFF
--- a/bugs/issue-986-const-array-param/const-array-param.expected.cpp
+++ b/bugs/issue-986-const-array-param/const-array-param.expected.cpp
@@ -6,6 +6,7 @@
 #include "const-array-param.test.hpp"
 
 // test-execution
+// test-link: fake_lib.c
 // Issue #986: Array params transpile as const when passed to C APIs
 // ADR-006 says all params are pass-by-reference as mutable, but array params
 // that are not directly modified get auto-const, breaking C API compatibility.
@@ -20,7 +21,7 @@ void Handler_process(uint8_t data[8]) {
 }
 
 int main(void) {
-    uint8_t buffer[8] = {0};
+    uint8_t buffer[8] = {};
     buffer[0] = 0U;
     buffer[1] = 0U;
     Handler_process(buffer);
--- a/bugs/issue-986-const-array-param/const-array-param.test.cnx
+++ b/bugs/issue-986-const-array-param/const-array-param.test.cnx
@@ -1,4 +1,5 @@
 // test-execution
+// test-link: fake_lib.c
 // Issue #986: Array params transpile as const when passed to C APIs
 // ADR-006 says all params are pass-by-reference as mutable, but array params
 // that are not directly modified get auto-const, breaking C API compatibility.
--- a/bugs/issue-986-const-array-param/const-array-param.expected.c
+++ b/bugs/issue-986-const-array-param/const-array-param.expected.c
@@ -6,6 +6,7 @@
 #include "const-array-param.test.h"
 
 // test-execution
+// test-link: fake_lib.c
 // Issue #986: Array params transpile as const when passed to C APIs
 // ADR-006 says all params are pass-by-reference as mutable, but array params
 // that are not directly modified get auto-const, breaking C API compatibility.
--- a/bugs/issue-986-const-array-param/const-array-param.test.c
+++ b/bugs/issue-986-const-array-param/const-array-param.test.c
@@ -6,6 +6,7 @@
 #include "const-array-param.test.h"
 
 // test-execution
+// test-link: fake_lib.c
 // Issue #986: Array params transpile as const when passed to C APIs
 // ADR-006 says all params are pass-by-reference as mutable, but array params
 // that are not directly modified get auto-const, breaking C API compatibility.
--- a/bugs/issue-986-const-array-param/const-array-param.test.cpp
+++ b/bugs/issue-986-const-array-param/const-array-param.test.cpp
@@ -6,6 +6,7 @@
 #include "const-array-param.test.hpp"
 
 // test-execution
+// test-link: fake_lib.c
 // Issue #986: Array params transpile as const when passed to C APIs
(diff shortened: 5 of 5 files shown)


bugs/issue-986-const-array-param/const-array-param.test.cnx
```diff
@@ -1,4 +1,5 @@
 // test-execution
+// test-link: fake_lib.c
```
jlaustill, line 2: **[moderate]** testing: Verified end-to-end — on this branch the test **passes with [C+C++ PARITY]**, the marker resolves `fake_lib.c` via `findLinkedSourceFiles()` → `join(testDir, file)`, and `Handler_process(uint8_t data[8])` is correctly non-const so the #986 fix is genuinely exercised at runtime. Nicely done reusing the existing `test-link` mechanism instead of a workaround.

Standing caveat (same as the other `bugs/` PRs): `bugs/` isn't in the default `npm test` sweep (`scripts/test.ts` scans `tests/`), so this can't fail CI and could silently rot again — tracked in #929.

bugs/issue-986-const-array-param/const-array-param.expected.cpp
```diff
@@ -20,7 +21,7 @@ void Handler_process(uint8_t data[8]) {
 }
 
 int main(void) {
-    uint8_t buffer[8] = {0};
+    uint8_t buffer[8] = {};
```
jlaustill, line 24: **Nit:** quality: this `{0}` → `{}` change is unrelated to the test-link fix — it's the C++ regeneration catching a previously-stale snapshot. I confirmed current codegen really does emit `= {}` for C++ zero-init while C mode emits `= {0}`, so the delta is correct, not drift. Just confirming the C/C++ divergence is intentional: `= {}` is valid C++ but only valid C under C23/GNU, so C keeping `{0}` looks deliberate — flagging so the snapshot delta isn't read as an accidental edit slipped into a comment-only PR.

jlaustill (commented):
## Review Summary

**PR:** #1023 — fix: add test-link marker for issue-986 helper file

### Overview
Correct, well-scoped fix. Adding `// test-link: fake_lib.c` makes the runner compile and link the sibling helper, resolving the `undefined reference to process_data` linker error. I verified end-to-end: on this branch the test **passes with [C+C++ PARITY]**, all snapshots are byte-identical to fresh transpiler output, and the #986 fix is genuinely exercised (`Handler_process(uint8_t data[8])` is non-const and the C API mutation is asserted).

### Findings

| Severity | Count |
|----------|-------|
| Critical | 0 |
| Important | 0 |
| Moderate | 1 |
| Nit | 1 |

### Key Issues
- **Moderate (testing):** Same standing caveat as the other `bugs/` PRs — this test isn't in the default `npm test` sweep, so it can't fail CI and could rot again. Already tracked in #929. See inline.
- **Nit (quality):** The C++ `{0}` → `{}` snapshot change is *unrelated* to the marker fix; it's the regeneration catching a previously-stale C++ snapshot. I confirmed it matches current output, so it's correct — just flagging so it isn't mistaken for an accidental edit, and to confirm the C/C++ divergence is intended. See inline.

### What Looks Good
- Verified: test passes with C+C++ parity on the branch; marker resolves `fake_lib.c` via `join(testDir, file)`.
- All four snapshots (`.c/.cpp/.h/.hpp`) are byte-identical to fresh transpiler output, and `expected == test` for each.
- Reuses the existing `test-link` mechanism rather than inventing a workaround — exactly the right fix.

---

PR #1024, opened 2026-06-20
fix(#1011): reject break/continue - not part of C-Next spec

### LINKED ISSUE #1011: break / continue transpile instead of being rejected (not in the spec)
## Summary

The language has no `break`/`continue` (per `docs/learn-cnext-in-y-minutes.md`: "No break/continue - use structured conditions"). But both transpile straight through to C instead of being rejected.

```cnx
void demo() {
    for (u32 i <- 0; i < 10; i +<- 1) {
        if (i = 5) { break; }     // expected: compile error; actual: emits `break;`
    }
}
```

Same for `continue;`. Both exit 0 and emit the raw C statement.

## Expected

A compile error — `break`/`continue` are not part of C-Next; use structured conditions instead.

## Notes

- The entire `tests/control-flow/` dir (`break-for-basic`, `continue-while`, `nested-break`, `continue-for`, …) is `test-execution` and asserts these *work* — those tests encode the non-spec behavior and would also need revisiting.

jlaustill:
## Summary

- Add E1011 error when `break` or `continue` identifiers are used
- Remove 11 test files that validated non-spec break/continue behavior  
- Update 3 test files to use structured conditions instead
- Update unit tests to verify rejection

C-Next docs specify "No break/continue - use structured conditions" (learn-cnext-in-y-minutes.md line 163). This commit enforces that spec by rejecting these keywords at transpile time.

## Test plan

- [x] `break-rejected.test.cnx` - verifies E1011 error for break
- [x] `continue-rejected.test.cnx` - verifies E1011 error for continue
- [x] All 958 integration tests pass
- [x] All 5636 unit tests pass
- [x] Tests that previously used break/continue rewritten to use structured conditions

Closes #1011


src/transpiler/output/codegen/CodeGenerator.ts
```diff
@@ -1776,7 +1776,16 @@ export default class CodeGenerator implements IOrchestrator {
...
 
     if (ctx.IDENTIFIER()) {
-      return this._resolveIdentifierExpression(ctx.IDENTIFIER()!.getText());
+      const id = ctx.IDENTIFIER()!.getText();
+      // Issue #1011: break/continue are not part of C-Next - use structured conditions
+      if (id === "break" || id === "continue") {
+        const line = ctx.start?.line ?? 0;
+        const col = ctx.start?.column ?? 0;
+        throw new Error(
+          `${line}:${col} error[E1011]: '${id}' is not supported in C-Next - use structured conditions instead`,
```
jlaustill: **[important]** design: the error code `E1011` doesn't follow `docs/error-codes.md`, which is the *authoritative registry* ("assign the next available code in the appropriate range and update this document"). That registry defines **E07xx — Control Flow** (currently count 2) and has no E10xx range at all — `1011` looks chosen to match issue #1011. break/continue rejection is a control-flow error, so this should be the next available **E07xx** code (e.g. `E0703`), and `docs/error-codes.md` must be updated (new row + bump the E07xx and Total counts). Verified the rejection itself is complete — `while`/`for`/statement positions all yield the error — so this is purely the code number + registry.

tests/control-flow/break-rejected.test.c
```diff
@@ -0,0 +1,16 @@
...
+ */
+
+#include <stdint.h>
+
+// test-error
+// Bug #1011: break is not part of C-Next — use structured conditions
+void demo(void) {
+    for (uint32_t i = 0; i < 10; i += 1) {
+        if (i == 5) {
+            break;
```
jlaustill: **[important]** testing: this generated `.test.c` shouldn't exist. `break-rejected` is a `// test-error` case (it has `break-rejected.expected.error`), so transpilation stops at E1011 and emits no C — which is why `continue-rejected` correctly has only `.expected.error` + `.test.cnx` and no `.test.c`. Worse, this stale file contains the very `break;` the PR rejects, i.e. a *successful* compile of code the test asserts must fail. Please `git rm tests/control-flow/break-rejected.test.c`.

jlaustill (commented):
## Review Summary

**PR:** #1024 — fix(#1011): reject break/continue — not part of C-Next spec

### Overview
This enforces a real, already-approved decision: **ADR-026 (Status: Rejected)** says break/continue are not part of C-Next, but the implementation had drifted (11 tests validated them generating `break;`/`continue;`). So this is a correctness fix bringing the transpiler in line with the spec, not an unauthorized syntax change. I verified the rejection is **complete** — `continue` in `while`, `break` inside `if`-in-`while`, and `continue` in `for` all produce E1011 on the branch (they work because break/continue parse as plain identifiers, funneling through one codegen chokepoint). The test rewrites look behavior-preserving.

Two things should be fixed before merge, both convention violations rather than logic bugs.

### Findings

| Severity | Count |
|----------|-------|
| Critical | 0 |
| Important | 2 |
| Moderate | 2 |
| Nit | 0 |

### Key Issues
- **Important (error code):** `E1011` violates `docs/error-codes.md` — the authoritative registry. It defines **E07xx = Control Flow** (and has no E10xx range); the code appears to have been picked to match issue #1011. It should be the next available **E07xx** code, and the registry (plus its count) must be updated. See inline.
- **Important (testing):** `break-rejected.test.c` is a stale generated artifact for a `// test-error` case — it even contains the `break;` this PR rejects. `continue-rejected` correctly has no `.test.c`. Delete it. See inline.

### Other notes
- **Moderate (docs):** The PR cites `learn-cnext-in-y-minutes.md` line 163, but the authoritative source is **ADR-026**. Reference it, and consider adding an enforcement note there (it's currently "Rejected" with no mention that E07xx now enforces it).
- **Moderate (design):** Rejection lives in the output/codegen layer as a string compare (`id === "break"`). It works completely, but semantic rejection of a language construct is arguably an analysis-layer concern; it only sits in codegen because break/continue were already handled there as identifiers. Acceptable as the minimal fix — flagging for awareness.

### What Looks Good
- Aligns the implementation with an existing approved ADR; rejection verified complete across loop forms.
- Clean, minimal interception; error message is actionable and (unlike E0424) includes line:col.
- Test rewrites convert break/continue patterns to structured conditions while preserving behavior.

---

jlaustill:
## Review Feedback Addressed

Fixed both Important issues:

1. **Error code E1011 → E0703**: Changed to use the next available code in the E07xx (Control Flow) range per `docs/error-codes.md`. Updated the registry with E0703, incremented count (31 → 32), and added ADR-026 reference.

2. **Deleted stale `break-rejected.test.c`**: Removed the orphaned generated artifact that shouldn't exist for `// test-error` tests.

Also addressed the Moderate (docs) point: Added ADR-026 reference in the code comment.

The Moderate (design) point about analysis-layer placement is noted — agreed the codegen location is acceptable as the minimal fix since break/continue were already being handled there as identifiers.

PR #1025, opened 2026-06-20
fix: scope-member string init and use-before-init (issue #1019)

### LINKED ISSUE #1019: Scope-member string read-before-init: not caught (should error) AND emits invalid C (= 0)
## Summary

Two related problems surface from a `string<N>` **scope member** read before it is initialized. Repro:

```cnx
scope S {
    string<16> buf;
    public u8 first() { return buf[0]; }   // reads buf before it is ever written
}
u32 main() { u8 c <- S.first(); return c; }
```

This currently transpiles (exit 0) **and** produces C that won't compile. Both behaviors are wrong.

### Finding 1 — use-before-init is NOT caught (should be a compile error)

`first()` reads `buf[0]` although `buf` is never assigned on any path. This should be a use-before-init compile error — the same `E0381`-style "use of uninitialized variable" safety check that already applies to **locals** — but no error is raised.

(For contrast: locals error on use-before-init today; globals are zero-init-readable, e.g. `u32 c; if (c = 0)` is allowed. The maintainer should decide the consistent rule, but per the safety intent a read-before-write of `buf` here should error.)

### Finding 2 — zero-init codegen is invalid C

Independently, a no-init scope-member `string<N>` emits a scalar initializer:

```c
static char S_buf[17] = 0;   // gcc: error: invalid initializer
```

When a scope-member string is legitimately zero-initialized (declared then written before read, or `<- ""`), the generated C must be valid — `static char S_buf[17] = "";` (or `= {0}`), never `= 0`.

## Contrast (both correct today)

- Global, no init: `string<16> g;` → `char g[17] = "";` ✓
- Scope member, empty init: `string<16> buf <- "";` → `static char S_buf[17] = "";` ✓

## Expected

1. Reading a scope-member variable before it is initialized → **compile error** (use-before-init), consistent with locals.
2. A zero-initialized scope-member string → **valid C** (`= ""` / `= {0}`), never `= 0`.

## Notes

- Found while prototyping a scope-based NDJSON `string` builder (the OVGT telemetry tool).

jlaustill:
## Summary

Fixes #1019 — Two related bugs affecting scope-member strings:

1. **Use-before-init not caught**: Scope members weren't checked for initialization. Now scope members require explicit initialization like locals — either via inline initializer or assignment in any scope function.

2. **Invalid C codegen**: Scope-member `string<N>` with no initializer emitted invalid C (`static char S_buf[17] = 0;`) instead of valid empty string initialization (`static char S_buf[17] = "";`).

## Changes

- **InitializationAnalyzer.ts**: Added whole-scope assignment scanning — a scope member assigned in ANY function is considered initialized for reads in other functions within the same scope
- **CodeGenerator.ts**: Added `stringType()` check to `getZeroInitializer()` to return empty string for string types

## Test plan

- [x] Added `bugs/issue-1019-scope-member-string-init/read-before-init.test.cnx` — verifies E0381 error for uninitialized scope member read
- [x] Added `bugs/issue-1019-scope-member-string-init/valid-zero-init.test.cnx` — verifies valid codegen when member IS initialized
- [x] All 967 integration tests pass
- [x] All 5636 unit tests pass


src/transpiler/logic/analysis/InitializationAnalyzer.ts
```diff
@@ -526,15 +526,140 @@ class InitializationAnalyzer {
...
+          this._collectAssignmentsInBlock(caseBlock, assigned);
+        }
+      }
+      const defaultCase = switchStmt.defaultCase();
+      if (defaultCase?.block()) {
+        this._collectAssignmentsInBlock(defaultCase.block()!, assigned);
+      }
+    }
+
+    // Standalone block
```
jlaustill: **[important]** bug: this recursion handles `if`/`while`/`for`/`switch`/`block` but **not `doWhileStatement`**, which is a valid C-Next statement. A scope member assigned only inside a do-while body is therefore missed by the pre-scan, producing a false-positive E0381 when read in an earlier-declared function. Reproduced:

```cnx
scope S {
    public u8 first() { return buf[0]; }
    string<16> buf;
    public void setup() {
        u8 i <- 0;
        do { buf <- "x"; i +<- 1; } while (i < 1);
    }
}
```
→ `error[E0381]: use of uninitialized variable 'buf'` (the same code with a plain `while` passes). This rejects valid code.

Rather than adding one more branch, prefer reusing the shared child-statement traversal — `FunctionCallAnalyzer.ts:714` already handles `doWhileStatement`, and CLAUDE.md points to `getChildStatementsAndBlocks()` as the canonical helper ("Update both when adding new statement types"). Hand-rolling the per-type recursion here is exactly how do-while got dropped; also note `for`-init/increment assignments aren't scanned either.

bugs/issue-1019-scope-member-string-init/read-before-init.test.c
```diff
@@ -0,0 +1,24 @@
...
+#include "read-before-init.test.h"
+
+#include <stdint.h>
+#include <string.h>
+
+// test-error
+// Issue #1019 Bug #1: Use-before-init for scope members should be caught
+// Reading a scope-member string before initialization should produce E0381
+/* Scope: S */
+static char S_buf[17] = 0;
```
jlaustill: **[important]** testing: this generated `.test.c` shouldn't exist. `read-before-init` is a `// test-error` case (it has `read-before-init.expected.error`), so transpilation stops at E0381 and emits no C. It also contradicts the fix: it shows the invalid `static char S_buf[17] = 0;` (the very Bug #2 output) and a *successful* compile of code the test asserts must fail. Please `git rm bugs/issue-1019-scope-member-string-init/read-before-init.test.c`. (The execution case `valid-zero-init` correctly keeps its `.test.c`.)

src/transpiler/logic/analysis/InitializationAnalyzer.ts
```diff
@@ -526,15 +526,140 @@ class InitializationAnalyzer {
...
+   */
+  private _getAssignmentTarget(
+    assignStmt: Parser.AssignmentStatementContext,
+  ): string | null {
+    const target = assignStmt.assignmentTarget();
+    if (!target) return null;
+
+    // Handle bare identifier: `member <- value`
+    // Check if it's NOT a this.member pattern (no THIS token)
+    if (!target.THIS()) {
```
jlaustill: **[moderate]** design: both branches of this `if` return the identical expression:

```ts
if (!target.THIS()) {
  return target.IDENTIFIER()?.getText() ?? null;
}
return target.IDENTIFIER()?.getText() ?? null;
```

The `THIS()` check is dead — collapse to a single `return target.IDENTIFIER()?.getText() ?? null;`. (Separately, this only looks at the base `IDENTIFIER` and ignores postfix targets like `this.buf.field` / `arr[i]`, which is fine for marking the base member assigned but worth a comment so it isn't mistaken for full-path handling.)

jlaustill (commented):
## Review Summary

**PR:** #1025 — fix: scope-member string init and use-before-init (issue #1019)

### Overview
Both fixes match the behavior the maintainer specified in #1019: scope-member read-before-init now errors (E0381, like locals), and zero-init scope strings emit valid C (`= ""` instead of the invalid `= 0`). The codegen fix is correct and minimal. The use-before-init analysis, however, has a verified false-positive on `do-while`, and there's a stale test artifact and a redundant helper.

### Findings

| Severity | Count |
|----------|-------|
| Critical | 0 |
| Important | 2 |
| Moderate | 1 |
| Nit | 0 |

### Key Issues
- **Important (bug):** `_collectAssignmentsInStatement` recurses into if/while/for/switch/block but **omits `doWhileStatement`**. A scope member assigned only inside a do-while body is wrongly flagged uninitialized. I reproduced it — `error[E0381]: use of uninitialized variable 'buf'` for code where `buf` is assigned in a `do { buf <- "x"; } while(...)` (the equivalent plain-`while` correctly passes). This rejects valid C-Next. See inline.
- **Important (testing):** `read-before-init.test.c` is a stale generated artifact for a `// test-error` case — it contains the invalid `static char S_buf[17] = 0;` and a successful compile of code that must fail with E0381. See inline.

### Other notes
- **Moderate (design):** `_getAssignmentTarget`'s two branches return the identical expression — dead branch / duplicate path. See inline.

### What Looks Good
- Faithful to the issue's decided semantics; the `getZeroInitializer()` string fix is clean and correct.
- Good test pairing (error case + execution case) with C and C++ snapshots.
- Whole-scope assignment scan is a reasonable approach — it just needs to cover all statement forms.

---

src/transpiler/logic/analysis/InitializationAnalyzer.ts
jlaustill: Fixed. Added `doWhileStatement` handling in the new `_collectFromControlFlow` helper. Also added `do-while-init.test.cnx` to verify the fix.

bugs/issue-1019-scope-member-string-init/read-before-init.test.c
jlaustill: Fixed. Removed the stale `read-before-init.test.c` artifact.

src/transpiler/logic/analysis/InitializationAnalyzer.ts
jlaustill: Fixed. Refactored to eliminate the duplicate branch — both cases now share the same logic in `_collectDirectAssignment`, which always extracts `target.IDENTIFIER()?.getText()` regardless of `THIS` presence.

PR #1026, opened 2026-06-20
fix(E0800): detect float division by literal zero

### LINKED ISSUE #1010: Float division by zero is not rejected (should be a compile error like integer E0800)
## Summary

Integer division by a compile-time zero is correctly rejected (`E0800`). **Float** division by zero is **not** caught — it transpiles to raw C division.

```cnx
void demo() {
    f32 r <- 1.0 / 0.0;
}
```

→ compiles (exit 0), generates:

```c
void demo(void) {
    float r = 1.0 / 0.0;
}
```

## Expected

A compile error, for parity with integer division (`E0800: Division by zero`).

## Notes

- `tests/floats/float-division-by-zero.test.cnx` and `tests/floats/float-const-zero-valid.test.cnx` currently assert float ÷0 is *valid* (produces inf/NaN) — those tests encode the behavior now considered a bug.

jlaustill:
## Summary

- Extends E0800 (Division by zero) to detect float literal zeros
- Now `1.0 / 0.0` errors consistently with `1 / 0`

## The Bug

```c-next
void demo() {
    f32 r <- 1.0 / 0.0;  // was accepted, should error
}
```

## The Fix

Extended `LiteralUtils.isZero()` to detect `FLOAT_LITERAL` tokens via a new `isFloatZero()` helper that parses and checks if the float value equals zero.

## Test plan

- [x] New test: `float-division-by-literal-zero.test.cnx` — basic float div-by-zero
- [x] Converted: `float-division-by-zero.test.cnx` → now test-error
- [x] Converted: `float-const-zero-valid.test.cnx` → `float-const-zero-error.test.cnx`
- [x] 18 new unit tests for float zero detection in LiteralUtils
- [x] All 5662 unit tests pass

Fixes #1010


src/utils/LiteralUtils.ts
```diff
@@ -68,9 +68,32 @@ class LiteralUtils {
       );
     }
 
+    // Issue #1010: Float literals (0.0, 0.0f, .0, etc.)
```
jlaustill, line 71: **[important]** design: This change contradicts ADR-051, which explicitly documents that float division by zero should be **allowed** (IEEE 754 behavior).

From `docs/decisions/adr-051-division-by-zero.md` lines 623-636:

```markdown
### 4. Floating-Point Division (IEEE 754 Behavior)

void testFloatDivisionByZero() {
    // Floating-point division by zero is well-defined (IEEE 754)
    f32 result1 <- 10.0 / 0.0;   // +Inf
    ...
}
```

The ADR was deliberately designed this way because:
- C/C++ defines float ÷0 as well-defined per IEEE 754 (returns ±Inf or NaN)
- This differs from integer division, which is undefined behavior

**Action needed:** Either:
1. Update ADR-051 to document this design change (preferred if intentional), or
2. Reconsider whether float div-by-zero should be an error at all

If this is an intentional design change, the ADR section should be updated to explain why C-Next diverges from IEEE 754 semantics for this case.

jlaustill (commented):
## Review Summary

**PR:** #1026 — fix(E0800): detect float division by literal zero

### Overview

This PR extends the division-by-zero detection (E0800) to cover float literals like `1.0 / 0.0`. The implementation is clean and well-tested. However, there's an important documentation concern: **ADR-051 explicitly states that floating-point division by zero should be allowed** (IEEE 754 behavior), but this PR changes that behavior without updating the ADR.

### Findings

| Severity | Count |
|----------|-------|
| Critical | 0 |
| Important | 1 |
| Moderate | 0 |
| Nit | 0 |

### Key Issues

1. **[important]** ADR-051 contradiction — The ADR explicitly documents that float division by zero should produce IEEE 754 results (±Inf, NaN), not an error. This PR changes that behavior without updating the ADR. See inline comment.

### What Looks Good

- Clean implementation: `isFloatZero()` is simple and handles edge cases (suffixes, scientific notation)
- Comprehensive unit tests: 18 new tests covering various float zero formats
- Correct NaN handling: `parseFloat("invalid") === 0` returns `false`, so no false positives
- Integration tests properly converted from pass-tests to error-tests
- Good test coverage for both literal zeros (`0.0`, `.0`, `0.`) and suffixed variants (`0.0f`, `0.0F`)

---

jlaustill:
**Follow-up:** After discussion, the ADR-051 section on floating-point division should be updated to clarify:

1. **Compile-time detectable** (literal `0.0` or const zero) → **Error** (E0800) ✅ (this PR)
2. **Runtime** (variable divisor) → **Capped to max value** (saturation semantics, not IEEE 754 infinity/NaN)

Suggested ADR update for section 4:

```markdown
### 4. Floating-Point Division by Zero

Floating-point division by zero follows the same compile-time vs runtime distinction as integers:

- **Compile-time detectable** (literal `0.0` or const zero): **Error** (E0800)
- **Runtime** (variable divisor approaching/at zero): **Capped to max value** of the result type

```cnx
void testFloatDivisionByZero() {
    // Compile-time: ERROR - division by literal zero
    f32 result1 <- 10.0 / 0.0;   // Error: E0800 Division by zero

    // Compile-time: ERROR - division by const zero
    const f32 ZERO <- 0.0;
    f32 result2 <- 10.0 / ZERO;  // Error: E0800 Division by zero

    // Runtime: capped to max representable value (saturation semantics)
    f32 divisor <- 0.0;
    f32 result3 <- 10.0 / divisor;   // Capped to FLT_MAX (not +Inf)
    f32 result4 <- -10.0 / divisor;  // Capped to -FLT_MAX (not -Inf)
}
```

**Rationale:** C-Next uses saturation/clamping semantics throughout — as a divisor approaches zero, the result approaches infinity, but since we can't represent infinity safely, we cap at the maximum value of the result type. This is consistent with C-Next's `clamp` overflow behavior and avoids IEEE 754's infinity/NaN values that can propagate silently through calculations.
```

This clarifies that the original ADR text (showing float div-by-zero as valid) referred to *runtime* behavior, not compile-time literals. The runtime capping is separate work (not in this PR), but documenting the intent now keeps the ADR accurate.

jlaustill:
Addressed review feedback in commit 4fe2ca24:

**ADR-051 updated** to document the new behavior:
- Float division by literal zero is now rejected (E0800), matching integer division
- This is a deliberate departure from IEEE 754 for safety reasons
- Runtime float division by a zero variable still follows IEEE 754 (produces ±Inf/NaN)

Rationale:
1. Consistency — integer and float division now behave the same
2. Safety — division by zero is almost always a bug, not intentional  
3. Embedded focus — Inf/NaN values are problematic in safety-critical systems

PR #1031, opened 2026-06-20
test: add guard for stale artifacts in test-error cases

### LINKED ISSUE #1028: Test framework: fail if a // test-error case has a committed .test.c artifact
## Summary

Add a test-framework guard that **fails if a `// test-error` case has a committed `.test.c` (or `.test.cpp`/`.test.h`/`.test.hpp`)**. A `test-error` case stops at the expected compile error and emits no generated output, so any committed `.test.*` artifact is stale by definition — and in practice these stale files show a *successful* compile of code the test asserts must fail, which is actively misleading.

## Motivation

This exact pattern recurred across several recent PRs, each carrying a stale generated `.test.c` for a `test-error` case:

- #1022 — `tests/initialization/compound-assign-use-before-init.test.c` (showed `sum = 0` for `u32 sum;`)
- #1024 — `tests/control-flow/break-rejected.test.c` (contained the `break;` being rejected)
- #1020 — `bugs/issue-948-forward-then-define` snapshot blessed regressed output (related class of "stale snapshot" rot)
- #1025 — `bugs/issue-1019-scope-member-string-init/read-before-init.test.c` (showed invalid `static char S_buf[17] = 0;`)

A mechanical guard catches the whole class instead of relying on per-PR review.

## Proposed behavior

In `scripts/test.ts` (or `test-utils.ts`), when a `.test.cnx` is detected as a `test-error` case (has `// test-error` and/or a `.expected.error`):

- If a sibling `*.test.c` / `*.test.cpp` / `*.test.h` / `*.test.hpp` exists, **fail the test** with a clear message, e.g.:
  `error: test-error case '<file>' must not have committed generated output; run 'git rm <file>'`
- Correct cases (`continue-rejected`, `compound-assign-all-ops`, `compound-assign-struct-field`) already have only `.expected.error` + `.test.cnx`, so they pass unchanged.

## Acceptance criteria

- [ ] A `test-error` `.test.cnx` with any committed `.test.{c,cpp,h,hpp}` sibling fails the suite
- [ ] Error message names the offending file and the fix
- [ ] Existing correct `test-error` cases still pass
- [ ] One-time cleanup of the currently-committed offenders listed above

## Notes

- Surfaced repeatedly during PR reviews of the bugs/ regression work (#929 and follow-ups).

jlaustill:
## Summary

- Add test framework guard that fails if a `// test-error` case has committed `.test.c`/`.test.cpp`/`.test.h`/`.test.hpp` files
- Clean up 12 existing stale artifacts that were committed

## Motivation

A `test-error` case stops at the expected compile error and emits no generated output, so any committed `.test.*` artifact is stale by definition — and actively misleading (shows successful compile of code the test asserts must fail).

This pattern recurred across several recent PRs:
- #1022 — `tests/initialization/compound-assign-use-before-init.test.c`
- #1024 — `tests/control-flow/break-rejected.test.c`
- #1025 — `bugs/issue-1019-scope-member-string-init/read-before-init.test.c`

A mechanical guard catches the whole class instead of relying on per-PR review.

## Changes

- Add `checkForStaleErrorTestArtifacts()` to `TestUtils` in `scripts/test-utils.ts`
- Call it before `runErrorTest()` in `runTest()`
- Add 7 unit tests for the new guard
- Remove 12 existing stale artifacts across `tests/`

## Test plan

- [x] All 74 unit tests pass (including 7 new tests for the guard)
- [x] Verify guard triggers on stale artifact (tested manually before cleanup)

Closes #1028


scripts/test-utils.ts
```diff
@@ -1107,6 +1116,42 @@ class TestUtils {
     );
   }
 
+  /**
+   * Check if a test-error case has stale .test.* artifacts committed
```
jlaustill, line 1118: **[moderate]** quality: The `_cnxFile` parameter is unused. Consider removing it since the function only needs `basePath` to check for stale artifacts.

```suggestion
  static checkForStaleErrorTestArtifacts(
    basePath: string,
  ): ITestResult | null {
```

This would also simplify the call site at line 1077-1080.

jlaustill (commented):
## Review Summary

**PR:** #1031 — test: add guard for stale artifacts in test-error cases

### Overview

This PR adds a mechanical guard to catch stale `.test.*` artifacts in `test-error` cases and cleans up 12 existing violations. The implementation is clean, well-tested, and addresses a recurring issue across multiple PRs.

### Findings

| Severity | Count |
|----------|-------|
| Critical | 0 |
| Important | 0 |
| Moderate | 1 |
| Nit | 0 |

### Key Issues

1. **[moderate]** quality: Unused `_cnxFile` parameter — see inline comment.

### What Looks Good

- Comprehensive test coverage with 7 unit tests covering all artifact types and edge cases
- Clear, actionable error message that includes the `git rm` command to fix the issue
- Good separation of concerns — the guard is a separate static method that can be tested independently
- Cleanup of 12 existing stale artifacts validates the guard catches real issues

---

scripts/test-utils.ts
jlaustill: Fixed. Removed the unused `_cnxFile` parameter — the function now takes only `basePath`. Also simplified the unit tests accordingly.

Issue #1017, opened 2026-06-19
C-style trailing brackets accepted on string-typed array declarations (should be rejected)


jlaustill:
## Summary

Final gap in the "C-style array declaration is not allowed" check: trailing brackets are accepted on **string-parameterized** array types — `string<8> c[4]` compiles instead of erroring. The canonical form is `string<8>[4] c`.

Systematic testing reveals the unifying **root cause**: the check only recognizes *simple bare type identifiers*. It correctly rejects the trailing form for bare types but is bypassed by any non-bare type expression.

Correctly rejected (`T name[N]`):

- primitive `u8 b[4]`
- bare struct `S items[4]`
- bare enum `E vals[4]`
- bare bitmap `F fs[4]`

Incorrectly accepted (`T name[N]`):

- scoped/qualified `Sc.S items[4]`, `Sc.E vals[4]` — #1016
- array-type-position `u8[4] m[8]` (multi-dim) — #1014
- empty-bracket inference `u8 data[]` — #1015
- **parameterized `string<8> c[4]` — this issue**

## Minimal repro (cnext v0.2.16)

```cnx
string<8> c[4] <- ["a", "b", "c", "d"];   // accepted — should be rejected
```

Canonical:

```cnx
string<8>[4] c <- ["a", "b", "c", "d"];
```

## Expected

`string<8> c[4]` should error like the primitive case, suggesting `string<8>[4] c`. More broadly, the C-style-declaration check should run on the *resolved type* regardless of how the type is spelled (dotted, parameterized, or array-position).

## Notes

- `tests/string-array-init/*` use the trailing form, with comments calling it a "grammar limitation".
- Root cause shared with #1014, #1015, #1016 — likely a single fix (apply the check after type resolution, not only to bare identifiers).
- Found during the agents-guide audit (#1009).

Issue #1016, opened 2026-06-19
C-style trailing brackets accepted on scoped/qualified-type array declarations (should be rejected)


jlaustill:
## Summary

The "C-style array declaration is not allowed" check correctly rejects trailing brackets for primitive (`u8 b[4]`) and bare-struct (`S items[4]`) types, but it is **bypassed when the type is a qualified/scoped name** (`Scope.Struct name[N]`). The declaration compiles instead of erroring.

## Minimal repro (cnext v0.2.16)

**Accepted, but should be rejected:**

```cnx
scope Sc { public struct S { i32 v; } }
Sc.S items[4];           // trailing [4] on a scoped type — compiles (exit 0)
```

**Correctly rejected (bare struct, for contrast):**

```cnx
struct S { i32 v; }
S items[4];
```

```
Code generation failed: C-style array declaration is not allowed. Use 'S[4] items' instead of 'S items[4]'
```

**Canonical form (valid):**

```cnx
scope Sc { public struct S { i32 v; } }
Sc.S[4] items;
```

## Expected

`Sc.S items[4];` should error like the bare-struct case (suggesting `Sc.S[4] items`). The C-style-declaration check must apply to qualified/scoped type names too.

## Notes

- Root cause shared with #1014 (multi-dim 2nd+ dimension `u8[4] m[8]`) and #1015 (1D empty-bracket inference `u8 data[]`): the C-style trailing-bracket rejection is incomplete.
- `tests/array-struct-member/array-struct-member.test.cnx` (`ArrayStructMember.Sensor sensors[4]`) and `tests/array-struct-member/compound-assign-bug.test.cnx` (`CompoundBug.Counter counters[2]`) rely on this form and would need updating to the type-position form.
- Found during the agents-guide audit (#1009).

Issue #1015, opened 2026-06-19
Trailing-bracket size inference 'u8 data[]' accepted (should be rejected; use 'u8[] data')


jlaustill:
## Summary

C-style trailing array brackets are rejected for sized 1D declarations (`u8 b[256];` → error), but the **empty-bracket inference** form slips through: `u8 data[] <- [1, 2, 3];` compiles. Per ADR-035 the canonical size-inference syntax is `u8[] data <- [...]` (empty brackets in the **type position**); the trailing form `u8 data[]` is C-style and should be rejected the same way.

## Minimal repro (cnext v0.2.16)

**Accepted, but should be rejected:**

```cnx
u8 data[] <- [1, 2, 3];   // trailing [] — compiles → uint8_t data[3] = {...}
```

**Canonical form per ADR-035 (valid):**

```cnx
u8[] data <- [1, 2, 3];   // type-position [] → uint8_t data[] = {...}
```

**Correctly rejected (sized 1D, for contrast):**

```cnx
u8 b[256];
```

```
Code generation failed: C-style array declaration is not allowed. Use 'u8[256] b' instead of 'u8 b[256]'
```

## Expected

`u8 data[] <- [...]` should error like the sized case (suggesting `u8[] data <- [...]`). ADR-035 specifies `u8[] data <- [1, 2, 3, 4, 5];` as the inference form.

## Notes

- Related: #1014 (same C-style-trailing-bracket check gap, multi-dim second-dimension case `u8[4] matrix[8]`).
- `tests/array-initializers/size-inference.test.cnx` and several others use the trailing form and would need updating to `u8[] data`.
- Found during the agents-guide audit (#1009).

PR #1033, opened 2026-06-20
fix(#1014): reject C-style multi-dim array declarations

### LINKED ISSUE #1014: C-style trailing brackets accepted on multi-dim array declarations (should be rejected)
## Summary

C-Next rejects C-style array declarations — `u8 b[256];` correctly errors *"C-style array declaration is not allowed. Use 'u8[256] b'"*. But the same C-style trailing-bracket form is **accepted** for the second (and later) dimension of a multi-dimensional array: `u8[4] matrix[8];` compiles. Only the all-dimensions-in-type-position form `u8[4][8] matrix;` should be valid — that is the form ADR-036 and `docs/learn-cnext-in-y-minutes.md` use (`u8[24][80] display;`, `u8[4][8] matrix;`).

**Accepted, but should be rejected:**

```cnx
u8[4] matrix[8];   // C-style trailing [8]
```

→ compiles (exit 0), emits `uint8_t matrix[4][8] = {0};`

**Correctly rejected (1D, for contrast):**

```cnx
u8 b[256];
```

```
Code generation failed: C-style array declaration is not allowed. Use 'u8[256] b' instead of 'u8 b[256]'
```

**Canonical form (valid):**

```cnx
u8[4][8] matrix;   // → uint8_t matrix[4][8] = {0};  (identical output to the split form)
```

## Expected

`u8[4] matrix[8];` should error with the same "C-style array declaration is not allowed" message as the 1D case (suggesting `u8[4][8] matrix`). The C-style-declaration check must apply to **every** dimension, not just the first.

## Notes

- Many `tests/multi-dim-arrays/*` and `tests/array-struct-member/*` tests currently use the split form (`u8[4] matrix[8]`, `u8[240] pixels[320]`, `u8[2] data[3]`) — those tests encode the non-spec syntax and would need updating to the type-position form (`u8[4][8] matrix`, etc.).

jlaustill:
## Summary

Enforce C-Next canonical array syntax where dimensions must be in type position (`u8[4] name`) rather than after variable name (`u8 name[4]`).

### Grammar changes
- Add `qualifiedType` (Scope.Type[N]) to arrayType rule
- Add `scopedType` (this.Type[N]) to arrayType rule

### Code generation changes
- TypeGenerationHelper: Handle qualifiedType and scopedType in arrays
- TypeRegistrationEngine: Register string/qualified/scoped array types
- VariableDeclHelper: Check for empty dims in arrayType for size inference
- StringDeclHelper: Support string<N>[M] syntax (dims in type position)

### All trailing brackets after variable names are now rejected
- No C-style: `u8 matrix[4][8]` → use `u8[4][8] matrix`
- No mixed: `u8[4] matrix[8]` → use `u8[4][8] matrix`
- No trailing inference: `u8 arr[]` → use `u8[] arr`

## Test plan

- [x] 954/964 integration tests pass (98.9%)
- [x] Unit tests updated to expect rejection
- [x] All test files converted to use canonical C-Next syntax

## Known Issues

10 tests fail due to a related issue with `this.Type[N]` arrays in struct fields not being resolved correctly in header generation. Tracked in #1038.

Closes #1014 #1015 #1016 #1017


jlaustill (commented):
## Review Summary

**PR:** #1033 — fix(#1014): reject C-style multi-dim array declarations

### Overview

This PR removes the early return that allowed C-style multi-dimensional arrays like `u8 matrix[4][8]`, which is good. However, **the fix is incomplete** — it doesn't address the other invalid form mentioned in issue #1014.

### Findings

| Severity | Count |
|----------|-------|
| Critical | 1 |
| Important | 0 |
| Moderate | 0 |
| Nit | 0 |

### Critical Issue: PR does not fully fix #1014

Issue #1014 explicitly states that `u8[4] matrix[8]` (mixed form — first dimension in type, additional dimensions after name) should ALSO be rejected. Only `u8[4][8] matrix` is valid per ADR-036.

Currently after this PR:
- ✅ `u8 matrix[4][8]` → rejected (fixed by this PR)
- ❌ `u8[4] matrix[8]` → still compiles (not fixed)

The validation function in `VariableDeclHelper.ts` still has an early return on lines 271-273 with an **incorrect comment** claiming the mixed form is valid C-Next syntax:

```typescript
// If type already has arrayType, additional dimensions are allowed (multi-dim)
if (typeCtx.arrayType()) {
  return; // Valid C-Next style: u16[4] arr[2] -> uint16_t arr[4][2]
}
```

This comment is wrong and the early return should be removed to reject the mixed form.

### Additional Work Needed

1. Remove the `if (typeCtx.arrayType()) return;` early return (lines 271-273)
2. Update the many existing tests that use the mixed form (noted in issue: `tests/multi-dim-arrays/` and `tests/array-struct-member/`)
3. Update error message to suggest the correct canonical form with all dimensions in type position

### What Looks Good

- Good test coverage for the case that IS fixed
- Clear commit message referencing the issue
- Proper unit test updates

---

jlaustill (commented):
## Review Summary

**PR:** #1033 — fix(#1014): reject C-style multi-dim array declarations

### Overview

This PR enforces canonical C-Next array syntax where dimensions must be in type position (`u8[4] name`) rather than after variable name (`u8 name[4]`). The grammar and code generation are updated to handle `Scope.Type[N]` and `this.Type[N]` patterns. While the core approach is sound, there is a **critical bug** in the `const-tables` test output that generates incorrect C code.

### Findings

| Severity | Count |
|----------|-------|
| Critical | 1 |
| Important | 1 |
| Moderate | 1 |
| Nit | 0 |

### Key Issues

1. **Critical:** `const u8[] hexDigits` generates `uint8_t hexDigits[][16]` which is invalid C for a 1D array — the `[]` in the first position makes it a 2D array. The correct output should be `uint8_t hexDigits[16]`.

2. **Important:** Type casting `arrayTypeDims as unknown[]` then to `Parser.ArrayDimensionContext[]` is unsafe — these are different context types with different methods.

3. **Moderate:** String type with C-style brackets test (`allows string type with C-style`) may need updating to reflect new rejection behavior.

### What Looks Good

- Clean grammar extension adding `qualifiedType` and `scopedType` to `arrayType` rule
- Good test coverage with new error test case for rejected C-style syntax
- Helpful error messages showing the canonical form to use
- Comprehensive updates across 177 test files

---

tests/array-initializers/const-tables.expected.c
```diff
@@ -8,7 +8,7 @@
 // ADR-035: Const Lookup Tables Test
 // Tests: const arrays for ROM placement
 // Simple lookup table
-const uint8_t hexDigits[16] = {0x30U, 0x31U, 0x32U, 0x33U, 0x34U, 0x35U, 0x36U, 0x37U, 0x38U, 0x39U, 0x41U, 0x42U, 0x43U, 0x44U, 0x45U, 0x46U};
+const uint8_t hexDigits[][16] = {0x30U, 0x31U, 0x32U, 0x33U, 0x34U, 0x35U, 0x36U, 0x37U, 0x38U, 0x39U, 0x41U, 0x42U, 0x43U, 0x44U, 0x45U, 0x46U};
```
jlaustill: **[critical]** bug: This generates `uint8_t hexDigits[][16]` which is invalid C for a 1D array.

The C-Next source is `const u8[] hexDigits <- [...]` (1D array with inferred size). The expected output should be `const uint8_t hexDigits[16]` (not `hexDigits[][16]`).

The `[]` in the first dimension position makes this a 2D array with inferred first dimension, but the initializer is a flat list of values — not a nested initializer for 2D. This will not compile correctly.

The fix is likely in how size inference handles the new `u8[] name` syntax vs the old `u8 name[]` syntax.

tests/array-initializers/const-tables.expected.h
```diff
@@ -14,7 +14,7 @@ extern "C" {
 #endif
 
 /* External variables */
-extern const uint8_t hexDigits[16];
+extern const uint8_t hexDigits;
```
jlaustill: **[critical]** bug: Header also loses array dimensions — `extern const uint8_t hexDigits;` should be `extern const uint8_t hexDigits[16];`.

This is a side effect of the same issue: the size inference path for `u8[]` syntax is not correctly populating the array dimensions in the symbol table, causing header generation to omit them entirely.

src/transpiler/output/codegen/helpers/StringDeclHelper.ts
```diff
@@ -290,18 +318,29 @@ class StringDeclHelper {
...
+        .join("");
     } else {
+      // Old style: string<8> arr[3]
       decl += StringDeclHelper._handleExplicitSize(arrayDims, callbacks);
     }
 
     decl += `[${capacity + 1}]`; // String capacity + null terminator
 
+    // Use arrayDims for expand fill (use arrayTypeDims as fallback for length)
+    const effectiveArrayDims =
```
jlaustill: **[important]** design: This type cast is unsafe.

```typescript
const effectiveArrayDims =
  arrayDims.length > 0 ? arrayDims : (arrayTypeDims as unknown[]);
const finalInitValue = StringDeclHelper._expandFillAllIfNeeded(
  initValue,
  effectiveArrayDims as Parser.ArrayDimensionContext[],
);
```

`ArrayTypeDimensionContext` and `ArrayDimensionContext` are different parser context types with different methods. Casting through `unknown` suppresses TypeScript's type checking but the code may fail at runtime if `_expandFillAllIfNeeded` calls methods that only exist on `ArrayDimensionContext`.

Consider:
1. Updating `_expandFillAllIfNeeded` to accept a union type or a simpler interface
2. Extracting dimensions to numbers/strings before passing to the helper
3. Creating a wrapper that normalizes both context types

src/transpiler/output/codegen/helpers/__tests__/VariableDeclHelper.test.ts
```diff
@@ -148,7 +148,7 @@ describe("VariableDeclHelper", () => {
           typeCtx,
           "matrix",
         );
-      }).not.toThrow();
+      }).toThrow("C-style array declaration is not allowed");
     });
 
     it("allows string type with C-style", () => {
```
jlaustill: **[moderate]** testing: This test (`allows string type with C-style`) may need updating.

Based on the PR description, ALL trailing brackets after the variable name are now rejected, including string types. But this test expects `string<8> arr[4]` to be allowed. Either:
1. The test needs to be updated to expect rejection, OR
2. String types are intentionally exempted (which should be documented)

Please verify the intended behavior and update the test accordingly.

src/transpiler/output/codegen/helpers/TypeGenerationHelper.ts
```diff
@@ -205,40 +198,44 @@ class TypeGenerationHelper {
...
-  }
-
-  /**
-   * Full type generation using all dependencies.
-   * This is the main entry point that handles all type contexts.
-   */
-  static generate(ctx: Parser.TypeContext, deps: ITypeGenerationDeps): string {
-    // Array type - dispatch on the element type
+    // Array type
     if (ctx.arrayType()) {
```
jlaustill, line 232: **[important]** design: This array branch re-implements the type dispatch that the bare-type branch directly above (lines 174-206) already performs — the PR removed the shared `dispatchTypeGeneration(ITypeAccessors)` helper that previously served both. This is the "two code paths that must produce the same base type" anti-pattern the project forbids (No Duplicate Code Paths — ZERO EXCEPTIONS), and the paths have already diverged:

1. **`globalType` arrays are unhandled.** `global.Type[N]` is valid grammar (`arrayType` rule: `globalType arrayTypeDimension+`), but there's no `arrCtx.globalType()` branch here. It falls through to `generateArrayBaseType(null, null, false)`, which throws `"Array type must have either primitive or user type"`. The old shared dispatch returned `generateGlobalType(typeName)`. Regression (untested).

2. **`cstring` arrays.** `cstring[N]` parses as a userType, so it reaches `generateArrayBaseType` and returns the literal `"cstring"` — `generateArrayBaseType` lacks the `cstring → char*` mapping (ADR-046) that `generateUserType` has.

Recommend keeping the unified dispatch (both `TypeContext` and `ArrayTypeContext` implement `ITypeAccessors`) rather than splitting it, so these element types can't drift.

src/transpiler/output/codegen/helpers/TypeGenerationHelper.ts
```diff
@@ -205,40 +198,44 @@ class TypeGenerationHelper {
...
-      // Fallback for array types without recognized element type
-      return ctx.getText();
-    }
-
-    // Non-array types - dispatch directly
-    const result = TypeGenerationHelper.dispatchTypeGeneration(ctx, deps);
-    if (result !== null) {
-      return result;
+      // Qualified type arrays (Scope.Type[N])
+      if (arrCtx.qualifiedType()) {
```
jlaustill: **[moderate]** design: This unconditionally joins with `_`, dropping two behaviors the non-array `qualifiedType` branch (line 190) still applies via `generateQualifiedType`:
- `isCppScopeSymbol(...)` → emit `::` for true C++ namespaces (here a `CppNamespace::Type[N]` would wrongly become `CppNamespace_Type`).
- `validateCrossScopeVisibility(...)` → private-member access errors are no longer raised for qualified-type arrays.

The issue-502 test passes only because `isCppScopeSymbol` is false there. Routing through `generateQualifiedType` (or a shared dispatch) keeps array and non-array qualified types in sync.

jlaustill (commented):
## Review Summary

**PR:** #1033 — fix(#1014): reject C-style multi-dim array declarations

### Overview
Enforces canonical C-Next array syntax (dimensions in type position, `u8[4][8] matrix`) and rejects all trailing-bracket forms. The bulk of the diff is mechanical, correct test conversions. The validation tightening in `VariableDeclHelper` and the size-inference plumbing in `VariableCollector` look solid. My concern is concentrated in the `TypeGenerationHelper` refactor, which replaced a single shared type-dispatch with two divergent ones.

### Findings

| Severity | Count |
|----------|-------|
| Critical | 0 |
| Important | 1 |
| Moderate | 1 |
| Nit | 0 |

### Key Issues

**Duplicate/divergent code path in `TypeGenerationHelper.generate()`** (see inline). The previous code routed both bare types and array element types through one shared `dispatchTypeGeneration(ITypeAccessors)` helper. This PR deletes that helper and inlines two separate dispatch blocks. Per this project's strongest rule ("No Duplicate Code Paths — ZERO EXCEPTIONS"), this is the wrong direction — and the two paths have already drifted:
- **`globalType` arrays** (`global.Type[N]`, valid per grammar `arrayType` rule line 548) are no longer handled in the array branch → they fall through to `generateArrayBaseType(null, null, …)` which **throws** `"Array type must have either primitive or user type"`. The old shared path returned the type name. Untested, but a real regression.
- **`qualifiedType` arrays** drop the `isCppScopeSymbol` `::` handling and the `validateCrossScopeVisibility` call that the non-array branch still performs.
- **`cstring` arrays** emit literal `cstring` instead of `char*` (ADR-046), since `generateArrayBaseType` lacks the `cstring` mapping that `generateUserType` has.

These are edge cases (the 966-test suite is green), but they're exactly the latent divergences the project guidelines warn against. Recommend restoring a single shared dispatch that both paths call.

### Minor note
The PR description's "Grammar changes" section (add `qualifiedType`/`scopedType` to `arrayType`) isn't part of this diff — those rules already exist on `main` (`grammar/CNext.g4` and the generated parser are unchanged here). Worth trimming the description so reviewers aren't looking for grammar edits that aren't present.

### What Looks Good
- The `VariableDeclHelper` rewrite produces accurate, helpful error messages that reconstruct both the type-position and trailing dimensions.
- The `hasEmptyArrayTypeDim` guard correctly prevents double-emitting the inferred dimension suffix.
- Test conversions are consistent and the new `c-style-trailing-bracket-rejected` + unit tests lock in the intended rejections.

---

PR #1035, opened 2026-06-21
fix: generate strcpy for string variable initialization (#1030)

### LINKED ISSUE #1030: String variable initialization from another string generates invalid C
## Description

When a string variable is initialized from another string variable, the transpiler generates invalid C code.

## Reproduction

```cnx
string<32> source <- "Test";
string<64> dest <- source;  // Fails to compile in C
```

Generated C:
```c
char source[33] = "Test";
char dest[65] = source;  // Invalid: can't initialize array from another array
```

## Expected Output

Should generate:
```c
char source[33] = "Test";
char dest[65]; strcpy(dest, source);
// OR
char dest[65] = {0}; strncpy(dest, source, 64); dest[64] = '\0';
```

## Affected Tests

- `tests/string/string-assignment-exec.test.cnx`

## Workaround

Use assignment after declaration:
```cnx
string<64> dest <- "";
dest <- source;
```

jlaustill:
## Summary

- Fixes string initialization from another string variable generating invalid C code
- `string<64> dest <- source` now generates valid `strncpy` instead of invalid array initialization
- Adds error for string variable initialization at global scope (strcpy cannot be called outside function body)

Closes #1030

## Test plan

- [x] All 968 integration tests pass
- [x] All StringDeclHelper unit tests pass (35 tests)
- [x] New error test for global scope case
- [x] Existing string tests continue to work


src/transpiler/output/codegen/helpers/StringDeclHelper.ts
```diff
@@ -207,16 +207,68 @@ class StringDeclHelper {
...
+    }
+
+    // Validate capacity: dest >= source
+    if (srcCapacity > capacity) {
+      throw new Error(
+        `Error: Cannot assign string<${srcCapacity}> to string<${capacity}> (potential truncation)`,
+      );
+    }
+
+    // Generate safe string copy code
```
jlaustill, line 261: **[moderate]** quality: The generated output has inconsistent indentation — the declaration line gets indent from the caller, but the `strncpy` line adds ANOTHER level via `${indent}`, resulting in double-indentation:

```c
    char big[65] = "";
        strncpy(big, small, 64); big[64] = '\0';  // <-- double indent
```

This follows the existing pattern in `_generateConcatDecl` and `_generateSubstringDecl`, so it's not a regression — but it may be worth filing a follow-up issue to fix the indentation across all these methods.

Possible fix: Either don't include indent here (let caller handle it), or include indent on ALL lines consistently.

tests/string/string-assignment-exec.test.cnx
```diff
@@ -18,7 +18,7 @@ i32 main() {
 
    
 // Test 7-9: Assignment of empty string
-    string<32> emptySource;
+    string<32> emptySource <- "";
```
jlaustill, line 22: **Nit:** This change is reasonable but slightly alters the test's semantics. The original test was checking initialization from an *uninitialized* string (which was causing E0381). The new test checks initialization from an *empty* string.

Both are valid test cases, but they test different things. Consider adding a comment noting this is intentional, or keeping both scenarios if uninitialized string variable init should also work.

jlaustill (commented):
## Review Summary

**PR:** #1035 — fix: generate strcpy for string variable initialization (#1030)

### Overview

This PR correctly fixes issue #1030 where `string<N> dest <- source` was generating invalid C code (`char dest[65] = source;`). The fix generates proper `strncpy` code with null termination. The implementation follows existing patterns in the codebase for similar operations (concatenation, substring).

### Findings

| Severity | Count |
|----------|-------|
| Critical | 0 |
| Important | 0 |
| Moderate | 1 |
| Nit | 1 |

### Key Issues

1. **[moderate]** The generated C code has inconsistent indentation (see inline comment) — though this follows an existing pattern in the codebase, it may be worth fixing as a follow-up.

2. **[nit]** The test file modification changes `string<32> emptySource;` to `string<32> emptySource <- "";` which is reasonable but slightly changes the test's purpose from "uninitialized string variable init" to "empty string variable init".

### What Looks Good

- ✅ Correct fix matching the expected solution from issue #1030
- ✅ Proper validation: rejects source > dest capacity with clear error message
- ✅ Proper validation: rejects global scope usage with helpful guidance
- ✅ Good unit test coverage (35 tests, new test for global scope error)
- ✅ Integration test properly updated with execution validation

---

src/transpiler/output/codegen/helpers/StringDeclHelper.ts
jlaustill: Acknowledged. This follows the existing pattern in `_generateConcatDecl` and `_generateSubstringDecl`. Filed as a follow-up issue would be appropriate to fix indentation consistency across all three methods.

tests/string/string-assignment-exec.test.cnx
jlaustill: Added clarifying comment noting this tests explicit empty init (uninitialized variables are caught by E0381).

PR #1039, opened 2026-06-21
fix: support scoped type arrays in struct fields

### LINKED ISSUE #1038: Scoped type arrays in struct fields not handled correctly (this.Type[N])
## Description

When a struct field uses a scoped type array declaration like `this.Object[4] objects`, the type is not being resolved correctly in header generation. The generated header shows the literal text `this.Object objects[4]` instead of the resolved C type `DeepThreeLevel_Object objects[4]`.

## Reproduction

```c-next
scope DeepThreeLevel {
    public struct Object {
        i32 id;
    }

    public struct Scene {
        this.Object[4] objects;  // <-- This is not resolved correctly
        u32 count;
    }
}
```

Expected header output:
```c
typedef struct DeepThreeLevel_Scene {
    DeepThreeLevel_Object objects[4];  // Resolved type name
    uint32_t count;
} DeepThreeLevel_Scene;
```

Actual header output:
```c
typedef struct DeepThreeLevel_Scene {
    this.Object objects[4];  // Unresolved - causes C compilation error
    uint32_t count;
} DeepThreeLevel_Scene;
```

## Affected Tests

10 tests fail due to this issue:
- tests/array-initializers/{i8,i16,i32,i64,u64}-array-init.test.cnx
- tests/array-initializers/struct-array.test.cnx
- tests/enum/cross-file-struct-member/single-file-struct-enum-assign.test.cnx
- tests/header-generation/const-struct-array-inferred.test.cnx
- tests/postfix-chains/array-struct-chain.test.cnx
- tests/postfix-chains/deep-three-plus-levels.test.cnx

## Root Cause

The issue is in how struct field types are collected and stored in the symbol table. When the type is `this.Type[N]`, the TType is stored with the unresolved scoped reference rather than the fully qualified C name.

The fix likely needs to happen in the symbol collection phase where struct fields are processed.

## Related

This issue was discovered while implementing #1014-#1017 (rejecting C-style trailing brackets). The main fix for those issues is complete with 954/964 tests passing (98.9%).

jlaustill:
## Summary
- Adds grammar support for `this.Type[N]`, `Scope.Type[N]`, and `global.Type[N]` syntax in struct field declarations
- Fixes type resolution so scoped array types generate correct C names in headers (e.g., `MixedAccess_Sensor` instead of `this.Sensor`)
- Updates code generation to handle the new array type variants

## Test plan
- [x] All 966 integration tests pass
- [x] All 5690 unit tests pass
- [x] Added new test `scoped-type-array.test.cnx` covering `this.Type[N]` and `Scope.Type[N]` syntax
- [x] Verified header output contains resolved type names

Fixes #1038


src/transpiler/output/codegen/helpers/TypeGenerationHelper.ts
```diff
@@ -212,6 +212,30 @@ class TypeGenerationHelper {
       if (arrCtx.stringType()) {
         return "char";
       }
+      // Scoped type array (this.Type[N])
+      if (arrCtx.scopedType()) {
```
jlaustill: **[important]** design: This scoped/qualified/global handling for array element types is a near-verbatim copy of the bare-type dispatch already in this same method (the `ctx.scopedType()` / `ctx.globalType()` / `ctx.qualifiedType()` branches at ~lines 174–199). Both produce the same output domain (C/C++ type strings), so this is a true duplicate code path — the project's CLAUDE.md flags "update in 2 places instead of 1" as the worst anti-pattern, ZERO EXCEPTIONS.

Since `ArrayTypeContext` exposes the same `scopedType()`/`qualifiedType()`/`globalType()`/`stringType()`/`primitiveType()`/`userType()` accessors as `TypeContext`, consider extracting one private dispatcher that takes a structurally-typed `{ scopedType(); qualifiedType(); ... }` and calling it from both the bare-type path and the array-element path.

src/transpiler/logic/symbols/cnext/utils/TypeUtils.ts
```diff
@@ -26,14 +26,35 @@ function resolveStringType(stringCtx: Parser.StringTypeContext): string {
...
+  scopeName?: string,
+): string {
   if (arrayTypeCtx.primitiveType()) {
     return arrayTypeCtx.primitiveType()!.getText();
   }
   if (arrayTypeCtx.userType()) {
     return arrayTypeCtx.userType()!.getText();
   }
+  // Handle scoped type array: this.Type[N]
+  if (arrayTypeCtx.scopedType()) {
```
jlaustill: **[important]** design: `resolveArrayInnerType` now branches over scoped/qualified/global/string/user/primitive — the exact same dispatch (with the exact same `_`-join and `resolveScopedType`/`resolveStringType` calls) that `getTypeName` performs just below. These are the same function in the same output domain (internal `_`-joined symbol names), so the scoped-array path and the non-array path can silently diverge if the naming rules ever change.

Suggest factoring the per-variant extraction into a shared helper that both `getTypeName` and `resolveArrayInnerType` call, so the array element type and the bare type are guaranteed to resolve identically.

tests/header-generation/scoped-type-array.test.cnx
```diff
@@ -0,0 +1,42 @@
...
+    public struct Collection {
+        this.Item[4] items;
+        u32 count;
+    }
+
+}
+
+// Test qualified type array from outside scope
+scope User {
+    public struct Data {
```
jlaustill, line 19: **[moderate]** testing: This file covers `this.Type[N]` (line 13) and `Scope.Type[N]` (here), but the PR also adds a `global.Type[N]` array branch to the grammar, `TypeUtils.resolveArrayInnerType`, and `TypeGenerationHelper.generate` — and the PR description claims to support it. That branch currently has no test. Consider adding a struct field like `global.SomeGlobalType[N]` inside a scope so the `globalType()` array path is exercised in both the symbol and codegen layers (new-code coverage gate is ≥80%).

jlaustill (commented):
## Review Summary

**PR:** #1039 — fix: support scoped type arrays in struct fields

### Overview
Extends the `arrayType` grammar rule to allow `this.Type[N]`, `Scope.Type[N]`, and `global.Type[N]` as array element types in struct fields (Issue #1038), and wires the new variants through both symbol resolution (`TypeUtils`) and code generation (`TypeGenerationHelper`). The change is well-scoped and the leaf helpers (`generateScopedType`/`generateGlobalType`/`generateQualifiedType`, `resolveScopedType`) are correctly reused. No correctness bugs found — the `.IDENTIFIER()` arity matches the grammar (single for `scopedType`/`globalType`, array for `qualifiedType`), and the new `stringType` branch in `resolveArrayInnerType` is behavior-preserving.

### Findings

| Severity | Count |
|----------|-------|
| Critical | 0 |
| Important | 2 |
| Moderate | 1 |
| Nit | 0 |

### Key Issues
- **Duplicate dispatch logic** in both `TypeGenerationHelper.generate()` and `TypeUtils` — the new array-element handling re-implements the same six-variant type dispatch that already exists for the bare-type case in the *same file* and *same output domain*. Per the project's "No Duplicate Code Paths" rule, changing the scoped/qualified/global naming convention would now require edits in two places per file. See inline comments.
- **`global.Type[N]` is untested** — the PR adds the `globalType` array branch to the grammar and both TS files, and the description claims support for it, but the new `scoped-type-array.test.cnx` only exercises `this.Type[N]` and `Scope.Type[N]`. The global branch in both files has no coverage.

### What Looks Good
- Reuses existing leaf helpers rather than re-deriving C/C++ name formatting, so `::` vs `_` and visibility validation stay centralized.
- The `mixed-access-ultimate.expected.error` update is a genuine improvement: the file now parses past `this.Sensor[4]`/`MixedAccess.Controller[2]` and surfaces the real semantic error (self-scope reference) instead of a parse failure.
- Snapshot coverage across `.c`/`.cpp`/`.h`/`.hpp` confirms resolved type names (`Container_Item items[4]`) in headers.

---

src/transpiler/output/codegen/helpers/TypeGenerationHelper.ts
jlaustill: Fixed. Extracted a shared `dispatchTypeGeneration()` helper that handles both bare types and array element types via the `ITypeAccessors` interface. Both paths now use the same dispatch logic.

src/transpiler/logic/symbols/cnext/utils/TypeUtils.ts
jlaustill: Fixed. Extracted a shared `dispatchTypeResolution()` helper using the same pattern. Both `getTypeName()` and array element type resolution now delegate to the same dispatcher.

tests/header-generation/scoped-type-array.test.cnx
jlaustill: I attempted to add `global.Type[N]` test coverage, but discovered a pre-existing bug where global structs used inside scoped struct fields generate headers in wrong declaration order (the scoped struct referencing GlobalItem appears before GlobalItem is defined). Filed as a separate issue to address the header ordering bug. The `global.Type[N]` code path is exercised by the shared dispatch helper refactoring.

PR #1045, opened 2026-06-24
fix: use bounded strncpy for string init from variable (#1044)

### LINKED ISSUE #1044: String initialization from another string variable generates unsafe strcpy (flawfinder CWE-120)
## Summary

When a `string<N>` is **initialized from another string variable**, the transpiler emits an unbounded `strcpy`, which flawfinder flags at level 4 (CWE-120, MS-banned). The **reassignment** path for the same operation correctly emits a bounded `strncpy` — so the two paths diverge, and the init path is unsafe.

This is why the **C Static Analysis (flawfinder)** CI check is currently red (on `main` as well as PRs). The flagged file is `tests/string/string-assignment-exec.test.c`. It is a non-blocking check today, but the generated code is genuinely unsafe.

## Minimal reproduction

```cnx
void copyOnInit() {
    string<32> small <- "Test";
    string<64> big <- small;   // initialization from a string variable
}
```

Generated C:

```c
char small[33] = "Test";
char big[65] = "";
strcpy(big, small);            // <-- unbounded; flawfinder CWE-120
```

flawfinder (the exact CI invocation):

```
$ flawfinder --minlevel=3 --error-level=3 --dataonly --quiet strcpy-init.c
strcpy-init.c:11:  [4] (buffer) strcpy: Does not check for buffer overflows ... (CWE-120).
$ echo $?
1
```

## Contrast: reassignment is already safe

```cnx
string<32> dest <- "";
string<16> src <- "hi";
dest <- src;                   // reassignment
```

Generated C (bounded, not flagged):

```c
strncpy(dest, src, 32); dest[32] = '\0';
```

## Expected

String initialization-from-variable should generate the same bounded copy as reassignment (`strncpy` + explicit null-termination, sized to the destination capacity), eliminating the divergent code path and the CWE-120 finding.

## Notes

- Pre-existing; **not** introduced by #1043 (that PR's changed files are not among the flawfinder-flagged files). Discovered while running `/pr-check` on #1043.
- Fix should unify the init and reassign string-copy code paths (single source of truth) rather than patching only the init path.

jlaustill:
Fixes #1044.

## Problem

Initializing a `string<N>` from another string **variable** emitted an unbounded `strcpy`:

```cnx
string<32> small <- "Test";
string<64> big <- small;
```
```c
char big[65] = "";
strcpy(big, small);          // ❌ CWE-120, flawfinder MS-banned
```

The **reassignment** path already emits a bounded copy, so the two paths diverged. The unsafe `strcpy` also directly contradicts **ADR-008** ("Safe string operations — No `strcpy` — use length-aware operations only") and is the cause of the currently-red **C Static Analysis (flawfinder)** check on `main`.

## Fix

Unify the init path onto the same single source of truth the reassignment path uses, `StringUtils.copyWithNull`:

```c
char big[65] = "";
strncpy(big, small, 64); big[64] = '\0';   // ✅ bounded, null-terminated
```

`N` is the destination capacity; source-capacity-fits validation already runs upstream (`_validateStringInit`), so no valid data is truncated. This removes the divergent code path and the CWE-120 finding.

## Changes

- `StringDeclHelper._generateBoundedStringWithInit`: replace raw `strcpy` with `StringUtils.copyWithNull(name, src, capacity)`.
- Updated the `StringDeclHelper` unit test to assert the bounded copy (was asserting `strcpy`).
- Added an execution regression test under `tests/bugs/issue-1044-string-init-strcpy/`.
- Regenerated `string-assignment-exec` snapshots (`strcpy` → bounded `strncpy`).

## Verification

- `npm run validate:c -- --tool flawfinder` → **All static analysis checks passed** (was failing on this file)
- Unit: 5721 passed · Integration: 972 passed, 0 failed (C + C++)
- `oxlint` / `cspell` clean


tests/bugs/issue-1044-string-init-strcpy/string-init-bounded-copy.test.cnx
```diff
@@ -0,0 +1,12 @@
+// test-execution
+// Issue #1044: initializing a string from another string variable must use a
+// bounded copy (strncpy + null terminator), not an unbounded strcpy (CWE-120,
+// flawfinder MS-banned). This must match the reassignment code path exactly.
+i32 main() {
+    string<32> small <- "Test";
+    string<64> big <- small;          // init from variable
+    if (big != "Test") return 1;
+    if (big.char_count != 4) return 2;
+    if (big.capacity != 64) return 3;
```
jlaustill: **Nit:** The `.capacity` property compiles to a literal on both sides, producing `if (64 != 64) return 3;` which is a no-op. This is a pre-existing codegen quirk (same pattern in `string-assignment-exec`), not introduced here. The test still validates correctly since the condition is always false — just mentioning for awareness.

jlaustill (commented):
## Review Summary

**PR:** #1045 — fix: use bounded strncpy for string init from variable (#1044)

### Overview

This PR correctly fixes a security issue (CWE-120) where string initialization from a variable used unbounded `strcpy` instead of bounded `strncpy`. The fix unifies the init path with the existing reassignment path via `StringUtils.copyWithNull`, eliminating a dual code path.

### Findings

| Severity | Count |
|----------|-------|
| Critical | 0 |
| Important | 0 |
| Moderate | 0 |
| Nit | 1 |

### What Looks Good

- **Single source of truth**: The fix correctly reuses `StringUtils.copyWithNull` rather than duplicating the bounded copy logic, adhering to the project's "no duplicate code paths" principle.
- **Comprehensive testing**: Both unit tests and execution regression tests are included.
- **Clear documentation**: The comment in `StringDeclHelper.ts` explains the issue number, the problem (CWE-120), and why the bounded copy is necessary.
- **Snapshot updates**: All affected test snapshots have been regenerated consistently.

### Notes (not actionable in this PR)

The double-indent on the `strncpy` line (8 spaces instead of 4) and the `if (64 != 64)` pattern for `.capacity` checks are pre-existing behaviors visible in `string-assignment-exec.expected.c` — not introduced by this PR.

### Verdict

This is a clean, well-scoped security fix. The approach is correct and follows project conventions.

---

tests/bugs/issue-1044-string-init-strcpy/string-init-bounded-copy.test.cnx
jlaustill: Fixed — removed the `if (big.capacity != 64)` line. As you noted it compiled to a constant `64 != 64` (no-op), so it validated nothing at runtime. The `big != "Test"` content check already proves the bounded copy is correct. Snapshots regenerated.

PR #1047, opened 2026-06-25
fix: remove execute-only test mode; full pipeline in one pass (#1018)

### LINKED ISSUE #1018: Some test-execution tests fail to compile (stale syntax; possible CI gap)
## Summary

At least two `test-execution` tests (expected to compile, run, and return 0) **fail to compile** — they use syntax the compiler now correctly rejects. They should be updated; that they're still present as `test-execution` suggests the runner/CI isn't compiling every such test (a CI gap).

### 1. `tests/references/array-pass-by-ref.test.cnx:30`

Uses a C-style array parameter `u32 dst[4]` (correctly rejected):

```
error: C-style array parameter is not allowed. Use 'u32[4] dst' instead of 'u32 dst[4]'
```

Fix: `void copyArray(u32[4] src, u32[4] dst, u32 count)`.

### 2. `tests/basics/primitive-types.test.cnx:48`

Verifies zero-init by **reading an uninitialized local**:

```cnx
u32 uninitialized;
if (uninitialized != 0) return 13;   // error[E0381]: use of uninitialized variable 'uninitialized'
```

`E0381` (use-before-explicit-assignment) is intended and fires on every read (if-condition, `return`, initializer). The test's method (read-before-assign) is invalid. Fix: drop that sub-test, or assign explicitly first.

## Notes

- Both compiler behaviors are CORRECT — these are stale tests, not language bugs.
- The concern is process: `test-execution` tests that don't compile imply CI isn't compiling/validating every one of them.
- Found during the agents-guide audit (#1009); more may surface as the audit continues.

jlaustill:
Fixes #1018.

## Problem

A `test-execution` test is only fully validated when it is **transpiled AND compiled AND run**. CI split this across two jobs — an `--transpile-only` job and an `--execute-only` job. The execute-only mode **skips transpilation** and validates the test against the pre-existing committed `.test.c`, so a stale `.cnx` (one the transpiler now rejects) could pass the execute job without anyone re-transpiling it.

**Proven with a probe** — a `test-execution` test whose `.cnx` was made stale (a bare-bool condition, now rejected) but whose committed `.test.c` was still valid:
- `--transpile-only` → **FAILS** (`Transpilation failed: E0701 …`)
- `--execute-only` → **PASSES** (compiles + runs the old `.test.c`, never re-transpiles)

So execute-only is a structural hazard: a run mode that can "validate" a test while skipping the transpile check.

## Fix

Remove the escape hatch so **no run path can skip transpilation**:

- Delete `--execute-only`: the CLI flag + parsing (`scripts/test.ts`), `ITestOptions.executeOnly`, the `executeOnly` branch in `runTestMode` (`scripts/test-utils.ts`), and the `integration:execute` / `integration:bugs:execute` npm scripts.
- **Collapse the CI transpile+execute job pairs into single full-pipeline jobs:** `Integration Tests` (`npm test`) and `Bugs Regression` (`npm run test:bugs`) each transpile + compile + execute in one pass. Updated `needs:`/the final gate and repointed the static-analysis artifact dependency to the merged `integration` job.
- `--transpile-only` is kept as a fast local check.

Also removed the **`test-no-exec`** marker (per maintainer): **0 tests** used it, and it was behaviorally redundant with having no marker (both compile, neither executes).

## Marker inventory (973 `.test.cnx`, for context)

`test-execution` 472 · _no marker_ 240 · `test-error` 219 · `test-c-only` 33 · `test-transpile-only` 24 · `test-cpp-only` 24 · **`test-no-exec` 0** (removed).

## Verification

- A stale `test-execution` `.cnx` is now caught by the only run path (transpile failure); no mode can bypass it.
- Unit: 5722 passed · Integration: 973 passed, 0 failed (C + C++) · `oxlint` / `cspell` clean · workflow YAML validated.


scripts/test-utils.ts
```diff
@@ -741,39 +731,13 @@ class TestUtils {
...
-        if (existsSync(helperImplFile)) {
-          helperImplFiles.push(helperImplFile);
-        }
-      }
-    } else {
-      // Normal mode: Transpile via CLI
+    // Always transpile via CLI: every test is re-transpiled in the same pass
+    // that compiles and executes it, so a stale .cnx can never be silently
+    // validated against pre-existing generated files (Issue #1018).
+    {
```
jlaustill: **[nit]** quality: With the `if (options.executeOnly) {...} else {...}` gone, this is now a naked block whose only effect is scoping `transpileResult` and minimizing the diff. Since nothing after it reuses `transpileResult`, you could unwrap the block and de-indent the body for clarity (the leading comment already explains the intent). Purely cosmetic — fine to leave if you prefer the smaller diff.

.github/workflows/pr-checks.yml
```diff
@@ -453,7 +381,7 @@ jobs:
   c-validation:
     name: C Static Analysis (${{ matrix.tool }})
     runs-on: ${{ (github.event_name == 'push' || !github.event.pull_request.head.repo.fork) && 'self-hosted' || 'ubuntu-latest' }}
-    needs: [integration-transpile]
+    needs: [integration]
```
jlaustill, line 384: **[nit]** design: `c-validation` (and `verify-clean`) now gate on the full `integration` job (transpile + compile + execute) instead of the old fast transpile-only job, so static analysis starts later in the run. It still only needs the `generated-test-files` artifact, which exists as soon as transpilation finishes. This is an acceptable correctness-over-latency tradeoff (and `c-validation` is advisory/`continue-on-error`), so no change needed — just flagging the added critical-path latency in case CI wall-clock matters on the self-hosted runners.

jlaustill (commented):
## Review Summary

**PR:** #1047 — fix: remove execute-only test mode; full pipeline in one pass (#1018)

### Overview
Solid, well-reasoned infrastructure fix. The `--execute-only` mode validated `test-execution` tests against the pre-committed `.test.c` **without re-transpiling**, so a `.cnx` the transpiler now rejects could still pass the execute job. This PR deletes that escape hatch (CLI flag, `ITestOptions.executeOnly`, the `executeOnly` branch, and the two npm scripts) and collapses each CI transpile+execute job pair into a single full-pipeline job. It also drops the unused `test-no-exec` marker (0 tests used it, behaviorally identical to no marker). This eliminates a genuine dual-path hazard — directly in the spirit of the project's "no duplicate code paths" rule, where the transpile and execute checks could silently diverge.

### Findings

| Severity | Count |
|----------|-------|
| Critical | 0 |
| Important | 0 |
| Moderate | 0 |
| Nit | 2 |

### Verification I ran
- **No dangling references:** grepped the whole repo for `executeOnly` / `execute-only` / `integration:execute` / `test-no-exec` / `hasNoExecMarker` / the old job ids. The only remaining `test-no-exec` mentions are in `CHANGELOG.md` and dated `docs/plans/2026-02-24-*.md` — historical records that correctly preserve past state. The active doc (`tests/real-libraries/README.md`) was updated.
- **CI graph is intact:** every `needs:` resolves to an existing job; the gate array was updated to `integration`/`bugs`. The merged `integration` job still uploads `generated-test-files` and `bugs` still uploads `generated-bugs-files`, so the `c-validation` and `verify-clean` artifact consumers are unbroken. Since the artifact glob only captures `*.test.c/.cpp/.h` (transpiler output, identical whether or not execution follows), `verify-clean`'s git-clean check is unaffected by the added compile/execute work.
- **Signature change covered:** `runTestMode` lost a positional `shouldExec` param; its single caller (test-utils.ts:1056) was updated.
- **No orphaned imports:** `existsSync`/`basename`/`dirname`/`join`/`readFileSync` are all still heavily used. `oxlint` reports 0 warnings/0 errors on the three changed `.ts` files.

### What Looks Good
- The fix removes the divergence at the root (one run path) rather than trying to keep two paths in sync — the correct structural resolution.
- The probe described in the PR (stale `.cnx` passes execute-only but fails transpile-only) is a clean demonstration of the actual hazard, not a hypothetical.
- Comments and `ITestOptions` docs were updated to explain *why* the full pipeline is now the only path, which should prevent the escape hatch from being reintroduced.

### Verdict
Looks good — recommend merge. Two nits below, both optional.

---

scripts/test-utils.ts
jlaustill: Done — unwrapped the naked block and de-indented the body (tsc confirms no scope collisions, full suite still green). Behavior unchanged.

.github/workflows/pr-checks.yml
jlaustill: Agreed and intentional — correctness over latency. Static analysis (advisory / continue-on-error) gating on the full `integration` job adds critical-path latency, but the `generated-test-files` artifact is produced as soon as transpilation finishes within that job, so it's a small wall-clock cost on self-hosted runners for the guarantee that nothing runs against un-re-transpiled output. Leaving as-is.

Issue #1048, opened 2026-06-26
Several example .cnx files fail to transpile (stale syntax)


jlaustill:
## Summary

Four files under `examples/` use outdated syntax that the transpiler now correctly rejects, so they fail `bin/cnext.js`. They are not covered by CI (only `tests/` and `bugs/` are), so they have silently rotted. Discovered while fixing #1040 (the all-paths-return diagnostic surfaced `examples/teensy4/blink.cnx`, now fixed; these four remain).

## Failing examples (each is its own minimal reproduction)

| File | Error |
| ---- | ----- |
| `examples/bit_test.cnx:7` | C-style array declaration `u8 buffer[16]` — use `u8[16] buffer` |
| `examples/references.cnx:38` | C-style array parameter `u8 buffer[]` — use `u8[] buffer` |
| `examples/uart_buffer.cnx:110` | C-style array parameter `char msg[]` — use `char[] msg` |
| `examples/nucleo-f446re/blink.cnx:1` | Self-scope reference `SysTick.CTRL` — use `this.CTRL` |

Reproduce:

```
./bin/cnext.js examples/bit_test.cnx -o /tmp/out.c
```

## Expected

All shipped examples should transpile cleanly.

## Suggested fix

Update each example to current syntax. Consider adding the examples to a CI transpile check so they can't rot again (related to the test-coverage gap closed in #1018).

## Notes

- Pre-existing; not introduced by #1040.

Issue #10, opened 2026-01-12
Investigate: Skipped scope tests for global and this contexts


jlaustill:
## Description
Two scope tests are skipped but the reason is unknown. Need investigation to determine root cause.

## Skipped Test Files
- `tests/scope/global-all-types.test.cnx.skip`
- `tests/scope/this-all-types.test.cnx.skip`

## Current Status
❓ **UNKNOWN** - These files are skipped but not documented in any bug reports

## Investigation Needed
1. Read the test files to understand what they're testing
2. Try running them to see what error occurs
3. Determine if this is:
   - A genuine bug that needs fixing
   - A semantic validation issue (expected behavior)
   - A test framework issue
   - An incomplete feature

## Possible Related Context
- Other scope tests ARE passing:
  - ✅ `scope/this-global-test.test.cnx`
  - Working: `global` keyword for scope-qualified access
  - Working: `this` keyword for current scope access

## Priority
MEDIUM - Need to understand why these comprehensive tests are disabled

## Acceptance Criteria
- [ ] Root cause identified and documented
- [ ] Either: Tests pass and .skip removed
- [ ] Or: New issue created for underlying bug
- [ ] Or: Tests confirmed as invalid and deleted

Issue #11, opened 2026-01-12
Test: Enable float modulo error validation test


jlaustill:
## Description
Float modulo error test is skipped but should be enabled as a proper error validation test.

## Skipped Test File
- `tests/floats/float-modulo-error.test.cnx.skip`

## Current Status
✅ **EXPECTED BEHAVIOR** - Test correctly validates that `%` operator is not supported for floating-point types

## Test Content
```cnx
// EXPECT-ERROR: Modulo operator not supported for floating-point types
void test_invalid_modulo() {
    f32 a <- 10.5;
    f32 b <- 3.2;
    
    f32 result <- a % b;  // Should ERROR: % only valid for integers
}
```

## Why Skipped
Test framework may not properly handle expected error tests, or this was temporarily disabled during development.

## Related Tests
Other float error tests that ARE working:
- ✅ `tests/floats/float-division-by-zero.test.cnx` (valid behavior)
- ✅ Other float arithmetic tests pass

Coverage document shows:
- `Float modulo **(ERROR)** | [x] | floats/float-modulo-error.test.cnx`

This suggests it may have been working at some point.

## Task
1. Verify the error is correctly caught by the transpiler
2. Ensure test framework handles expected errors properly
3. Remove .skip extension
4. Verify test passes (by correctly failing compilation)

## Priority
LOW - Semantic validation is working; just need to enable the test

## Acceptance Criteria
- [ ] Test file renamed to remove .skip
- [ ] Test passes by correctly detecting the type error
- [ ] Error message is clear and helpful
- [ ] No regressions in other float tests

PR #1049, opened 2026-06-26
feat: all-paths-return diagnostic E0704 (#1040, ADR-112)

### LINKED ISSUE #1040: Bug: non-void functions with no return statement are accepted (missing all-paths-return diagnostic)
## Summary

C-Next accepts a function declared with a non-void return type that never executes a `return` statement, and silently emits C that falls off the end of the function. A "safer C" should reject this at compile time — reaching the end of a value-returning function and then using the result is undefined behavior in C.

## Minimal reproduction

`missing_return.cnx`:

```c-next
scope Demo {
    u8 getValue() {
        u8 someByte <- 123;
        u8 nextByte <- someByte;
    }
}
```

Transpile:

```
npx tsx src/index.ts missing_return.cnx
```

## Actual behavior

Transpiles successfully (exit 0), generating a non-void function with no `return`:

```c
uint8_t Demo_getValue(void) {
    uint8_t someByte = 123U;
    uint8_t nextByte = someByte;
}
```

No diagnostic is emitted. `cppcheck` catches it downstream:

```
error: Found an exit path from function with non-void return type that has missing return statement [missingReturn]
```

## Expected behavior

C-Next should emit a compile-time error when a non-void function can reach the end of its body without returning a value (all-paths-return analysis).

## Impact / discovery

- Surfaced by `npm run analyze`, which has been red since **2026-01-03** because `examples/teensy4/blink.cnx` (`u8 doSomething()`, line 130) hits this exact case. The committed `examples/teensy4/blink.c` also contains the missing-return function.
- Pre-existing across all `v0.2.x` releases; **not** a regression.
- There is currently no all-paths-return / control-flow-reachability analyzer in `src/`.

## Notes

- This is a **behavior change** (code that currently compiles would become an error), so it requires an **ADR + approval** before implementation, per project policy.
- Likely scoped to a future minor release (e.g. `0.3.0`) rather than a patch.
- `examples/teensy4/blink.cnx` is intentionally left broken for now so this gate stays visibly red until the diagnostic lands; the example should be fixed as part of that work.

jlaustill:
Fixes #1040.

## Summary

A function with a **non-void return type** that can reach the end of its body without returning a value is now a **compile error (E0704)**, instead of silently emitting C that falls off the end (undefined behavior). This is the gap that has held `npm run analyze` red since 2026-01-03.

Designed via brainstorming → **ADR-112** (Accepted), then implemented test-first.

## The rule (strict, conservative, sound)

Per your "as strict as it can be" direction — every path must end in an explicit `return <value>`:

| Construct | guarantees a return? |
|-----------|----------------------|
| `return <expr>;` | yes |
| `return;` (bare, in non-void) | **no** — fails (returns no value) |
| block | iff any statement returns |
| `if`/`else` | iff there's an `else` **and** both branches return |
| `if` without `else` | no |
| `switch` | iff there's a `default` **and** every case + default return |
| `while` / `for` | no (body may not execute) |
| `do-while` | iff its body returns (body always runs once — keeps it sound) |

C-Next has no `break`/`continue` (ADR-026), which keeps the analysis simple. Strictness consequences and the one known limitation (a non-void function that *infinite-loops* is conservatively flagged) are documented in ADR-112.

## Changes

- **`ReturnPathAnalyzer`** (`logic/analysis`, registered as analyzer #10 in `runAnalyzers.ts`) + `IReturnPathError`.
- **ADR-112** + `architecture-decisions.md` index, `error-codes.md` (E0704), `learn-cnext` example.
- **6 `test-error` reproductions** (`tests/bugs/issue-1040-all-paths-return/`) + **16 unit tests** (8 negative, 8 positive incl. else-if chains, switch+default, do-while, critical blocks, void functions).
- Fixed **`examples/teensy4/blink.cnx`** (`u8 doSomething()` now returns) and regenerated `blink.c`/`blink.h`. Note: the committed `blink.c` was *also* stale w.r.t. current codegen (predated `U`-suffix/linkage changes), so the diff is larger than the one added return — it's the accurate current transpiler output, validated by `npm run analyze`.

## Verification

- **`npm run analyze` → exit 0, no `missingReturn`** (green again, first time since January).
- **Blast radius across `tests/` was zero** — no existing function was newly flagged.
- Unit: 5738 passed · Integration: 979 passed, 0 failed (C + C++) · `oxlint` / `cspell` / `prettier` clean.

## Follow-up tracked

While verifying, I found four other `examples/` files that fail to transpile on pre-existing stale syntax (unrelated to #1040) — filed as **#1048**.


src/transpiler/logic/analysis/__tests__/ReturnPathAnalyzer.test.ts
```diff
@@ -0,0 +1,191 @@
...
+        u8 f(bool done) {
+          if (done = true) { return 5; }
+          return;
+        }
+      `);
+      expect(errors).toHaveLength(1);
+    });
+  });
+
+  describe("accepts functions that return on all paths", () => {
```
jlaustill, line 107: **[moderate]** testing: The positive suite is thorough, but it's missing the one shape users will most often write to *satisfy* E0704 — a loop (which never counts as a guaranteed return) followed by a trailing `return`. It passes today because `blockDefinitelyReturns` uses `.some()`, but that's exactly why it's worth pinning down: if the analysis ever moves toward 'the last statement must return' semantics, this would silently regress into a false positive.

```cnx
u8 firstPositive(u8 n) {
  while (n > 0) { return n; }   // loop may not run — not a guaranteed return
  return 0;                     // trailing return guarantees a value
}
// expect 0 errors
```

jlaustill (commented):
## Review Summary

**PR:** #1049 — feat: all-paths-return diagnostic E0704 (#1040, ADR-112)

### Overview
Adds a strict, conservative compile-time diagnostic (E0704) rejecting non-void functions that can fall off the end — closing the long-standing `missingReturn` gap that kept `npm run analyze` red since January. The structural `definitelyReturns` analysis is well-reasoned and, importantly, **sound**: I verified it against `grammar/CNext.g4` and it holds up.

### Findings

| Severity | Count |
|----------|-------|
| Critical | 0 |
| Important | 0 |
| Moderate | 1 |
| Nit | 0 |

### Correctness verification (against the grammar)
- `functionDeclaration : type IDENTIFIER '(' parameterList? ')' block` — `type` and `block` are both mandatory, so `ctx.type()` / `ctx.block()` are never null. No null-deref risk.
- `ifStatement` yields 1 statement (no else) or 2; the `branches.length < 2` guard and `[0]/[1]` accesses are safe, and `else if` chains recurse correctly through the else statement.
- Every statement form that can guarantee a return (`return <expr>`, `if/else`, `switch`+`default`, `do-while`, `block`, `critical`) is handled; `while`/`for`/decls/assignments/expressions correctly fall to `false`.
- The `.some()` semantics in `blockDefinitelyReturns` are sound specifically because C-Next has no `break`/`continue`/`goto` (ADR-026) and loops are assumed to pass control through — so any statement that definitely returns is necessarily reached. The one false-positive case (non-void infinite loop) is correctly documented as a known limitation in ADR-112.

### Cross-reference checks (all pass)
- `docs/error-codes.md`: E0704 row added, E07xx count 3→4 and grand total 32→33 both updated, Related line updated.
- ADR index (`architecture-decisions.md`) updated; `learn-cnext` example added.
- Analyzer renumbered #10, comment validation correctly bumped to #11 with the early-return pattern matching the existing analyzers.
- `IReturnPathError` shape matches what `addError` produces; `.expected.error` fixtures match the emitted message verbatim.

### Key Issues
- **[moderate]** One positive test gap — see the inline comment on the test file: there's no test for the canonical *fix* pattern (a loop followed by a trailing `return`). It passes today via `.some()`, but a test would lock that behavior in.

### What Looks Good
- Genuinely test-first: 6 `test-error` reproductions + 16 unit tests (negative and positive), including else-if chains, switch+default, do-while, and critical blocks.
- The soundness argument is explicit and correct, and the strictness tradeoffs (and the infinite-loop limitation) are documented in the ADR rather than left implicit.
- The `blink.c` regeneration is honestly explained — the committed output was stale w.r.t. the `U`-suffix/linkage codegen, so the diff is the accurate current transpiler output, not hand edits.

---

src/transpiler/logic/analysis/__tests__/ReturnPathAnalyzer.test.ts
jlaustill: Added — `accepts a loop followed by a trailing return (the canonical E0704 fix)`, asserting 0 errors for `while (...) { return n; } return 0;`. You're right that this is the shape users will actually write, and pinning it guards against any future drift toward 'last statement must return' semantics turning it into a false positive. 17 unit tests now.

PR #1051, opened 2026-06-26
fix: modernize rotted example .cnx files + guard them in CI (#1048)

### LINKED ISSUE #1048: Several example .cnx files fail to transpile (stale syntax)

jlaustill:
Fixes #1048.

## Problem

Four files under `examples/` failed to transpile — they aren't part of the test suite, so they silently rotted across several language tightenings (unbounded array params disallowed, `.length` deprecated, `if`-condition rules, a scope/register name collision).

## Fixes (per example)

- **`bit_test.cnx`** — `.length` → explicit `.element_count` / `.bit_length` (ADR-058); `u8 buffer[16]` → `u8[16] buffer`.
- **`references.cnx`** — `zeroArray` now takes a bounded `u8[16]` and reads `.element_count` (unbounded array params are no longer allowed); fixed the C-style array decl/param and dropped the now-unused size const.
- **`uart_buffer.cnx`** — `UART_print` now takes a real **`string<64>`** and iterates it by `char_count` (it was wrongly a `char[]`); extracted a function call out of an `if` condition (E0702) into a variable with an explicit comparison.
- **`nucleo-f446re/blink.cnx`** — renamed `register SysTick` → `Regs` so it no longer collides with the enclosing `scope SysTick`; a bare bitmap bit used as a condition → explicit `= true` (E0701); `while (true)` → `while (1 = 1)`.

## Prevention (the lasting value)

`scripts/__tests__/examples-transpile.test.ts` transpiles **every** `examples/**/*.cnx` and fails if any stops compiling. It runs in `npm run unit` → the **Unit Tests** CI job, so examples can't rot unnoticed again. (Complements the test-integrity work in #1018.)

## Bonus: corrected a stale CLAUDE.md gotcha

While fixing `UART_print` I almost deformed it into a `u8[]` buffer to dodge a documented gotcha claiming string indexing emits `strcmp`. I verified that's **false** — `s[0] != 'H'` generates `s[0] != 'H'` (a char comparison), no `strcmp`. Corrected the note rather than route an example around a non-existent bug.

## Verification

- All 6 examples transpile clean; `npm run analyze` → 0 errors, no security issues.
- Guard: 7/7 green (6 examples + a "found examples" sanity check).
- Unit: 5746 passed · `oxlint` / `cspell` / `prettier` clean.


scripts/__tests__/examples-transpile.test.ts
```diff
@@ -0,0 +1,42 @@
...
+  (absolutePath) => [relative(repoRoot, absolutePath), absolutePath] as const,
+);
+
+describe("examples transpile cleanly (Issue #1048)", () => {
+  it("finds example .cnx files to check", () => {
+    expect(examples.length).toBeGreaterThan(0);
+  });
+
+  it.each(examples)("%s transpiles without errors", async (_label, file) => {
+    const outDir = mkdtempSync(join(tmpdir(), "cnext-examples-"));
```
jlaustill, line 66: **[nit]** quality: Each parameterized case creates a temp dir via `mkdtempSync` that's never removed, so a local `npm run unit` leaves one `cnext-examples-*` directory per example in the OS temp dir on every run. Harmless on ephemeral CI, but it accumulates for anyone running the suite locally. Consider cleaning up — simplest is per-case:

```ts
import { mkdtempSync, rmSync } from "node:fs";
// ...
const outDir = mkdtempSync(join(tmpdir(), "cnext-examples-"));
try {
  const pipeline = new Transpiler({ input: file, outDir, basePath: dirname(file) });
  const result = await pipeline.transpile({ kind: "files" });
  const errorText = result.errors.map((e) => `${e.line}:${e.column} ${e.message}`).join("\n");
  expect(result.success, errorText).toBe(true);
} finally {
  rmSync(outDir, { recursive: true, force: true });
}
```

jlaustill (commented):
## Review Summary

**PR:** #1051 — fix: modernize rotted example .cnx files + guard them in CI (#1048)

### Overview
Repairs four `examples/**` files that silently rotted against several language tightenings, and — the lasting value — adds a unit-suite guard that transpiles every example so they can't rot unnoticed again. The fixes are correct modernizations and the cross-file renames are complete.

### Findings

| Severity | Count |
|----------|-------|
| Critical | 0 |
| Important | 0 |
| Moderate | 0 |
| Nit | 1 |

### Cross-reference verification (independently checked against the head branch)
- **`register SysTick` → `Regs`** in `nucleo-f446re/blink.cnx`: no stray `this.SysTick.*` / old register references remain — all six `init()` accesses and the declaration are updated. The rename also genuinely resolves the scope/register name collision (there's no member left for a stale `this.SysTick` to mis-resolve to, so a miss would fail to compile rather than silently bind wrong).
- **`UART_print` signature** (`const char msg[]` → `const string<64> msg`): no call sites exist in the file, so the change breaks no callers.
- **`BUFFER_SIZE` const removal** in `references.cnx`: no remaining references.
- **`.length` migration**: no `.length` left in any example (all moved to `.element_count` / `.bit_length` / `.char_count`).

### Key Issues
None blocking. One nit on the new guard test (temp-dir cleanup) — see inline comment.

### What Looks Good
- The CI guard is the right fix for the root cause (examples weren't in any suite), and the separate "finds example .cnx files" assertion neatly prevents `it.each([])` from silently passing on an empty glob.
- The `if (UART_available(...))` → extract-to-variable + `= true` change correctly satisfies both E0702 (no call in condition) and E0701 (boolean condition).
- Nice discipline on the CLAUDE.md correction: rather than routing `UART_print` around the documented `str[0]`→`strcmp` gotcha, you verified the gotcha was false and fixed the doc — avoiding a workaround for a non-existent bug.

---

scripts/__tests__/examples-transpile.test.ts
jlaustill: Fixed — wrapped the case body in try/finally with `rmSync(outDir, { recursive: true, force: true })`. Verified: a fresh `npm run unit` now leaves zero `cnext-examples-*` dirs (and cleaned up the ones my dev runs had already left).

examples/nucleo-f446re/blink.cnx
```diff
@@ -235,7 +235,7 @@ scope LED {
         global.STM32F446.GPIOA.BitSetReset.LD2_RESET <- true;
     }
     public void toggle() {
-        if (global.STM32F446.GPIOA.OutputData.LD2) {
+        if (global.STM32F446.GPIOA.OutputData.LD2 = true) {
```
jlaustill, line 238: does this need global. ? I can't imagine this would have naming conflict?

examples/references.cnx
```diff
@@ -34,10 +34,11 @@ void divmod(const i32 dividend, const i32 divisor, i32 quotient, i32 remainder)
...
 }
 
-// Process an array (arrays naturally pass by reference)
-void zeroArray(u8 buffer[], const u32 size) {
+// Process an array (arrays naturally pass by reference). Array parameters must
+// have an explicit size; the element count comes from the array itself.
+void zeroArray(u8[16] buffer) {
     u32 i <- 0;
-    while (i < size) {
+    while (i < buffer.element_count) {
```
jlaustill, line 41: This is beautiful to me

examples/nucleo-f446re/blink.cnx
jlaustill: Great question — it uncovered a real bug, not just style. `global.` is NOT redundant here: removing it changed the codegen (to a different broken form). The actual issue: a bit-range write to a register field through a cross-scope `global.` path wasn't being expanded to mask/shift — it emitted a literal `[10, 2]` subscript, which doesn't compile. Filed as #1052 and fixed in this PR (AssignmentClassifier now recognizes scoped registers in the global path). The nucleo example now generates valid, compiling C. Also strengthened the examples guard to gcc-compile the generated C, not just transpile it — that's what would've caught this.

Issue #1071, opened 2026-06-27
MISRA C:2012 Rule 14.1: A loop counter shall not have essentially floating type (1 violations in generated code)


jlaustill:
## MISRA Rule

**Rule 14.1 (Required)**: A loop counter shall not have essentially floating type

## Violation Count

**1** violations in C-Next-**generated** code (`*.test.c` / generated `*.test.h`). External/third-party headers are excluded.

## How this surfaced

The MISRA portion of `validate:c` was a silent no-op (#1057) — it ran cppcheck's MISRA addon without `--enable=style`, so it detected nothing. Fixing #1057 enabled detection and surfaced this rule, which had **no tracking issue**.

## Example (generated code)

`tests/floats/float-comparison.test.c:68`:

```c
while (temp < 100.0) {
```

## Reproduction

```
cppcheck --addon=misra --enable=style --inline-suppr --quiet \
  --suppress=missingIncludeSystem --suppress=unusedFunction --suppress=floatConversionOverflow \
  -I tests/include tests/floats/float-comparison.test.c
```

## Status

Baselined per-rule in `scripts/misra-baseline.mjs` so it does not fail CI today. Remove `misra-c2012-14.1` from the baseline as part of fixing this issue to enforce it going forward.

## Related

- Surfaced by #1057 (validate:c MISRA no-op)
- Part of #839 (MISRA compliance umbrella)

Issue #1070, opened 2026-06-27
MISRA C:2012 Rule 10.1: Operands shall not be of an inappropriate essential type (1 violations in generated code)


jlaustill:
## MISRA Rule

**Rule 10.1 (Required)**: Operands shall not be of an inappropriate essential type

## Violation Count

**1** violations in C-Next-**generated** code (`*.test.c` / generated `*.test.h`). External/third-party headers are excluded.

## How this surfaced

The MISRA portion of `validate:c` was a silent no-op (#1057) — it ran cppcheck's MISRA addon without `--enable=style`, so it detected nothing. Fixing #1057 enabled detection and surfaced this rule, which had **no tracking issue**.

## Example (generated code)

`tests/bit-indexing/bit-range-width-u32.test.c:107`:

```c
value = (value & ~(0xFFFFFFFFU << 0)) | ((4294967295 & 0xFFFFFFFFU) << 0);
```

## Reproduction

```
cppcheck --addon=misra --enable=style --inline-suppr --quiet \
  --suppress=missingIncludeSystem --suppress=unusedFunction --suppress=floatConversionOverflow \
  -I tests/include tests/bit-indexing/bit-range-width-u32.test.c
```

## Status

Baselined per-rule in `scripts/misra-baseline.mjs` so it does not fail CI today. Remove `misra-c2012-10.1` from the baseline as part of fixing this issue to enforce it going forward.

## Related

- Surfaced by #1057 (validate:c MISRA no-op)
- Part of #839 (MISRA compliance umbrella)

Issue #1065, opened 2026-06-27
MISRA C:2012 Rule 12.2: The right hand operand of a shift operator shall lie in the range zero to one less than the width in bits of the essential type of the left hand operand (48 violations in generated code)


jlaustill:
## MISRA Rule

**Rule 12.2 (Required)**: The right hand operand of a shift operator shall lie in the range zero to one less than the width in bits of the essential type of the left hand operand

## Violation Count

**48** violations in C-Next-**generated** code (`*.test.c` / generated `*.test.h`). External/third-party headers are excluded.

## How this surfaced

The MISRA portion of `validate:c` was a silent no-op (#1057) — it ran cppcheck's MISRA addon without `--enable=style`, so it detected nothing. Fixing #1057 enabled detection and surfaced this rule, which had **no tracking issue**.

## Example (generated code)

`tests/slice-assignment/bit-range-unchanged.test.c:21`:

```c
flags = (flags & ~(1U << 15)) | (1U << 15);
```

## Reproduction

```
cppcheck --addon=misra --enable=style --inline-suppr --quiet \
  --suppress=missingIncludeSystem --suppress=unusedFunction --suppress=floatConversionOverflow \
  -I tests/include tests/slice-assignment/bit-range-unchanged.test.c
```

## Status

Baselined per-rule in `scripts/misra-baseline.mjs` so it does not fail CI today. Remove `misra-c2012-12.2` from the baseline as part of fixing this issue to enforce it going forward.

## Related

- Surfaced by #1057 (validate:c MISRA no-op)
- Part of #839 (MISRA compliance umbrella)

Issue #1064, opened 2026-06-27
MISRA C:2012 Rule 8.5: An external object or function shall be declared once in one and only one file (53 violations in generated code)


jlaustill:
## MISRA Rule

**Rule 8.5 (Required)**: An external object or function shall be declared once in one and only one file

## Violation Count

**53** violations in C-Next-**generated** code (`*.test.c` / generated `*.test.h`). External/third-party headers are excluded.

## How this surfaced

The MISRA portion of `validate:c` was a silent no-op (#1057) — it ran cppcheck's MISRA addon without `--enable=style`, so it detected nothing. Fixing #1057 enabled detection and surfaced this rule, which had **no tracking issue**.

## Example (generated code)

`tests/header-generation/const-array-dimensions.test.h:26`:

```c
extern TReading readings[4][4];
```

## Reproduction

```
cppcheck --addon=misra --enable=style --inline-suppr --quiet \
  --suppress=missingIncludeSystem --suppress=unusedFunction --suppress=floatConversionOverflow \
  -I tests/include tests/header-generation/const-array-dimensions.test.h
```

## Status

Baselined per-rule in `scripts/misra-baseline.mjs` so it does not fail CI today. Remove `misra-c2012-8.5` from the baseline as part of fixing this issue to enforce it going forward.

## Related

- Surfaced by #1057 (validate:c MISRA no-op)
- Part of #839 (MISRA compliance umbrella)

PR #1073, opened 2026-06-27
fix: make validate:c MISRA check actually detect violations (#1057)

### LINKED ISSUE #1057: validate:c MISRA check is a silent no-op (runMisra() missing --enable=style)
## Summary

`npm run validate:c` runs cppcheck's MISRA addon **without `--enable=style`**, so it reports **nothing**. cppcheck emits all MISRA findings at `style` severity, which is suppressed unless style checks are enabled. The result: the MISRA portion of our C static analysis is a **silent no-op** — it passes with exit 0 even on files riddled with violations.

This means none of the open MISRA-violation issues (#841–#869, and the #839 umbrella) are actually being detected by CI/local validation. The check has never failed because it never checks anything.

## Location

`scripts/batch-validate.mjs`, `runMisra()` (~line 322):

```js
execFileSync("cppcheck", [
  "--addon=misra",
  "--inline-suppr",
  "--error-exitcode=1",
  "--suppress=missingIncludeSystem",
  "--suppress=unusedFunction",
  "--suppress=floatConversionOverflow",
  "--quiet",
  "-I", INCLUDE_DIR,
  "-I", dirname(file),
  file,
], ...)
```

Note: `runCppcheck()` (the non-MISRA path) **does** pass `--enable=warning,performance`, but `runMisra()` passes no `--enable` flag at all.

## Minimal Reproduction

`probe.c` — contains real Rule 14.4, 10.4, and 8.4 violations:

```c
#include <stdint.h>
void f(uint8_t x, uint8_t *p) {
    if (x) { x = 0U; }       /* MISRA 14.4: non-Boolean controlling expr */
    if (p) { p[0] = 0U; }    /* MISRA 14.4 */
}
```

Run with the **exact flags** `runMisra()` uses:

```
$ cppcheck --addon=misra --inline-suppr --error-exitcode=1 \
    --suppress=missingIncludeSystem --suppress=unusedFunction \
    --suppress=floatConversionOverflow --quiet probe.c
$ echo $?
0          # <-- passes, no output, despite real violations
```

Add `--enable=style` and the violations appear:

```
$ cppcheck --addon=misra --enable=style --quiet probe.c
probe.c:3:8: style: misra violation ... [misra-c2012-14.4]
probe.c:4:8: style: misra violation ... [misra-c2012-14.4]
```

Verified on cppcheck 2.13.0.

## Impact

- Running the full generated test suite (694 `.c` files) with `--enable=style` surfaces thousands of real findings (e.g. 10.4: 3467, 15.5: 2525, 8.4: 935) — all currently invisible to `validate:c`.
- The `c-validation` CI job (advisory, `continue-on-error: true` per #839) passes trivially for MISRA.

## Suggested Fix (needs its own plan — do NOT just flip the flag)

Adding `--enable=style` to `runMisra()` will immediately surface thousands of pre-existing violations and, with `--error-exitcode=1`, fail. A safe rollout needs:

1. Enable `--enable=style` (or `--enable=all`) in `runMisra()`.
2. Baseline/suppress the rules already tracked by open issues (#841–#869) so only *new* violations fail.
3. Keep the job advisory initially, then promote per-rule as each tracked issue is fixed.

## Discovered while

Investigating #848 (Rule 14.4). That rule turned out to have **0** current violations (fixed by #884 + #1042), which is how this no-op surfaced — the scan that "passed" was passing for the wrong reason.

jlaustill:
## Summary

The MISRA portion of `npm run validate:c` was a **silent no-op**. `runMisra()` invoked cppcheck's MISRA addon without `--enable=style`; cppcheck emits every MISRA finding at `style` severity, so the addon reported nothing and exited 0. The check "passed" by never checking anything — none of the tracked MISRA issues (#841–#869) were actually being detected.

Verified with cppcheck 2.13.0: the exact flags `runMisra()` used exit 0 with no output on a file with real Rule 14.4/8.4 violations; adding `--enable=style` surfaces them.

## What changed

- **New shared module `scripts/misra-baseline.mjs`** (unit-tested) owns the cppcheck invocation and the failure decision — single source of truth for the runner and tests, removing the duplicate inline arg list.
  - `buildArgs()` always passes `--enable=style` (the fix).
  - `parseViolations()` extracts MISRA findings, ignoring non-MISRA style noise.
  - `findFailures()` gates only on **C-Next-generated** output (`*.test.c` / generated `*.test.h`) and **un-baselined** rules.
- **Generated-only scope**: cppcheck also flags transitively-included third-party libs (cJSON, FreeRTOS) and hand-written fixtures (`widget.h`, c-interop headers). Those are not C-Next's code and must never gate the build.
- **Per-rule baseline**: every rule with current generated violations is allow-listed and mapped to its tracking issue. The check is **green today** but **fails on any new rule class** (or any baselined rule once its entry is removed as the issue is fixed).
- Real cppcheck execution failures (exit status ≠ 1) are now surfaced instead of swallowed.

## Bug tracking

Enabling detection surfaced **14 previously-untracked** MISRA rules with violations in generated code. Per the project's zero-exceptions bug-tracking rule, filed one issue per rule and added each to the baseline:

| Rule | Issue | Rule | Issue |
|------|-------|------|-------|
| 8.6 | #1059 | 19.2 | #1067 |
| 5.9 | #1060 | 5.7 | #1068 |
| 5.6 | #1061 | 21.7 | #1069 |
| 7.2 | #1062 | 10.1 | #1070 |
| 5.8 | #1063 | 14.1 | #1071 |
| 8.5 | #1064 | 9.2 | #1072 |
| 12.2 | #1065 | | |
| 20.1 | #1066 | | |

The 16 rules already tracked by #841–#869 are referenced in the baseline. Rules whose issues are effectively resolved (10.3 → #845, 2.4 → #868) have **0** current violations and stay **out** of the baseline, so they remain enforced.

## Verification

- New unit tests `scripts/__tests__/misra-baseline.test.ts` (9 tests, written test-first).
- Full unit suite: 5764 passing.
- `npm run validate:c -- --tool misra` exits 0 with all firing rules baselined (688 files).
- End-to-end: injecting a temporary un-baselined Rule 14.4 violation makes the check fail with exit 1 — proving it is no longer a no-op.
- prettier / oxlint / cspell / jscpd (0.82% < 3%) all clean.

Closes #1057.

### DIFF
--- a/docs/misra-compliance.md
+++ b/docs/misra-compliance.md
@@ -26,6 +26,26 @@ This document tracks C-Next's compliance with MISRA C:2012 guidelines. MISRA C i
 
 ---
 
+## How violations are validated
+
+`npm run validate:c` runs cppcheck's MISRA addon over every generated C test
+file via `scripts/batch-validate.mjs`. cppcheck emits all MISRA findings at
+`style` severity, so the runner passes `--enable=style` — omitting it makes the
+check a silent no-op (issue #1057).
+
+The failure decision lives in `scripts/misra-baseline.mjs`:
+
+- **Generated-only**: cppcheck also flags transitively-included third-party libs
+  and hand-written fixtures; only violations in C-Next-generated output
+  (`*.test.c` / generated `*.test.h`) count.
+- **Per-rule baseline**: every rule that currently has generated violations is
+  listed in `BASELINE`, mapped to its tracking issue (#841–#869, #1059–#1072).
+  Baselined rules do not fail the build, so the check is green today but fails
+  on any **new** rule class. When a rule's issue is fixed, remove its entry from
+  `BASELINE` to start enforcing it.
+
+---
+
 ## Directives
 
 ### Dir 1 - Implementation Compliance
--- a/docs/misra-compliance.md
+++ b/docs/misra-compliance.md
@@ -39,8 +39,8 @@ The failure decision lives in `scripts/misra-baseline.mjs`:
   and hand-written fixtures; only violations in C-Next-generated output
   (`*.test.c` / generated `*.test.h`) count.
 - **Per-rule baseline**: every rule that currently has generated violations is
-  listed in `BASELINE`, mapped to its tracking issue (#841–#869, #1059–#1072).
-  Baselined rules do not fail the build, so the check is green today but fails
+  listed in `BASELINE`, each mapped to its tracking issue. Baselined rules do
+  not fail the build, so the check is green today but fails
   on any **new** rule class. When a rule's issue is fixed, remove its entry from
   `BASELINE` to start enforcing it.
 
--- a/scripts/__tests__/misra-baseline.test.ts
+++ b/scripts/__tests__/misra-baseline.test.ts
@@ -44,7 +44,7 @@ describe("parseViolations", () => {
 
   it("returns an empty array when there are no MISRA findings", () => {
     const noise =
-      "probe.c:1:1: style: Variable unused. [unreadVariable]\nn" +
+      "probe.c:1:1: style: Variable unused. [unreadVariable]\n" +
       "probe.c:2:2: warning: something. [someWarning]";
     expect(MisraBaseline.parseViolations(noise)).toEqual([]);
   });
--- a/scripts/__tests__/misra-baseline.test.ts
+++ b/scripts/__tests__/misra-baseline.test.ts
@@ -1,0 +1,91 @@
+/**
+ * Unit tests for misra-baseline.mjs
+ *
+ * Regression guard for #1057: the MISRA portion of `validate:c` was a silent
+ * no-op because runMisra() invoked cppcheck without `--enable=style`, and
+ * cppcheck emits all MISRA findings at `style` severity. These tests lock in:
+ *   1. the cppcheck argv always enables style (so MISRA findings appear),
+ *   2. MISRA violations are parsed from output (ignoring non-MISRA noise),
+ *   3. only violations of un-baselined rules are treated as failures.
+ */
+
+import MisraBaseline from "../misra-baseline.mjs";
+
+// Real cppcheck output captured with `--enable=style`. Mixes MISRA findings
+// with non-MISRA style findings to prove the parser ignores the latter.
+const SAMPLE_OUTPUT = `probe.c:3:14: style: Assignment of function parameter has no effect outside the function. [uselessAssignmentArg]
+probe.c:3:16: style: Variable 'x' is assigned a value that is never used. [unreadVariable]
+probe.c:2:6: style: misra violation (use --rule-texts=<file> to get proper output) [misra-c2012-8.4]
+probe.c:3:8: style: misra violation (use --rule-texts=<file> to get proper output) [misra-c2012-14.4]
+probe.c:4:8: style: misra violation (use --rule-texts=<file> to get proper output) [misra-c2012-14.4]`;
+
+describe("buildArgs", () => {
+  it("enables style so cppcheck emits MISRA findings (the #1057 bug)", () => {
+    const args = MisraBaseline.buildArgs("foo.test.c", "tests/include");
+    expect(args).toContain("--enable=style");
+  });
+
+  it("loads the MISRA addon and targets the given file", () => {
+    const args = MisraBaseline.buildArgs("foo.test.c", "tests/include");
+    expect(args).toContain("--addon=misra");
+    expect(args).toContain("foo.test.c");
+  });
+});
+
+describe("parseViolations", () => {
+  it("extracts MISRA rule IDs and ignores non-MISRA style findings", () => {
+    const violations = MisraBaseline.parseViolations(SAMPLE_OUTPUT);
+    expect(violations.map((v) => v.ruleId)).toEqual([
+      "misra-c2012-8.4",
+      "misra-c2012-14.4",
+      "misra-c2012-14.4",
+    ]);
+  });
+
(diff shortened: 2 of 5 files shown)


scripts/misra-baseline.mjs
```diff
@@ -0,0 +1,125 @@
...
+  ["misra-c2012-5.7", "#1068"],
+  ["misra-c2012-21.7", "#1069"],
+  ["misra-c2012-10.1", "#1070"],
+  ["misra-c2012-14.1", "#1071"],
+  ["misra-c2012-9.2", "#1072"],
+]);
+
+// Matches a cppcheck finding line and captures the file path and MISRA rule id:
+//   path/to/file.test.c:12:8: style: misra violation ... [misra-c2012-14.4]
+const MISRA_LINE = /^(.+?):\d+:\d+:.*\[(misra-c2012-\d+\.\d+)\]/;
```
jlaustill, line 74: **[moderate]** quality: The regex uses a non-greedy match `(.+?)` for the file path, which works but could be more precise.

Current pattern matches until the first `:`, which is correct for Unix paths but could theoretically match incorrectly on Windows paths with drive letters (e.g., `C:\path\file.c:12:8:...`). Since this is an embedded/Unix-focused project, this is likely fine, but worth noting if Windows CI ever becomes a consideration.

No action needed — just documenting the assumption.

scripts/__tests__/misra-baseline.test.ts
```diff
@@ -0,0 +1,91 @@
...
+    expect(violations.map((v) => v.ruleId)).toEqual([
+      "misra-c2012-8.4",
+      "misra-c2012-14.4",
+      "misra-c2012-14.4",
+    ]);
+  });
+
+  it("returns an empty array when there are no MISRA findings", () => {
+    const noise =
+      "probe.c:1:1: style: Variable unused. [unreadVariable]\nn" +
```
jlaustill: Nit: There's a stray `n` in this string literal (`\nn` instead of `\n`). It doesn't affect test correctness since the parser still won't find MISRA violations in either line, but it's technically not the intended output format.

```suggestion
      "probe.c:1:1: style: Variable unused. [unreadVariable]\n" +
```

docs/misra-compliance.md
```diff
@@ -26,6 +26,26 @@ This document tracks C-Next's compliance with MISRA C:2012 guidelines. MISRA C i
...
+`style` severity, so the runner passes `--enable=style` — omitting it makes the
+check a silent no-op (issue #1057).
+
+The failure decision lives in `scripts/misra-baseline.mjs`:
+
+- **Generated-only**: cppcheck also flags transitively-included third-party libs
+  and hand-written fixtures; only violations in C-Next-generated output
+  (`*.test.c` / generated `*.test.h`) count.
+- **Per-rule baseline**: every rule that currently has generated violations is
+  listed in `BASELINE`, mapped to its tracking issue (#841–#869, #1059–#1072).
```
jlaustill: Nit: Minor inconsistency — the issue ranges here say `#841–#869, #1059–#1072` but the baseline in the code has 16 pre-existing + 14 new = 30 rules, and some pre-existing issues (#841–#869) actually span more rules than the range suggests (e.g., there are gaps). This is fine for human readers, but if someone tries to audit by counting, the ranges are approximate rather than exhaustive. Consider just saying "mapped to their tracking issues" without the specific range, or keep it as-is for quick reference.

jlaustill (commented):
## Review Summary

**PR:** #1073 — fix: make validate:c MISRA check actually detect violations (#1057)

### Overview

Excellent bug fix that addresses a critical silent failure in the MISRA validation pipeline. The MISRA check was passing by never actually checking anything — a textbook example of a "false green" that erodes CI trustworthiness. The fix correctly adds `--enable=style` and introduces a well-tested baseline module.

### Findings

| Severity | Count |
|----------|-------|
| Critical | 0 |
| Important | 0 |
| Moderate | 1 |
| Nit | 2 |

### Key Issues

No critical or important issues found. The implementation is solid:
- Single source of truth for cppcheck invocation (via `MisraBaseline`)
- Generated-only scope properly excludes third-party code
- Per-rule baseline with issue tracking follows project conventions
- Comprehensive unit tests (9 tests) with good coverage of the failure modes
- Real cppcheck execution failures are now surfaced instead of swallowed

### What Looks Good

1. **Test-driven approach**: Tests lock in the specific behaviors that caused #1057 — `--enable=style`, MISRA filtering, and baseline gating
2. **Documentation updated**: `docs/misra-compliance.md` now explains how the validation works, which is exactly where someone would look
3. **Issue hygiene**: All 14 newly-surfaced rules have tracking issues (#1059–#1072), following the project's zero-exceptions bug-tracking rule
4. **Clean separation**: The `MisraBaseline` class is reusable and testable, eliminating the inline arg duplication noted in the PR description

---

scripts/misra-baseline.mjs
jlaustill: Documented the assumption in a comment: paths here are POSIX (the runner globs `tests/`), so the non-greedy capture is unambiguous; noted that a Windows drive-letter path would need a more specific anchor.

scripts/__tests__/misra-baseline.test.ts
jlaustill: Fixed — removed the stray `n` (`\nn` → `\n`).

docs/misra-compliance.md
jlaustill: Reworded to "each mapped to its tracking issue" — dropped the approximate ranges so it doesn't mislead anyone auditing by count.

Issue #1081, opened 2026-06-28
MISRA C:2012 Rule 21.15: slice-assignment memcpy passes incompatible pointer types (30 violations, unmasked by #847)


jlaustill:
## MISRA Rule

**Rule 21.15 (Required)**: The pointers passed as arguments to `memcpy`, `memmove` and `memcmp` shall be pointers to qualified or unqualified versions of compatible types.

## Violation Count

30 violations in generated code, across 8 test files (all slice-assignment / scope string-param-slice).

## How this surfaced

Unmasked by **#847**. cppcheck's MISRA addon reports only **one** rule per source line. Slice assignment lowers `buffer[offset, length] <- value` to:

```c
memcpy(&buffer[0], &value32, 4);
```

which previously tripped **Rule 17.7** (discarded non-void return). #847 added a `(void)` cast, resolving 17.7 — and the *next* applicable rule on that line, **21.15**, now surfaces. The 21.15 violation pre-existed; it was simply hidden behind 17.7.

## Why it violates 21.15

The two pointer arguments point to incompatible types:
- `&buffer[0]` → `uint8_t *`
- `&value32`   → `uint32_t *`

Slice assignment deliberately reinterprets a wider integer as a byte sequence, so the destination (`uint8_t`) and source (`uint32_t`/`uint64_t`/…) are never the same type.

## Minimal reproduction

`tests/slice-assignment/basic-slice-copy.test.cnx`:

```c-next
u32 main() {
    u8[64] buffer;
    u32 value32 <- 0x12345678;
    buffer[0, 4] <- value32;   // -> (void)memcpy(&buffer[0], &value32, 4);
    return 0;
}
```

```
cppcheck --addon=misra --enable=style --inline-suppr --quiet \
  --suppress=missingIncludeSystem --suppress=unusedFunction \
  -I tests/include tests/slice-assignment/basic-slice-copy.test.c
# -> misra violation [misra-c2012-21.15] on each (void)memcpy(...) line
```

## Status

Baselined per-rule in `scripts/misra-baseline.mjs` (`misra-c2012-21.15` → this issue) so it does not fail CI today. Remove that baseline entry as part of fixing this issue to enforce it going forward.

## Possible fix directions (for research)

- Copy through a `uint8_t *` view of the source (e.g. cast `&value32` to `const uint8_t *`, or memcpy from a `uint8_t` temp), so both arguments are `uint8_t *` / compatible.
[...]

Issue #1080, opened 2026-06-28
ADR-115 (research): C-Next should enforce that every non-void return value is used or explicitly discarded


jlaustill:
## Summary

Research and write an ADR (next slot: **ADR-115**) for a C-Next language-level guarantee: **the return value of every non-void function must be used, or explicitly discarded with dedicated syntax.** Discarding a non-void return implicitly would become a compile error (`E07xx`).

This is the ambitious counterpart to **#847**. #847 makes the *generated C* MISRA-compliant by auto-emitting `(void)` casts on discarded non-void calls. That silences MISRA C:2012 Rule 17.7 in the output, but it does nothing for the C-Next author — the language still lets you silently drop a meaningful return value (an error code, a status, a parsed result). This issue proposes pushing the guarantee *up into the language* so the mistake is caught at the C-Next source level, the way C-Next already forbids other foot-guns.

## Motivation

Discarding a return value is a classic source of embedded bugs: ignoring an error/status code, dropping a `safe_div` result, throwing away a parse outcome. C-Next's whole premise is "a safer C" with explicit data flow (`<-` for assignment, no implicit conversions, no fallthrough). "You must acknowledge what a function hands back" fits that philosophy directly.

Concretely, today this transpiles without complaint:

```c-next
u32 next() { value <- value + 1; return value; }

i32 main() {
    next();          // return value silently dropped
    return 0;
}
```

With #847 the generated C becomes `(void)Counter_next();` — MISRA-clean, but the C-Next author was never told they dropped a value they may have wanted.

## Proposed direction (to be decided in the ADR)

Make an implicitly-discarded non-void return a compile error, and provide explicit, ergonomic opt-out syntax for the "I really do mean to ignore this" case. Exact syntax/semantics are **open questions for the ADR**, but the design space includes:

- **Always-error + explicit discard syntax** (recommended starting point): discarding is an error unless written explicitly, e.g. `_ <- next();` or `discard next();`. Mirrors Rust's `let _ =` / `#[must_use]`-everywhere and Go vet's `errcheck` culture.
- **Opt-in `must_use` annotation**: only functions marked must-use error when discarded; everything else auto-`(void)`. Less safe by default, less disruptive.
- **Opt-out `discardable` annotation**: all non-void error on discard unless the function is marked discardable. Safe by default; library authors mark printf-likes.

## Open questions for the research/ADR

1. **Default stance:** safe-by-default (everything must be used) vs annotation-gated? C-Next leans safe-by-default.
[...]

PR #1082, opened 2026-06-28
fix: MISRA C:2012 Rule 17.7 — cast discarded non-void returns (#847)

### LINKED ISSUE #847: MISRA C:2012 Rule 17.7: Return value of non-void function shall be used (30 violations)
## MISRA Rule

**Rule 17.7 (Mandatory)**: The value returned by a function having non-void return type shall be used.

## Violation Count

30 violations across test files.

## Description

Return values from non-void functions must be used or explicitly cast to void.

## Example Violation

```c
printf("hello");  // MISRA 17.7 violation - return value ignored
(void)printf("hello");  // Compliant
```

## Fix Strategy

1. Update CodeGenerator to cast unused return values to `(void)`
2. Review function call patterns in generated code

## Related

- Part of #839 breakdown
- MISRA C:2012 Mandatory rule

jlaustill:
## Summary

Fixes **#847**. Generated C discarded the return values of non-void functions called as statements, violating MISRA C:2012 Rule 17.7. Every discarded non-void call now gets an explicit `(void)` cast.

The issue's stated "30 violations" predated #1057 (which enabled full MISRA detection) — the real surface was ~100+ across four emission paths, all now covered.

## What changed

**Statement-level calls** (`printf`, scope getters, `safe_div`/`safe_mod` builtins) are cast once at the `CodeGenerator.generateStatement` expression-statement chokepoint:
- `CallExprGenerator` records the resolved callee on `CodeGenState.lastCallTarget`.
- New `ReturnValueCast` helper decides (a) whether the statement is a single bare call and (b) whether the callee returns non-void.
- **Only known non-void callees are cast.** Void C-Next functions and void stdlib/Arduino calls (`digitalWrite`, `delay`, `free`, …) are deliberately left alone — no churn, no redundant casts.

**String ops** (`strncpy`/`strncat`): the call text + cast now come from a single `StringUtils.boundedCopy`/`boundedConcat` primitive. `StringDeclHelper` had duplicate `strncpy`/`strncat` literals — those were unified onto the same primitive (removing a duplicate code path).

**Slice assignment** (`memcpy`): cast in `ArrayHandlers`.

**Shared stdlib metadata**: header + void-ness moved out of `FunctionCallAnalyzer`'s private map into a new `StdlibFunctions` constants module, so codegen and analysis share one list instead of a parallel copy.

**Enforcement**: `misra-c2012-17.7` removed from the MISRA baseline; baseline integrity test guards it stays out.

## ⚠️ Unmasked a pre-existing bug → #1081

cppcheck reports **one rule per line**. Resolving 17.7 on slice `memcpy` lines unmasked a pre-existing **Rule 21.15** (incompatible pointer types: `uint8_t*` vs `uintN_t*`). It pre-existed but was hidden behind 17.7. Filed as **#1081** and baselined (same pattern as the other tracked rules).

## Testing

- New focused test: `tests/misra/rule-17-7-void-cast/` (scope getter, void scope call, `printf`, string assignment) — written test-first (RED→GREEN).
- `npm run unit` — 5794 pass (incl. updated StringUtils/StringDeclHelper/ArrayHandlers expectations + new baseline guard).
- `npm run test:q` — 991/991 integration pass (C + C++).
- `npm run validate:c` — all MISRA checks pass; **0** `misra-c2012-17.7` in generated code.
- prettier / oxlint / cspell / knip / madge / depcruise — clean.

## Related

- Closes #847
- #1081 — Rule 21.15 (unmasked here)
- #1080 — follow-up: ADR-115 research to enforce return-value use at the C-Next *source* level
- #1057 — validate:c MISRA detection (surfaced the full count)


jlaustill:
Closing per design review. The PR conflated two cases that must be handled differently:

1. **Compiler-internal lowering** (`string`/slice features emit `strncpy`/`strncat`/`memcpy` the dev never wrote) → auto-`(void)` is correct.
2. **User-written discarded non-void calls** (`printf`, scope getters, `safe_div`) → should be a **compile error** requiring explicit discard, NOT a silent auto-`(void)` that swallows a developer decision.

Case 2 is a language-behavior change that belongs in the ADR work (#1080), and it makes sense to design both cases together there rather than ship case 1 alone. Folding all of it into #1080. Branch `fix/847-misra-17.7-void-cast` is left intact as a starting implementation. #847 stays open (blocked on #1080).

Issue #1074, opened 2026-06-27
ADR-113: Implement `forever` infinite-loop statement (core — blocks #849)


jlaustill:
## Summary

Implement the `forever` infinite-loop statement per **ADR-113 (Accepted, 2026-06-27)**. This is the **core** of ADR-113 and is the **prerequisite for #849** (ADR-114 dead-code / reachability) — it introduces the shared *divergence primitive* that #849's reachability pass consumes, and it removes the one known dead-code offender in the repo (`blink.cnx`'s trailing `return 0;`) before #849's breaking rule turns on.

ADR: `docs/decisions/adr-113-forever-loops.md`

## Current behavior

C-Next has no dedicated infinite-loop construct. The only ways to write "loop forever" are hacks:

```cnx
while (1 = 1) {   // 1 = 1 is equality (== ) in C-Next
    loop();
}
```

Because `ReturnPathAnalyzer` (ADR-112) treats a `while` body as *possibly skipped*, a non-void function whose only terminal path is such a loop fails **E0704** unless a dead trailing `return` is added. The flagship example ships exactly this dead code (`examples/nucleo-f446re/blink.cnx`):

```cnx
i32 main() {
    setup();
    while (1 = 1) { loop(); }
    return 0;          // unreachable — present only to satisfy E0704
}
```

## Proposed behavior

```cnx
void main() {
    setup();
    forever {
        loop();
    }
}                      // no return type to honor, no dead code
```

## Scope (this issue — ADR-113 core)

- [ ] **Grammar**: add `FOREVER : 'forever';` token and `foreverStatement : FOREVER block ;` (braces always required — no single-statement form), added to the `statement` alternatives. Run `npm run antlr`, update `Parser.*Context` references, remove dead methods.
- [ ] **Codegen**: lower `forever { … }` → `for (;;) { … }`, emitting a comment noting the MISRA C:2012 Rule 14.3 infinite-loop carve-out.
- [ ] **Divergence primitive**: introduce the "divergent statement" concept into `ReturnPathAnalyzer` — a `forever` loop is a terminal path, like an unconditional `return`. A function whose every path either returns a value or diverges via `forever` satisfies E0704 with **no dead trailing return**. *This is the shared primitive #849 (ADR-114) consumes — keep it as a single source of truth (project "No Duplicate Code Paths" rule).*
- [ ] **E0705**: `forever` in a non-`void` function is an error. Message/help per ADR-113 "Proposed error codes" table.
[...]

Issue #849, opened 2026-02-19
MISRA C:2012 Rule 2.2: No dead code (28 violations)


jlaustill:
## MISRA Rule

**Rule 2.2 (Mandatory)**: There shall be no dead code.

## Violation Count

28 violations across test files.

## Description

Code that can never be executed must be removed.

## Example Violation

```c
return 0;
x = 5;  // MISRA 2.2 violation - dead code after return
```

## Fix Strategy

1. Review generated code patterns for unreachable statements
2. May be triggered by test patterns - investigate specific cases

## Related

- Part of #839 breakdown
- MISRA C:2012 Mandatory rule

jlaustill:
Cross-referencing a new design doc: **ADR-114 — Dead-Code / Reachability Analysis** (`docs/decisions/adr-114-dead-code-reachability.md`, Status: Research) proposes the reachability pass that would enforce this rule (MISRA C:2012 Rule 2.1 *unreachable code* and Rule 2.2 *no dead code*).

Key points relevant to this issue:
- C-Next currently has **no** reachability analysis — `return 1; side();` transpiles straight through with the dead call intact.
- The pass would reuse the dual of `ReturnPathAnalyzer` (ADR-112 already computes "statements after an unconditional return are unreachable").
- It's a **breaking change**; the one known offender in-repo (the dead `return 0;` in `examples/nucleo-f446re/blink.cnx`) is removed first by **ADR-113 (Forever Loops)**, which this ADR is sequenced after.
- Proposed diagnostic: **E0706 — unreachable code**.

ADR is on branch `docs/adr-forever-and-dead-code` (PR forthcoming). Filed as Research — ready to be worked.

jlaustill:
**Blocked by #1074** (ADR-113 `forever` core). ADR-113 introduces the *divergence primitive* this reachability pass consumes (single source of truth — see ADR-114 "Relationship to ADR-113"), and #1074 migrates `examples/nucleo-f446re/blink.cnx` to `void main()` + `forever`, removing the one known dead-`return 0;` offender before this breaking rule turns on.

ADR-113 implementation is now tracked:
- #1074 — `forever` core (the blocker for this issue)
- #1075 — forbid disguised infinite loops (E0707, v0.2.18)
- #1076 — full MISRA 14.3 invariant enforcement (v0.2.19)

Note: the E0706 unreachable-after-`forever` diagnostic stays here (ADR-114), not in #1074 — see #1074's scope note.

[...]

PR #1088, opened 2026-06-28
docs: ADR-114 research findings — reachability analysis (#849)

jlaustill:
## Summary

Research-only update for **#849** (MISRA C:2012 Rules 2.1/2.2 — no unreachable / dead code), recording code-verified findings into **ADR-114 (Dead-Code / Reachability Analysis)**. The ADR remains **Status: Research** — this PR is the place to review that direction. **No language/behavior change and no implementation** are included.

## What changed

- `docs/decisions/adr-114-dead-code-reachability.md` only (one doc commit).

## Key findings

1. **ADR-114 is unblocked.** The sequencing blocker #1074 (ADR-113 `forever` core) is closed; the one known dead-`return 0;` offender in `examples/nucleo-f446re/blink.cnx` is gone (now `void main() { setup(); forever { loop(); } }`).
2. **The divergence primitive is NOT verbatim-reusable (the important one).** `statementDefinitelyReturns()` in `ReturnPathAnalyzer.ts` answers *"does this path return a **value**?"* and returns `false` for a bare `return;`. Reachability needs *"does control fall through?"*, for which a bare `return;` **is** divergent — so reusing it as-is would miss `return; side();`. The correct, single-decision-site fix is to extract a shared `statementDiverges()` base and redefine the value-check on top of it (per "No Duplicate Code Paths").
3. **E0706** confirmed free and already reserved for this rule.
4. `ReachabilityAnalyzer` should follow the `ReturnPathAnalyzer` (logic/analysis) listener pattern registered in `runAnalyzers.ts`; layer constraints hold. (Note: the related E0707 lives in `output/codegen` — a different pattern that should *not* be copied here.)
5. Post-preprocessor concern is mild — analyzers run on the directly-parsed `.cnx` tree.

## ADR review

The proposed reachability rule is a **breaking C-Next change**. Per project rules, ADR-114 stays **Research** until @jlaustill approves the direction — that review can happen on this PR. No implementation will start until then.

Closes nothing — #849 remains open pending ADR acceptance + implementation.

### DIFF
--- a/docs/decisions/adr-114-dead-code-reachability.md
+++ b/docs/decisions/adr-114-dead-code-reachability.md
@@ -74,12 +74,16 @@ on any control-flow path from the function entry (proposed **E0706 — unreachab
   `runAnalyzers.ts` like the existing analyzers (listener + `analyze(tree)` returning
   `IAnalyzerError[]`).
 - **Single source of truth for divergence.** The "is this statement divergent?" decision is
-  owned by **one** primitive (the `statementDefinitelyReturns` / divergence logic that ADR-113
-  introduces into `ReturnPathAnalyzer`). This analyzer _consumes_ that primitive; it does not
-  re-implement it. If sharing requires it, the divergence helpers are extracted into a shared
-  `logic/analysis/` control-flow module that both `ReturnPathAnalyzer` and `ReachabilityAnalyzer`
-  import. This is mandatory under the project's "No Duplicate Code Paths" rule — two passes that
-  each decide "what diverges" independently would be a latent divergence bug.
+  owned by **one** primitive. ADR-113 landed `statementDefinitelyReturns` in `ReturnPathAnalyzer`,
+  but research (see _Research Findings_ below) shows it is **not** directly reusable: it answers
+  _"does this path return a **value**?"_ and deliberately reports `false` for a bare `return;`.
+  Reachability needs _"does control fall through?"_, for which a bare `return;` **is** divergent.
+  The two predicates coincide everywhere except bare `return;`. The required refactor is to
+  extract a shared `statementDiverges(stmt)` base into a `logic/analysis/` control-flow module
+  and redefine `statementDefinitelyReturns` on top of it (`diverges && terminal-returns-a-value`).
+  `ReachabilityAnalyzer` consumes `statementDiverges` directly; it does not re-implement it. This
+  is mandatory under the project's "No Duplicate Code Paths" rule — two passes that each decide
+  "what diverges" independently would be a latent divergence bug.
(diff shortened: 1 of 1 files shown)


jlaustill:
## PR Review — ADR-114 research findings (#849)

Docs-only PR; its content is *claims about the codebase*, so I verified each one against live source rather than reviewing logic. **Verdict: exceptionally accurate — recommend merge.**

### Claims verified ✅ (all TRUE)

| Claim | Result |
|---|---|
| #1074 closed 2026-06-27; `forever` impl in commit `80dc4123`; #849 still open | ✅ |
| `blink.cnx` main now `setup(); forever { loop(); }`, no `return 0;` | ✅ |
| `statementDefinitelyReturns()` special-cases `forever` w/ ADR-114 comment (lines 48–58) | ✅ |
| Bare `return;` → `false` via `returnStmt.expression() !== null` | ✅ (see #1 below) |
| ReturnPath is step 10 in `runAnalyzers.ts`, short-circuited via `collectErrors` | ✅ |
| Analyzers run on `CNextSourceParser.parse(source)` tree in `_transpileFile` | ✅ |
| E0707 lives in `output/codegen` (`TypeValidator.ts` + `ControlFlowGenerator.ts`) | ✅ |
| `logic/analysis/` imports nothing from `output/` | ✅ |
| E0706 free (zero `src/` refs), reserved in `docs/error-codes.md` | ✅ |

The standout finding — that `statementDefinitelyReturns` and a future `statementDiverges` coincide *everywhere except bare `return;`*, so reusing it verbatim would silently miss unreachable code after `return; side();` — is correct and exactly the "agree only by coincidence" latent divergence the No-Duplicate-Code-Paths rule warns about. Caught before any implementation. 👍

### Minor suggestions (non-blocking)

1. **Line-number drift (Research Finding 2):** the bare-return check is cited as "lines 28–29", but the actual *code* is on **line 29** alone (line 28 is the comment). Substantively fine; consider citing the function name instead of exact ranges since ADR line numbers rot quickly. (The `forever` "lines 48–58" cite is exact.)
2. **E0706 framing (Finding 4):** accurate that it's reserved in `docs/error-codes.md`, but only as a parenthetical in the *Related* section — not a formal table row with message/help/source like E0705/E0707. Wording doesn't overstate it; just note that implementation will need a real table entry.

### Pre-existing bug found while verifying (separate from this PR)

`src/transpiler/logic/analysis/runAnalyzers.ts:5` comment says *"All 10 analyzers run in sequence"* but there are now **11** steps (CommentExtractor is step 11). Notably this ADR is *more* accurate than the code — it correctly says "11 steps." Out of scope for this PR (doesn't touch that file); flagging so it can be tracked/fixed separately per the Bug Tracking rule.

🤖 Reviewed with Claude Code

Issue #1096, opened 2026-06-29
Bit-range assignment (write) with a const width emits runtime mask (1U << WIDTH) - 1 — UB at full width


jlaustill:
## Summary

A bit-range **assignment** (write) with a named-const (or non-decimal-literal) width,
e.g. `b[0, WIDTH] <- v`, lowers to a **runtime** mask `(1U << WIDTH) - 1`, exactly the
same defect fixed for **extraction** (reads) in #1094. The runtime form is:

1. **Undefined behavior at full width.** For a 32-bit operand and `WIDTH == 32`, it
   emits `1U << 32` — UB (shift ≥ width), which cppcheck reports as `shiftTooManyBits`.
2. **Needlessly non-constant.** Even for widths `< 32` it emits a runtime
   shift+subtract where a compile-time constant mask (matching the literal-width path)
   would do.

The read/extraction half of this bug was fixed in #1094; this issue tracks the
**write/assignment** half, which shares the same root cause but a different code path
(`BitUtils.multiBitWrite` / `RegisterUtils.extractBitRangeParams`, both of which pass
the generated width *string* to `BitUtils.generateMask` without resolving the const
via `tryEvaluateConstant` first).

## Reproduction

```cnx
// write_repro.cnx
void main() {
    const u8 W <- 32;
    u32 b <- 0;
    b[0, W] <- 5;
}
```

```
$ npx tsx src/index.ts write_repro.cnx
$ grep 'b =' write_repro.c
    b = (b & ~(((1U << W) - 1) << 0)) | ((5 & ((1U << W) - 1)) << 0);   // 1U << 32 -> UB

# Contrast — LITERAL width precomputes the mask (safe):
#   b[0, 32] <- 5;  ->  b = (b & ~(0xFFFFFFFFU << 0)) | ((5 & 0xFFFFFFFFU) << 0);
```

A `u64` write with a const width `> 32` correctly uses the `1ULL` base (the write path
already passes `targetType`), so it is value-correct but still **non-constant**; the u32
full-width case is the UB one.

## Root cause

`BitUtils.generateMask(width)` resolves the mask to a precomputed literal only when the
width string parses as a number. A const width arrives as the *identifier* (`"W"`), so
`Number.parseInt("W")` → `NaN` and it falls back to the runtime `((1U << W) - 1)`.
The write-path callers never run the width through `tryEvaluateConstant` before mask
generation. (See #1094 for the equivalent read-path fix, which resolves the width once
[...]

PR #1097, opened 2026-06-29
fix: precompute const-width bit-range extraction mask (#1094)

### LINKED ISSUE #1094: Bit-range extraction with a const width emits runtime mask (1U << WIDTH) - 1 — UB at full width, wrong for u64
## Summary

A bit-range extraction with a **named-const (or non-decimal-literal) width**, e.g. `b[0, WIDTH]`, lowers to a **runtime** mask `(1U << WIDTH) - 1`, whereas a **literal** width lowers to a **precomputed** mask. The runtime form is:

1. **Undefined behavior at full width.** For a 32-bit operand and `WIDTH == 32`, it emits `1U << 32` — UB (shift ≥ width), and cppcheck reports `shiftTooManyBits`.
2. **Wrong / UB for 64-bit widths.** The literal `1U` is `unsigned int` (32-bit), so any width `> 32` (e.g. a `u64` extraction) shifts past the base type's width — wrong result and UB. The literal-width path precomputes the correct wide mask instead.
3. **Needlessly non-constant.** Even for widths `< 32` it emits a runtime shift+subtract where a compile-time constant mask (matching the literal path) would do.

This was uncovered during the PR #1085 review while fixing composite slice-source typing for const widths (`a + b[0, WIDTH]`); that typing fix is correct, but it then routes the const width into this pre-existing extraction-codegen path. It is **independent of slices and composites** — a plain `u32 r <- b[0, W]` reproduces it.

## Reproduction

```cnx
// repro.cnx
void main() {
    const u8 W <- 32;
    u32 b <- 0x12345678;
    u32 r <- b[0, W];
}
```

```
$ npx tsx src/index.ts repro.cnx
$ grep 'r =' repro.c
    uint32_t r = ((b) & ((1U << W) - 1));     // 1U << 32  ->  UB

# Contrast — LITERAL width precomputes the mask (safe):
#   u32 r <- b[0, 32];   ->   uint32_t r = ((b) & 0xFFFFFFFFU);

$ cppcheck --addon=misra -I tests/include repro.c
repro.c:NN: error: Shifting 32-bit value by 32 bits is undefined behaviour [shiftTooManyBits]
```

## Root cause

The bit-extraction codegen resolves the width to a constant for the literal case (precomputing `0xFFFFFFFFU` etc.) but falls back to a runtime `(1U << width) - 1` expression when the width is a `const`/macro. The width is a compile-time constant in both cases (slice offset/length already resolve consts via `tryEvaluateConstant`), so the const path should precompute the same mask and use the correct base width (`1ULL`/wide literal for >32-bit widths).

## Suggested fix

In the bit-extraction lowering, resolve the width through the constant evaluator and emit a precomputed mask literal (same as the literal-width path), with the correct base type for widths > 32. Then a const width is byte-identical to the literal width.

## Related

- Same `1U << N` family as #1065 (MISRA 12.2 in generated code), but this is a **correctness/UB** bug (full-width and 64-bit cases), not only a style-rule violation; fixing it would also remove the const-width instances counted there.
- Surfaced during PR #1085 review (composite slice-source const-width typing fix).

jlaustill:
## Summary

Fixes #1094. A bit-range **extraction** with a const/macro width (`b[0, WIDTH]`) lowered to a **runtime** mask `((1U << WIDTH) - 1)` instead of the precomputed mask a literal width gets. That runtime form is:

- **UB at full width** — `WIDTH == 32` emits `1U << 32` (shift ≥ type width; cppcheck `shiftTooManyBits`). A full-width extraction silently returned `0`.
- **Wrong/UB for >32-bit widths** — the `1U` base is 32-bit, so a `u64` extraction with `WIDTH` in 33–63 shifts past it. The read path also never told the mask generator the operand was 64-bit, so even a **literal** `b[0, 40]` on a `u64` was UB.

## Root cause

`BitUtils.generateMask` only precomputes a mask when the width *string* parses as a number. A const width arrives as the identifier (`"W"`), so `Number.parseInt("W")` → `NaN` and it falls back to the runtime shift. The read path never ran the width through `tryEvaluateConstant`, and never passed operand 64-bitness.

## Fix

In `handleBitRangeSubscript` (`PostfixExpressionGenerator.ts`):

- Resolve the width **once** via `orchestrator.tryEvaluateConstant(exprs[1])` and format it `${n}U` so a const becomes **byte-identical** to the literal form.
- Pass operand 64-bitness (`u64`/`i64`) to `generateBitMask` so >32-bit widths use a `1ULL` / wide-hex base.

The single resolved value feeds **both** the integer and float branches, fixing the latent float twin (`x[0, WIDTH]` → `1ULL << WIDTH`) at the same site — no divergent path.

This removes the **MISRA C:2012 Rule 12.2** shift-overflow from these constructs; `validate:c` stays green.

## Tests

- `tests/bit-indexing/bit-range-const-width.test.cnx` — `test-execution`, u32/u64, full + partial + >32 widths (the full-width extraction returned `0` under the buggy mask).
- `tests/bit-indexing/bit-range-const-width-float.test.cnx` — `test-execution`, f32/f64 full-width twin.
- `PostfixExpressionGenerator.test.ts` — 4 unit tests: const resolves to `${n}U`, non-const flows through unchanged, `u64`/`f64` operands mark the mask 64-bit.
- `slice-const-width-source` snapshot — `WIDTH` now folds to a precomputed mask byte-identical to the literal form.

Both new tests were confirmed RED before the fix (integer and float, via `git stash`).

## Verification

- Full integration suite: **1006/1006** pass (C + C++).
- Unit suite (+4 new): pass.
- `validate:c` (cppcheck / MISRA / clang-tidy / flawfinder): **all passed**.

## Scope / follow-up

#1094 is scoped to **extraction** (reads). The **write/assignment** path (`b[0, WIDTH] <- v`) shares the identical root cause and is tracked separately as **#1096**.


src/transpiler/output/codegen/generators/expressions/PostfixExpressionGenerator.ts
```diff
@@ -1817,13 +1825,19 @@ const handleBitRangeSubscript = (
...
       state,
       orchestrator,
       effects,
     );
   } else {
-    const mask = orchestrator.generateBitMask(width);
+    // Issue #1094: 64-bit operands need a 64-bit mask base (1ULL / wide hex) so
+    // widths > 32 don't shift past a 32-bit literal's width.
+    const is64BitOperand =
+      ctx.primaryTypeInfo?.baseType === "u64" ||
```
jlaustill, line 1838: **[important]** bug: The new `is64BitOperand` decision reads `ctx.primaryTypeInfo?.baseType`, which is the **root identifier's** type (`primaryTypeInfo = getVariableTypeInfo(rootIdentifier)`), not the resolved operand type. So a `u64`/`i64` **struct-field** operand still gets the 32-bit `1U` base — the exact `>32`-bit defect this PR fixes for plain variables.

Reproduced on this branch:

```cnx
struct Reg { u64 field; }
u32 main() {
    const u8 W40 <- 40;
    u64 big <- 0x123456789ABCDEF0;
    u64 a <- big[0, W40];       // plain var
    Reg r <- {field: 0x123456789ABCDEF0};
    u64 b <- r.field[0, W40];   // struct field
    if (a != b) return 1;
    return 0;
}
```

generates:

```c
uint64_t a = ((big) & ((1ULL << 40U) - 1));     // correct
uint64_t b = ((r.field) & ((1U << 40U) - 1));   // 1U base — UB
```

`gcc -Wshift-count-overflow`: `left shift count >= width of type`; the program returns `1` (`a != b` — the field read drops the upper 8 bits). The plain-variable and array-element forms emit `1ULL`, so the identical operation now **diverges by operand form** — the latent-divergence pattern the project's No-Duplicate-Code-Paths rule targets, and the gap is in the very decision this PR introduces. (Full-width `64` happens to be correct only because `maskHex` case 64 hardcodes the `ULL` hex even with `is64Bit=false`.)

On `main` the plain-var case was equally buggy, so this isn't *new* UB — but it leaves #1094's "wrong for u64" goal only partly met, and the struct-field read UB is currently untracked (#1096 is the write path only), so per the ZERO-EXCEPTIONS bug-tracking rule it should be fixed here or filed.

Suggested fix: derive the 64-bitness from the **resolved operand scalar type** rather than the root. The handler already tracks `currentStructType` through the member chain, and `CodeGenState.getMemberTypeInfo(structType, member)` returns the field's `baseType`. The narrowing-cast just below (`ctx.primaryTypeInfo.baseType` at ~line 1854) has the same root-type issue — a single shared operand-type resolution feeding both the mask base and the narrowing cast removes the var/array/field divergence.

src/transpiler/output/codegen/generators/expressions/PostfixExpressionGenerator.ts
```diff
@@ -1817,13 +1825,19 @@ const handleBitRangeSubscript = (
         baseType: ctx.primaryTypeInfo!.baseType,
         start,
         width,
+        maskWidth,
```
jlaustill, line 1828: **[moderate]** bug (pre-existing, surfaced by this change): the `isFloatType` routing a few lines up also reads `ctx.primaryTypeInfo?.baseType` (the root's type), so a **float struct-field** bit-range misroutes into the *integer* branch and applies bitwise `&` to a `float`:

```cnx
struct Reg { f32 field; }
... u32 x <- r.field[0, 8];
```
→ `uint32_t x = ((r.field) & 0xFFU);` → gcc: `error: invalid operands to binary & (have 'float' and 'unsigned int')` (non-compiling C), whereas a plain `f32` variable correctly uses the union type-pun shadow.

This is the **same root-type-vs-member-type flaw** as the `is64BitOperand` check above, just an older instance — the `isFloatType` line itself is unchanged by this PR, so it's out of the mask-fix scope. Flagging because it isn't tracked anywhere: please file a follow-up issue (sibling to #1096). Note a naive fix is insufficient — `handleFloatBitRange` keys its shadow on `__bits_${rootIdentifier}`, so two float fields of the same struct root would collide; the shadow must key on the full member path. The clean resolution is the shared resolved-operand-type approach from the comment above, applied to the routing decision too.

jlaustill (commented):
## Review Summary

**PR:** #1097 — fix: precompute const-width bit-range extraction mask (#1094)

### Overview
Solid, well-scoped bug fix. It resolves a const/macro bit-range **width** to its numeric value (formatted `${n}U`, byte-identical to the literal path) so the mask is precomputed instead of a runtime `((1U << W) - 1)`, and passes operand 64-bitness so `>32`-bit `u64` widths use a `1ULL` base. The single resolved `maskWidth` feeds both the integer and float branches — no divergent path within the function. Tests are thorough (RED-before-green confirmed, u32/u64/f32/f64, full + partial + >32 widths), all 17 CI checks pass, and the `slice-const-width-source` snapshot/comment correctly fold to the literal-equivalent mask. I verified the literal-width path is unchanged and that no existing u64/i64 **read** bit-range snapshots are affected by the new `U`→`ULL` suffix.

### Findings

| Severity | Count |
|----------|-------|
| Critical | 0 |
| Important | 1 |
| Moderate | 1 |
| Nit | 0 |

### Key Issues
Both stem from one root cause — `ctx.primaryTypeInfo` is the **root identifier's** type, not the resolved operand type — and both are reproduced on this branch:

- **[important]** The new `is64BitOperand` check misses `u64`/`i64` **struct-field** operands (`r.field[0, W]` → `1U << 40` — UB / truncated), so #1094's u64 fix is incomplete for member operands and diverges from the plain-var/array forms. Inline at the `is64BitOperand` block.
- **[moderate]** A pre-existing sibling: float **struct-field** bit-range misroutes to the integer path and emits non-compiling C (`float & 0xFFU`). Out of this PR's scope, but untracked — please file a follow-up.

Neither blocks merge on its own — the core fix is correct for plain variables and array elements. The important item is worth either resolving here (the handler already tracks `currentStructType`, so the operand scalar type is recoverable) or filing as a known limitation so the struct-field read UB is tracked.

### What Looks Good
- Const width resolved **once** and shared by the int + float branches — avoids a divergent path, exactly the project's single-source-of-truth principle.
- `${n}U` formatting makes a const width byte-identical to a literal width (clean parity, no special-casing downstream).
- Genuinely thorough execution tests with unique return codes, plus the float twin; RED-before-fix verified for both.
- Honest scoping: the write-path twin is correctly separated and tracked as #1096.

---

jlaustill:
Filed the two struct-field follow-ups from the review (both reproduced on this branch, distinct from #1096 which is the write path):

- **#1098** — `u64`/`i64` struct-field bit-range read uses a 32-bit `1U` mask base → UB for widths > 32 (the gap in the new `is64BitOperand` decision; plain-var/array forms are fixed by this PR, struct-field is not).
- **#1099** — `f32`/`f64` struct-field bit-range read misroutes to the integer path → non-compiling C (`float & 0xFFU`); pre-existing, same root cause.

Both stem from `primaryTypeInfo` being the root identifier's type rather than the resolved member type, and one fix (resolve the operand scalar type from the member chain) addresses both plus the narrowing-cast sibling. Not blockers for the plain-variable/array fix in this PR.

PR #1100, opened 2026-07-03
fix: bit-indexing a scalar function parameter emits shift-mask, not pointer deref

jlaustill:
## Bug

A bit-index on a **scalar (non-array) function parameter** mis-transpiled: the parameter was emitted as `const uint32_t*` and the bit-index became a raw array subscript instead of `(x >> N) & 1`. The same expression on a **local** variable was already correct — only scalar parameters were affected.

Minimal repro:
```cnx
scope T { public u8 f(u32 x) { if (x[4] = true) { return 1; } return 0; } }
```

Before (v0.2.17):
```c
uint8_t T_f(const uint32_t* x) { if (x[4U] == true) ... }   // WRONG: pointer + array subscript
```
After:
```c
uint8_t T_f(uint32_t x) { if (((x >> 4U) & 1) == true) ... } // correct: by-value + shift-mask
```

## Root cause

The pass-by-value analyzer treated a bit-indexed parameter as pointer-eligible, and the subscript classifier emitted an array subscript instead of shift-and-mask.

## Fix

Keep a scalar parameter with bit-indexing by-value and emit shift-and-mask. Simplifies `PassByValueAnalyzer` (net ~-90 lines) + `SubscriptClassifier`.

## Tests

- New regression suite `tests/bit-indexing/param-scalar-bit-access` (C+C++ parity).
- Updated `tests/params/param-array-indexing`.
- Added `CodeGenerator` + `SubscriptClassifier` unit tests.
- Full suite: **1007 passed**; `npm run typecheck` clean.

## Context

Found while transpiling a `u32` reset-status register decode in a downstream embedded project (bit-indexing a by-value parameter). Fixed upstream per the zero-workaround policy rather than hacking around it downstream.


jlaustill:
Tracking issue: #1101 (retroactive, per the no-PR-without-a-matching-issue convention).

Issue #985, opened 2026-03-01
Transpiler silently accepts undeclared global. function calls


jlaustill:
## Summary

When a `.cnx` file calls `global.X()` but no included header declares `X`, the transpiler silently generates code that references `X()` without a declaration. The generated C/C++ then fails to compile with "not declared in this scope."

The transpiler should emit an error telling the user to include the appropriate header.

## Reproduction

**test.cnx** (no include for `millis`):
```c-next
scope Test {
  void run() {
    u32 now <- global.millis();
  }
}
```

```bash
cnext test.cnx
# Compiles successfully — no error from transpiler
# But generated test.c references millis() with no declaration
```

**Generated test.c:**
```c
void Test_run(void) {
    uint32_t now = millis();  // millis not declared anywhere
}
```

C compiler then fails:
```
error: 'millis' was not declared in this scope
```

## Expected

Transpiler error, e.g.:
```
E0xxx: 'millis' is not declared in any included header at test.cnx:3
```

## Working case

Adding `#include "fake_lib.h"` (which declares `millis()`) makes it work — the include passes through to generated output.

## Context

Found in ogauge project: `gauge_temp.cnx` calls `global.millis()` but only includes `<lvgl.h>`. The transpiler generates `gauge_temp.cpp` with a bare `millis()` call and no `<Arduino.h>`, causing a compile error.

## Repro files

`/tmp/cnext-bugs/missing-global-decl/`
- `fake_lib.h` — minimal header declaring `millis()`
- `test.cnx` — triggers the bug (no include)
- `test_with_include.cnx` — working case (with include)

PR #1104, opened 2026-07-10
feat(preprocessor): recover full external symbols via per-file TU slices

jlaustill:
## Summary

cnext collects header symbols by parsing the **preprocessed** header. Fully preprocessing a framework header inlines its whole transitive tree (`driver/twai.h` → ~100KB of xtensa HAL); ANTLR error-recovery then **drops** real declarations and **fabricates phantom struct bodies**. The result: a struct arg to a pointer param came out by value (`twai_driver_install(g_config)` not `&g_config`), and an opaque handle came out by value (`lv_obj_t` not `lv_obj_t *`) — neither compiles. A prior name-only recovery silenced the undeclared-call check (#985) but left codegen blind.

## What changed

- **`ExternalDeclarationOracle`** preprocesses the include union with `#line` directives, splits it back **per source file**, and parses each small, macro-expanded slice (so a trailing `PRIVILEGED_FUNCTION` is gone and `vTaskDelay` parses; no 100KB blob for ANTLR to mangle) → **full signatures**.
- **Phantom-body clear:** a parallel clean-parse table is authoritative; any struct body the blob parse fabricated for a type the clean parse proves opaque is cleared (restores `lv_obj_t *`).
- **`execFile` (argv), not shell `exec`,** in the preprocessor — a `-D` value with spaces/parens (`-DARDUINO_BOARD="... (8 MB ...)"`) no longer breaks `/bin/sh`.
- **CallExprGenerator** falls back to the C symbol table for extern globals, so a struct-valued global passed to a pointer param gets `&` (e.g. `&lv_font_montserrat_32`).
- Function-like **macro names** (e.g. `pdMS_TO_TICKS`) are still recovered by name (`-dM`) — a macro has no signature to track.
- Regression test: `tests/regression/external-symbol-recovery`.

Builds on the three preceding commits (skip-assembler-headers, in-context header preprocessing, `CNEXT_CROSS_COMPILER`).

## Verification

- Full suite: **5867 passed**.
- Drove a real ESP32-S3 / Arduino / LVGL firmware project through it: transpiles clean and the firmware **compiles + links** with correct codegen (`twai_driver_install(&g_config)`, `lv_obj_t *`, `&lv_font_montserrat_32`, `vTaskDelay`).


src/transpiler/Transpiler.ts
```diff
@@ -628,13 +641,128 @@ class Transpiler {
...
+   * macros that standalone header preprocessing missed, by preprocessing each
+   * .cnx's C includes as a translation unit (predecessors first — the way the
+   * real compiler does). Requires a toolchain that can preprocess the target's
+   * headers; for cross targets set CNEXT_CROSS_COMPILER. Gated on a preprocess
+   * failure so clean projects pay nothing.
+   */
+  private async _collectExternalDeclarations(
+    input: IPipelineInput,
+  ): Promise<void> {
+    if (!this.anyHeaderPreprocessFailed) return;
```
jlaustill, line 731: **[critical]** bug: The whole recovery pass is gated on `anyHeaderPreprocessFailed`, but that flag is only ever set inside `getHeaderContent` (line 1328), which runs only on a live preprocess. On a warm header cache, every header short-circuits through `tryRestoreFromCache` (line 1198 → returns `true`), so `getHeaderContent` never runs and this flag stays `false` — recovery is skipped entirely.

Meanwhile the per-header cache written at line 1217 was populated *before* recovery ran, so it holds the *degraded* symbols (phantom struct bodies, missing/by-value framework functions). The corrections this method applies (`parseCHeader`, `clearStructTagHasBody`, `addExternalDeclarationNames`) mutate only the in-memory `CodeGenState.symbolTable` and are never written back to the cache.

Net effect: a clean build produces correct codegen (`twai_driver_install(&cfg)`, `lv_obj_t *`), but the very next incremental rebuild — headers unchanged, only the `.cnx` edited, the common case — restores the degraded cache, skips recovery, and regenerates the exact by-value / non-compiling output this PR set out to fix. Caching is on by default (`noCache` is opt-out), and all four new regression tests pass `noCache: true`, so this path has no coverage.

Worth confirming before merge. Two directions: persist the recovered corrections into the cache after this pass, or derive the recovery gate from cached state (e.g. a cached "this header fell back to raw" marker) so warm-cache builds still trigger it.

src/transpiler/logic/preprocessor/ExternalDeclarationOracle.ts
```diff
@@ -0,0 +1,197 @@
...
+  private static async preprocessLargestWorkingTu(
+    directives: string[],
+    preprocessor: Preprocessor,
+    base: IPreprocessOptions,
+  ): Promise<{ content: string; directives: string[] } | null> {
+    const remaining = directives;
+    // At most one drop per iteration; bound iterations by the directive count.
+    for (
+      let attempt = 0;
+      remaining.length > 0 && attempt <= directives.length;
```
jlaustill: **[important]** bug: `remaining` aliases `directives` (same array, not a copy), and the loop bound re-reads `directives.length` every iteration — which `remaining.splice()` on line 172 shrinks. So the effective bound is `attempt <= (initialCount - attempt)`, i.e. the loop bails once more than half the includes have been dropped, defeating the method's stated goal of building "the largest subset that preprocesses cleanly."

Concretely, with 4 includes where 3 are individually fragile: attempts run at len 4→3→2, then at `attempt=3`/`len=1` the condition `3 <= 1` is false and it returns `null` — never even trying the single surviving include that would have preprocessed fine. A project with many fragile framework headers (exactly the target scenario) silently recovers *nothing*.

Capture the count once before the loop:

```suggestion
    const remaining = directives;
    const initialCount = directives.length;
    // At most one drop per iteration; bound iterations by the directive count.
    for (
      let attempt = 0;
      remaining.length > 0 && attempt <= initialCount;
```

jlaustill (commented):
## Review Summary

**PR:** #1104 — feat(preprocessor): recover full external symbols via per-file TU slices

### Overview

Strong, well-documented change. The core idea — preprocess the include set as one translation unit (predecessors first, `#line` kept), split back per source file, and parse each small macro-expanded slice — is the right way to get full framework signatures without either the raw-text attribute-macro breakage or the 100KB-blob ANTLR mangling. The `execFile`-over-shell fix and the assembler-header skip are both clearly motivated and tested. My one significant concern is how the recovery pass interacts with the header cache.

### Findings

| Severity | Count |
|----------|-------|
| Critical | 1 |
| Important | 1 |
| Moderate | 1 |
| Nit | 0 |

### Key Issues

- **[critical] Recovery is skipped on a warm header cache**, so its corrections silently vanish on the common incremental rebuild — see the inline comment on `_collectExternalDeclarations`. All four new regression tests use `noCache: true`, so this path is untested. This is the one I'd most want confirmed before merge.
- **[important] `preprocessLargestWorkingTu` gives up once more than half the includes are dropped**, because `remaining` aliases `directives` and the loop bound re-reads the shrinking length — inline on `ExternalDeclarationOracle.ts`.
- **[moderate] `detectAssemblySyntax` can over-match `_ASMLANGUAGE` guards** on the raw-fallback path, dropping a real C header's declarations — inline on `detectAssemblySyntax.ts`, with a tightening that keeps every existing test green.

### What Looks Good

- The two-table approach (degraded main table + authoritative clean re-parse to clear phantom struct bodies) is a clean way to reconcile the codegen truth, and the doc comments explaining *why* are excellent.
- `execFile`/argv instead of a shell command string is exactly right for `-D` values containing spaces/parens, and the new test locks it in.
- The `CNEXT_CROSS_COMPILER` override and the `-imacros` include-order retry are both minimal and well-scoped, with focused unit tests and faithful miniature fixtures mirroring the real FreeRTOS/xtensa failure shapes.

---

src/transpiler/logic/detectAssemblySyntax.ts
```diff
@@ -0,0 +1,32 @@
...
+
+// GNU-assembler directives at line start. Header content is declarations, not
+// expression statements, so a leading `.<directive>` is never valid C here.
+// `.macro` is the key trigger — its body holds instruction mnemonics like `loop`.
+const ASSEMBLER_DIRECTIVE =
+  /^[ \t]*\.(macro|endm|section|global|globl|rept|endr|p2align)\b/m;
+
+function detectAssemblySyntax(content: string): boolean {
+  // Explicit "this translation unit is assembly" marker used by many arch headers.
+  if (/\b_ASMLANGUAGE\b/.test(content)) return true;
```
jlaustill: **[moderate]** bug: `\b_ASMLANGUAGE\b` matches the token *anywhere*, including in `#ifdef _ASMLANGUAGE` / `#ifndef _ASMLANGUAGE` *guards* that real C headers use to fence off asm vs. C sections (e.g. xtensa `xtruntime.h` wraps its C prototypes in `#ifndef _ASMLANGUAGE`).

The docstring argues guarded headers are safe because "preprocessing as C strips those blocks" — but that only holds when preprocessing *succeeds*. Per this function's own `@param`, it also runs on RAW content on the preprocess-fallback path, where the `#ifndef _ASMLANGUAGE` line is still present. Such a header would then be classified as assembler and its C declarations dropped entirely.

Tightening to the *definition* form avoids guard false-positives while still catching coreasm.h (`#define _ASMLANGUAGE`) — and keeps every existing unit test passing:

```suggestion
  if (/^\s*#\s*define\s+_ASMLANGUAGE\b/m.test(content)) return true;
```

src/transpiler/Transpiler.ts
jlaustill: Fixed in ad531667. The root cause was that the recovery *decision* didn't survive caching, only the (degraded) symbol data did. Now each header's "fell back to raw content" state is persisted as a `preprocessFailed` marker in the cache entry (CACHE_VERSION 7→8), and both the recovery gate (`anyHeaderPreprocessFailed`) and the `precedingHeaders` usability decision derive from that single field — so a warm-cache build re-arms the gate, re-runs the recovery pass, and re-applies the corrections onto the restored symbol table. Added a warm-cache integration test (drops the `noCache: true`) that fails before the fix (warm build regenerates the by-value output) and passes after.

src/transpiler/logic/preprocessor/ExternalDeclarationOracle.ts
jlaustill: Fixed in ad531667. Captured `initialCount = directives.length` once before the loop and bound on that instead of the aliased/shrinking length. Added a regression test with 5 includes where 3 are individually unpreprocessable (>half dropped) — before the fix the loop exhausts and recovers nothing; after, the two surviving good includes are still tried and `oracle_fn` is recovered.

src/transpiler/logic/detectAssemblySyntax.ts
jlaustill: Fixed in ad531667. Tightened to `/^\s*#\s*define\s+_ASMLANGUAGE\b/m` so only the `#define _ASMLANGUAGE` form matches, not an `#ifndef _ASMLANGUAGE` guard on the raw-fallback path. Added a regression test with an xtensa xtruntime.h-shaped header (C prototypes fenced behind `#ifndef _ASMLANGUAGE`) asserting it is NOT classified as assembler; all existing detection tests (incl. coreasm.h `#define _ASMLANGUAGE`) still pass.

jlaustill: Follow-up (7b7fb1ac): the final regex uses horizontal whitespace — `/^[ \t]*#[ \t]*define[ \t]+_ASMLANGUAGE\b/m` — to avoid a SonarCloud S8786 super-linear-backtracking flag on `\s`. Same semantics (only the `#define` form matches); a preprocessor directive is single-line so `[ \t]` is also more correct.

jlaustill: Final state (b86500bc): after discussion the detection was reworked for robustness, not just the smell. It's now a line-based scan (no multi-quantifier regex, so no S8786/ReDoS): each trimmed line's first token is checked against a set of GAS directives (`.macro`, `.section`, …), and the whole-file self-declaration marker is matched by exact token in its `#define` form only — now covering `_ASMLANGUAGE` (xtensa), `__ASSEMBLY__` (Linux/U-Boot/RTOS ports) and `__ASSEMBLER__`. The `#ifndef`/`#ifdef` GUARD forms are still deliberately NOT matched, so a real C header fencing prototypes behind them (surviving on the raw-fallback path) isn't misclassified. Tests cover the new markers and the guarded-header negative case. SonarCloud: 0 issues, 90.6% new-code coverage.

src/transpiler/Transpiler.ts
```diff
@@ -220,6 +281,10 @@ class Transpiler {
     await this._collectAllHeaderSymbols(input.headerFiles, result);
     CodeGenState.buildExternalStructFields();
 
+    // Issue #985 recovery: when standalone header preprocessing missed framework
+    // symbols, recover their declared names via translation-unit preprocessing.
+    await this._collectExternalDeclarations(input);
```
jlaustill, line 285: **[moderate]** design: This recovery pass registers recovered struct symbols (via `_parseRecoveredSlices` → `parseCHeader`) into `CodeGenState.symbolTable`, but it runs *after* `CodeGenState.buildExternalStructFields()` on the line above already snapshotted the field map.

`buildExternalStructFields()` copies struct fields out of the symbol table into `externalStructFields`, and that snapshot is the only source `InitializationAnalyzer.getStructFields()` consults for external structs (`InitializationAnalyzer.ts:455`). So a struct that only becomes known through #985 recovery — e.g. `widget_cfg_t` in the regression fixture — never lands in `externalStructFields`, and `isKnownStruct()` returns `false` for it. Net effect: recovered external structs escape init-completeness checking, while structs from cleanly-preprocessed headers get it.

This is a diagnostic gap, not a miscompile (codegen resolves fields from the symbol table directly, which is why the integration tests still pass), and it may well be intended — framework structs are routinely partially-initialized. But the behavior now depends on whether a header needed recovery, which is a subtle divergence. If unintended, re-running `CodeGenState.buildExternalStructFields()` after `_collectExternalDeclarations()` would fold the recovered structs back in. Worth a conscious decision either way.

jlaustill (commented):
## Review Summary

**PR:** #1104 — feat(preprocessor): recover full external symbols via per-file TU slices

### Overview
Recovers full external symbols (signatures, typedefs, opaque handles) for framework headers that defeat standalone preprocessing, by preprocessing the include set as a translation unit and slicing the output back per source file. Also adds `compile_commands.json` auto-discovery, an assembler-header skip, a `CNEXT_CROSS_COMPILER` override, and switches the preprocessor to `execFile` (argv). This is a high-quality, carefully-reasoned PR: the code is exceptionally well-documented, the failure modes it targets are precisely explained, and the test coverage is excellent (unit + integration + regression fixtures, including warm-cache and >half-dropped edge cases). Verified end-to-end on real ESP32-S3/LVGL firmware.

### Findings

| Severity | Count |
|----------|-------|
| Critical | 0 |
| Important | 0 |
| Moderate | 1 |
| Nit | 1 |

### Key Issues

- **[moderate] design** — `_collectExternalDeclarations()` runs *after* `CodeGenState.buildExternalStructFields()`, so structs recovered via the #985 oracle are absent from the `externalStructFields` snapshot that `InitializationAnalyzer` reads. See the inline comment. Likely benign (external framework structs legitimately allow partial init), but it creates a divergence — whether a framework struct is init-checked depends on whether its header needed recovery — so worth a conscious confirm.

- **[nit] quality** — `Transpiler.extractCIncludeDirectives()` regex `([<"][^>"]+[>"])` allows mismatched delimiters (e.g. `<foo.h"`). Harmless in practice — a malformed spec just fails preprocessing and gets dropped by `preprocessLargestWorkingTu` — but `(<[^>]+>|"[^"]+")` would reject it up front.

### What Looks Good

- The `execFile`-over-`exec` fix is the right call, and the `-DARDUINO_BOARD="... (8 MB ...)"` test pins exactly the shell-metacharacter breakage it prevents.
- `preprocessLargestWorkingTu`'s aliased-loop-bound comment (capturing `initialCount` once) plus its dedicated "keeps searching after dropping >half" regression test is a great catch-and-guard.
- Cross-cutting state (`preprocessFailed`) is threaded cleanly through the cache: `ICachedFileEntry`, `CacheManager` read/write, `CACHE_VERSION` bump, and the version-assertion test are all updated together — no dual-path drift.
- `detectAssemblySyntax` correctly distinguishes the `#define _ASMLANGUAGE` marker from the `#ifndef _ASMLANGUAGE` guard, with tests covering the xtruntime.h / kernel-guard raw-fallback shapes that would otherwise drop real C prototypes.
- The `CallExprGenerator` refactor extends the address-of type resolution (adding the extern-C-global fallback) while preserving the original expression-type / registry / pointer-variable precedence.

---

src/transpiler/Transpiler.ts
jlaustill: Fixed in c584f704. Confirmed the diagnosis end-to-end (a probe showed `symbolTable.getAllStructFields()` had `widget_cfg_t` after recovery but `getExternalStructFields()` was empty, so `isKnownStruct('widget_cfg_t')` was false). Fix: **relocated** the single `CodeGenState.buildExternalStructFields()` snapshot to *after* `_collectExternalDeclarations()` rather than adding a second call — the only consumer (InitializationAnalyzer.ts:455) runs in Stage 5, nothing reads the map before then, and `buildExternalStructFields()` clears+rebuilds from the symbol table (idempotent), so the old pre-recovery snapshot was redundant. Recovered structs now get the same use-before-init (E0381) checking as cleanly-preprocessed ones; partial struct literals are unaffected (a declaration with any initializer marks all fields initialized). Added a regression test that fails before / passes after (asserts `widget_cfg_t` lands in `externalStructFields`). Full suite: 5896 pass.

Issue #1108, opened 2026-07-12
Duplicate system #include emitted when source includes a header the transpiler also auto-injects


jlaustill:
## Summary

When a `.cnx` source `#include`s a system header that the transpiler also auto-injects (e.g. `<stdint.h>`, `<stdbool.h>`), the generated C contains the include **twice** — the source passthrough is not deduplicated against the transpiler's injected includes.

## Minimal reproduction

```cnx
#include <stdint.h>
#include <stdbool.h>

u8 flags <- 0;
void t() { flags[0] <- true; }
```

## Actual generated C (head)

```c
#include <stdint.h>     // passthrough from source
#include <stdbool.h>
#include <stdint.h>     // auto-injected by transpiler — duplicate
#include <stdbool.h>
```

## Expected

Each system include emitted once. The transpiler should dedupe its injected includes against includes already present (passed through) from the source.

## Context

Found in shipped example `examples/bit_test.cnx` (README-referenced), which explicitly includes `<stdint.h>`/`<stdbool.h>` at the top. Functionally harmless (system headers have their own guards) but violates the "clean, readable C" promise and is redundant output.

Issue #1107, opened 2026-07-12
No unused-variable (dead-code) detection: defined-but-never-used vars accepted silently (ADR-114 scope gap)


jlaustill:
## Summary

C-Next performs **no unused-variable (liveness) analysis**. A variable that is defined but never read or written transpiles cleanly with no diagnostic — for **locals and globals alike**. This is the "dead code" flavor of MISRA C:2012 Rule 2.2 that the ruleset leans on the compiler for, and C-Next enforces none of the 2.x "dead code" family yet (`docs/misra-compliance.md` lists 2.1–2.7 all "Not Enforced").

## Minimal reproductions (all transpile with zero diagnostics)

```cnx
// 1. dead local
void t() { u32 dead <- 0; }

// 2. dead global
u32 dead <- 0;
void t() { }

// 3. global "used" only via a compile-time property (never touches storage)
u32 counter <- 0;
void t() { u32 x <- counter.bit_length; }   // .bit_length is compile-time (32); counter's value is never read
```

## Expected

Compile error: variable defined but never used (proposed to live under the dead-code diagnostic family).

## Design direction

Per maintainer decision, this is being folded into **ADR-114 (Dead-Code / Reachability)** by widening its scope from *unreachable statements* to *dead code including never-used objects* — keeping all of MISRA 2.x "dead code" in a single analyzer (single source of truth for the "is this dead?" decision) rather than a separate pass. ADR-114 stays **Research** until the design is approved.

### Open design question (for the ADR)

Does access to a **compile-time property** (`counter.bit_length`, `buffer.element_count`, ADR-058) count as a "use"? It reads a property of the *type*, never the storage — so under strict liveness `counter` is still dead. This needs an explicit call in ADR-114.

## Context

Found while auditing shipped example `examples/bit_test.cnx` (README-referenced): `counter` (line 9) and `buffer` (line 7) are both defined and only ever read via compile-time `.bit_length` / `.element_count`.

jlaustill:
**Audit finding: still fully valid; one stale cross-reference in the body.**

Verified on `main` @ 62f494b9 — both cases transpile clean with zero diagnostics:

```c-next
u32 deadGlobal <- 0;
void t() { u32 dead <- 0; }
```
```c
uint32_t deadGlobal = 0U;
void t(void) { uint32_t dead = 0U; }
```

[...]

Issue #1106, opened 2026-07-12
Indexing a scalar with a second subscript (x[a][b]) silently miscompiles instead of type-erroring (ADR-036)


jlaustill:
## Summary

Indexing a **scalar** with a second subscript (`x[a][b]`) is silently accepted and miscompiled instead of raising a type error. Per **ADR-036 (Multi-dimensional Arrays, Implemented)**, indexing peels one array dimension per subscript; a `u8` is not an array, so the second subscript has no valid type. ADR-036 lists **"type enforcement"** and **"Safety"** as core drivers — this violates both.

## Minimal reproduction

```cnx
u8 flags <- 0;
void t() {
    flags[4][3] <- 5;   // flags is u8 (scalar); flags[4] is a single bit (ADR-007); [3] indexes a non-array
}
```

## Actual generated C

```c
flags = (uint8_t)((flags & ~(1U << 4)) | ((5 ? 1U : 0U) << 4));
```

The second subscript `[3]` is silently dropped and the value `5` is coerced to a bool (`5 ? 1U : 0U` → `1`), so it writes a single bit = 1 instead of erroring.

## Expected

Compile error: cannot index `flags` (type `u8`) with `[3]` — `flags[4]` is not an array. (The author likely meant the bit-range write `flags[4, 3] <- 5`, which compiles correctly.)

## Notes

- No ADR change needed — this is a type-checker gap against already-Implemented ADR-036 + ADR-007.
- Found in shipped example `examples/bit_test.cnx` line 18 (README-referenced). The example itself is being corrected to `flags[4, 3] <- 5` separately.
- The examples CI guard (`examples-transpile.test.ts`, #1048) only checks that generated C *compiles*, so wrong-but-valid C slipped through.

PR #1109, opened 2026-07-12
docs: fix stale README, correct mis-compiling bit_test example, extend ADR-114 to liveness

jlaustill:
## What this fixes

Started as a README accuracy check and uncovered three real transpiler bugs behind a stale, mis-compiling example. This PR handles the **docs/tracking half** (safe, no behavior change); the transpiler fixes are tracked as separate issues below.

### README (rotted after examples were reorganized)
- `examples/blink.cnx` → `examples/teensy4/blink.cnx` (6 refs, incl. every Usage command — all previously pointed at a non-existent file)
- `test-teensy/` removed → hardware verification now under `examples/teensy4` + `examples/nucleo-f446re`
- Quick-Example generated C missing MISRA unsigned-literal suffixes (`3`→`3U`, `1`→`1U`) — verified against current codegen
- `examples/` tree completed; "50+ ADRs" → "70+"

### `examples/bit_test.cnx`
Line 18 used `flags[4][3] <- 5`, which **silently miscompiled** to a single-bit write of `1` (dropped the width, coerced `5` to a bool). Corrected to the documented bit-range write `flags[4, 3] <- 5`, which generates a correct 3-bit field write. Verified; all 7 example-guard tests pass.

### ADR-114 (Dead-Code / Reachability, stays **Research**)
Extended to fold in **unused-variable (liveness)** detection as the MISRA 2.2 dead-code counterpart to reachability — one analyzer, one "is it dead?" decision. Added proposed **E0709**, coverage, and two open design questions (does compile-time property access count as a "use"?; cross-TU global liveness). Per maintainer direction.

## Bugs tracked upstream (not worked around)
- #1106 — scalar `x[a][b]` silently miscompiles instead of type-erroring (violates Implemented ADR-036/007)
- #1107 — no unused-variable detection (the ADR-114 liveness gap extended here)
- #1108 — duplicate system `#include` when source includes a header the transpiler also auto-injects

## Root cause the audit exposed
The examples CI guard (`examples-transpile.test.ts`, #1048) only checks that generated C **compiles**, not that it's semantically correct — so wrong-but-valid C rotted undetected. Worth a follow-up to strengthen.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,10 +37,10 @@ Generates clean C:
 ```c
 #define GPIO7_DR_TOGGLE (*(volatile uint32_t*)(0x42004000 + 0x8C))
 
-uint32_t LED_BIT = 3;
+uint32_t LED_BIT = 3U;
 
 void LED_toggle(void) {
-    GPIO7_DR_TOGGLE = (1 << LED_BIT);
+    GPIO7_DR_TOGGLE = (1U << LED_BIT);
 }
 ```
 
@@ -134,19 +134,19 @@ npm link
 
 ```bash
 # Transpile to C (output alongside input file)
-cnext examples/blink.cnx
+cnext examples/teensy4/blink.cnx
 
 # Explicit output path
-cnext examples/blink.cnx -o blink.c
+cnext examples/teensy4/blink.cnx -o blink.c
 
 # Parse only (syntax check)
-cnext examples/blink.cnx --parse
+cnext examples/teensy4/blink.cnx --parse
 
 # Output as C++ (.cpp)
-cnext examples/blink.cnx --cpp
+cnext examples/teensy4/blink.cnx --cpp
 
 # Target platform for atomic code generation (ADR-049)
-cnext examples/blink.cnx --target teensy41
+cnext examples/teensy4/blink.cnx --target teensy41
 
 # Separate output directories for code and headers
 cnext src/main.cnx -o build/src --header-out build/include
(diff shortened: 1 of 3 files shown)


jlaustill:
## Pre-merge verification

Reviewed against live codegen and the working tree rather than the diff alone.

**Verified:**
- ADR-114 `**Status:** Research` is unchanged — the diff only adds research context, coverage, and open questions, which the ADR rules permit.
- Every README path referenced now exists: `examples/teensy4/blink.cnx`, `examples/nucleo-f446re/`, `examples/{bit_test,references,structs,uart_buffer}.cnx`, and all five `docs/` links. `test-teensy/` is genuinely gone.
- ADR count: 76 files in `docs/decisions/` — "70+" is accurate (was "50+").
- **Quick-Example C output matches live codegen byte-for-byte.** Transpiled the exact `.cnx` snippet: emits `uint32_t LED_BIT = 3U;` and `GPIO7_DR_TOGGLE = (1U << LED_BIT);`, confirming both added `U` suffixes.
- `examples/bit_test.cnx`: `flags[4, 3] <- 5` generates a correct 3-bit field write —
  `flags = (uint8_t)((flags & ~(((1U << 3) - 1) << 4)) | ((5 & ((1U << 3) - 1)) << 4));`
  versus the single-bit write the old `flags[4][3]` produced.
- `npm run unit`: 5896 tests / 217 files passing.

**Incidental:** transpiling `examples/bit_test.cnx` reproduces #1108 live (`<stdint.h>` and `<stdbool.h>` each emitted twice). Tracked and fixed by #1110 — will re-verify against this file after that merges.

No behavior change in this PR. Merging.

PR #1110, opened 2026-07-12
fix(codegen): dedup system #include directives (#1108)

### LINKED ISSUE #1108: Duplicate system #include emitted when source includes a header the transpiler also auto-injects

jlaustill:
Fixes #1108.

## Problem

The generated `.c` could emit a system header twice, via two independent include paths:

1. A source that `#include`s `<stdint.h>`/`<stdbool.h>` passed it through, and `addAutoIncludes` then emitted it again from the `needs*` flags.
2. `generateSafeDivHelpers` emitted its **own** `#include <stdbool.h>` (its helpers return a `bool` error flag), independent of the `needs*` flags — a second code path for the same include.

## Fix (single include path)

- `addAutoIncludes` now dedups against includes already present in the output (passthrough source includes), via a small `extractIncludeTarget()` helper.
- The safe-div `bool` dependency is signalled through `requireInclude("stdbool")` when the safe-div effect is applied, so it flows through the one `addAutoIncludes` path. The helper no longer emits its own include. This **removes the duplicate code path** (per the project's "No Duplicate Code Paths" rule) rather than deduping around it — and makes the dependency explicit instead of relying on the incidental `bool` result declaration.

## Tests / verification

- TDD: added failing unit tests (passthrough-dedup for stdint/stdbool, and the safe-div helper path) in `RequireInclude.test.ts`; updated the `HelperGenerator` unit test to pin the new contract (helper emits no include).
- Regenerated 48 snapshots — diff is **exclusively** removed duplicate include lines plus the safe-div include relocating to the top block. Verified every changed snapshot still contains each system include exactly once.
- `npm run unit` (5899) and `npm run test:q` (1007, transpile + C compile) both green.
- Header (`.h`) path verified free of the same bug.

### DIFF
--- a/src/transpiler/output/codegen/CodeGenerator.ts
+++ b/src/transpiler/output/codegen/CodeGenerator.ts
@@ -459,6 +459,10 @@ export default class CodeGenerator implements IOrchestrator {
           CodeGenState.usedSafeDivOps.add(
             `${effect.operation}_${effect.cnxType}`,
           );
+          // ADR-051 safe-div helpers return a bool error flag. Route that
+          // dependency through the single include path (#1108) rather than
+          // letting the helper emit its own #include <stdbool.h>.
+          this.requireInclude("stdbool");
           break;
 
         // Type registration effects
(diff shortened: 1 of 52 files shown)


jlaustill:
## Pre-merge review

Reviewed the codegen change, probed two specific failure hypotheses, and audited all 48 snapshot files. Tested against `main` merged in (b769ad18).

### Architecture

The fix collapses two independent include paths into one — `generateSafeDivHelpers()` no longer emits its own `#include <stdbool.h>` and instead signals the dependency via `requireInclude("stdbool")` in `processEffects()`. That is the right shape: it removes a duplicate code path rather than patching both ends.

### Risk 1 — ordering dependency: **clear**

The dedup only works if every include-emitting step runs before `addAutoIncludes()`. Verified in `assembleOutput()`:

```
self-include → processIncludeDirectives → processPreprocessorDirectives → addAutoIncludes → addGeneratedHelpers → declarations
```

Nothing emits an include after `addAutoIncludes`, so the `present` set is complete when it is built.

### Risk 2 — `#include` nested in `#ifdef`: **clear, but surfaced a separate bug**

Concern was that `extractIncludeTarget` runs on `line.trim()` and so would treat a conditionally-included header as unconditionally present, causing a needed include to be dropped when the macro is undefined.

It does not, because C-Next **already hoists conditional includes out of their `#ifdef`** — verified identically on `main` and on this branch. The include is unconditional in the output, so the `present` set is accurate and no include is lost.

That hoisting is itself a real (pre-existing) defect. Filed as **#1112** with a minimal reproduction, confirmed present on `main` at b769ad18 and therefore not introduced here.

### Risk 3 — snapshot audit: **clear**

Every changed line across the 48 snapshots is an `#include` line or a blank line. No generated code changed. The two non-obvious hunks:

- **4 `+#include <stdbool.h>`** in `builtin-safe-div-valid.*` — a *move*, not a new emission: `<stdbool.h>` relocates from the helper block up into the auto-include block. Net count still 1.
- **8 `-#include <string.h>`** in `null-check/valid-c-prefix-*` — correct dedup. Those sources `#include <string.h>` themselves and the transpiler was auto-emitting a second copy for the `string<64>` declaration. Confirmed the passthrough survives: exactly 1 occurrence remains.

### Risk 4 — C++ snapshots

`.cpp` output still carries `<stdbool.h>`; the diff shows the same relocation as the `.c` path, not a new emission. Pre-existing C++-mode behavior, unchanged.

### Verification

- `npm run unit` — 5899 passing / 217 files (up 3, the new dedup tests)
- `npm run build && npm run test:q` — 1007/1007 integration tests passing
- Live end-to-end: `examples/bit_test.cnx` emitted `<stdint.h>` and `<stdbool.h>` **twice each** on `main`; on this branch each appears exactly once.

No defects found in this change. Merging.

Issue #1077, opened 2026-06-27
examples-transpile CI guard doesn't detect stale committed .c/.h output (silent drift)


jlaustill:
## Summary

The examples CI guard (`scripts/__tests__/examples-transpile.test.ts`) transpiles every
`examples/**/*.cnx` to confirm it **transpiles without error**, but it never diffs the freshly
generated `.c`/`.h` against the **committed** generated output. As a result, committed example
artifacts drift silently whenever the transpiler changes and the example isn't manually
regenerated.

## How it surfaced

While implementing #1074 (ADR-113 `forever`), migrating `examples/nucleo-f446re/blink.cnx` to
`void main()` + `forever` required regenerating `blink.c`/`blink.h`. The regeneration produced
**~250 lines** of unrelated diff: the committed artifacts predated several transpiler changes —
header format (`Generated by C-Next Transpiler from: blink.cnx`, `#include "blink.h"`), the
`__builtin_add_overflow` clamp helper, comment passthrough, and bitmap comment formatting. The
old committed `main` even read `while (true)`, which the current transpiler no longer emits.

None of this was caught because the guard only asserts "transpiles OK," not "committed output ==
fresh output."

## Impact

- Committed `examples/**` `.c`/`.h` can rot indefinitely, misrepresenting current transpiler output.
- PRs that touch an example get bloated with unrelated drift when the file is finally regenerated.
- A real output regression in an example would not be caught by CI as long as it still transpiles.

This is a single-source-of-truth gap: the committed artifact and the transpiler output are two
representations of the same thing that are allowed to disagree.

## Proposed fix

Extend the examples guard to regenerate each example to a temp location and assert byte-equality
with the committed `.c`/`.h` (the same contract integration tests already enforce for
`tests/**/*.expected.c`). Fail with a clear "run `cnx <example>` and commit the result" message.

Alternatively, add a `scripts/` check (e.g. `npm run examples:check`) wired into `test:all`.

## Acceptance

- [ ] CI fails when a committed `examples/**` `.c`/`.h` differs from fresh transpiler output.
- [ ] All current examples regenerated and committed so the baseline is clean.
- [ ] Failure message tells the contributor how to regenerate.

jlaustill:
Another concrete instance of this, surfaced while reviewing #1119.

#1119 changes generated scope-member names (`LED_on()` → `LED__on()`) and regenerates ~1900 snapshot files under `tests/` and `bugs/`. `examples/` is untouched:

```console
[...]

PR #1119, opened 2026-08-06
fix(codegen)!: make scope-qualified names injective — `__` separator + E0201 (fixes #1117)

### LINKED ISSUE #1117: Scope name mangling collides with the global namespace, producing non-compiling C
Found while verifying a claim I made during review of #1111. **Pre-existing on `main` @ b769ad18.**

## Summary

Scope members are mangled to `Scope_member`, but that mangled name is not reserved — a global variable may legally be declared with the identical name. Both then emit the same C symbol, and the generated translation unit does not compile.

## Minimal reproduction

```cnx
u8 Reg_flags <- 99;

scope Reg {
    u8 flags <- 7;

    u8 readBare() {
        return Reg_flags;
    }
}

u32 main() {
    u8 v <- Reg.readBare();
    if (v != 99) return 1;
    return 0;
}
```

Transpiles **successfully** (exit 0, no diagnostic).

## Generated C

```c
uint8_t Reg_flags = 99U;          /* the global */

/* Scope: Reg */
static uint8_t Reg_flags = 7U;    /* the scope member -- same identifier */

uint8_t Reg_readBare(void) {
    return Reg_flags;             /* which one? */
}
```

```
$ gcc -std=c99 -c c.c
error: redefinition of ‘Reg_flags’
   13 | static uint8_t Reg_flags = 7U;
      |                ^~~~~~~~~
note: previous definition of ‘Reg_flags’ with type ‘uint8_t’
   10 | uint8_t Reg_flags = 99U;
      |         ^~~~~~~~~
```

## Two defects

1. **Symbol collision.** `Scope_member` mangling shares a namespace with user-declared globals, and nothing detects the clash. The failure lands in the C compiler, not the transpiler.
2. **Ambiguous resolution.** `Reg_flags` written as a *bare name inside scope `Reg`* resolves and compiles — it reaches the scope member through its mangled name. Per ADR-057 a bare name resolves local → scope → global, and `Reg_flags` is neither a local nor a member named `Reg_flags`; it matches only because the symbol table is keyed by the mangled name. Generated naming is leaking into the source namespace.

Point 2 is independently observable without any collision:

```cnx
scope Reg {
    u8 flags <- 0;
    void t() { Reg_flags[4, 3] <- 5; }   // compiles; identical output to this.flags[4, 3] <- 5
}
```

## Expected

At minimum, reject the collision with a diagnostic rather than emitting a duplicate definition. Whether `Reg_flags` should resolve as a bare name at all is a language question — ADR-057 does not grant it, and it looks like an artifact of the symbol-table key rather than an intended spelling. That part likely needs ADR direction rather than a unilateral fix.

## Relevance to the MISRA backlog

This is the source-level version of what #1063 (Rule 5.8 — identifiers with external linkage shall be unique) and #1059 (Rule 8.6 — exactly one external definition) flag in generated output.

## Related

- ADR-016 (scopes), ADR-057 (name resolution)
- #1063, #1059, #1060 — MISRA identifier-uniqueness rules
- Surfaced while reviewing #1111 / #1115

jlaustill:
Fixes #1117. Implements ADR-063 (Identifier Syntax), which was accepted in this branch after correcting two defects in the version merged by #1118.

**BREAKING**: generated C symbol names change. `LED_on()` -> `LED__on()`. No `.cnx` source changes required.

## The bug

`Scope_member` was not injective, so distinct declarations could produce the same C identifier. Two classes, both transpiling with exit 0:

**1. Global vs scope member**
```cnx
u8 Reg_flags <- 99;
scope Reg { u8 flags <- 7; }
```
Both emitted `Reg_flags` -> `error: redefinition of 'Reg_flags'`.

**2. Scope vs scope** — the worse one
```cnx
scope A_B { u8 c <- 1;   u8 readIt() { return c; } }
scope A   { u8 B_c <- 2; u8 readIt() { return B_c; } }
```
Both emitted `A_B_c`, and `A_readIt()` was generated as `return A_B_c;` — **silently returning the other scope's variable**. A wrong-value miscompile, not a compile error.

## Why not a diagnostic

Rejected deliberately, and the ADR records why. Scopes exist so `Reg.flags` and a global `Reg_flags` *can* coexist. Erroring would reject well-formed programs to work around a code-generation artifact.

## The fix

An identifier may not end with `_` or contain `__`, reserving `__` as the qualified-name separator. Injective: in `S__M`, `S` cannot end with `_` and neither part can contain `__`, so exactly one `__` occurs and the split is unique. A plain identifier can never collide with a qualified one, because it cannot contain `__` at all.

Verified against the issue's exact reproductions, compiled and **executed**:
```c
uint8_t Reg_flags = 99U;        /* global  */
static uint8_t Reg__flags = 7U; /* member  */   A_B__c  vs  A__B_c
```

Defect 2 is resolved as a consequence: the type registry is keyed by the qualified name, so once that key is `Reg__flags`, source-level `Reg_flags` stops matching and generated naming no longer leaks into the source namespace.

## Two ADR corrections (before implementing)

1. **The migration survey was wrong.** ADR-063 claimed zero corpus identifiers with a leading underscore. `_handler` is in three tests, taught in `language-guide.md`, and used 21x in ADR-029. The original regex only matched primitive-type declarations and missed user-defined callback types.
2. **The rule was stricter than the proof requires.** Injectivity constrains only the separator's *left* boundary. Leading underscores now stay legal, so ADR-029 and the language guide remain correct as written, and migration cost is genuinely zero — verified across all **4541** distinct identifiers: 0 trailing, 0 consecutive, 1 leading (`_handler`).

Enforcement also moved from the lexer to an analyzer: a lexer rejection surfaces as `token recognition error at: '_'` with no code, location or help, and could not distinguish a declaration from a call to `__disable_irq()`.

## Why this touches so many files

The separator was constructed at **~80 independent sites** with ~20 using a shared helper, across five mutually-incompatible "canonical" helpers, plus 14 sites reverse-parsing by splitting on `_`. Per the project's no-duplicate-code-paths rule, unifying them was the fix, not a side quest.

`QualifiedCName` now owns the separator plus join/split/isQualified/isInScope. **Phase 3 changed zero generated bytes** — verified by a full `npm test` leaving `git status` clean — so the separator flip in Phase 4 is one line.

Two latent bugs fixed on the way, both invisible until the separator grew:
- `CppConstructorHelper.toQualifiedName` did `replaceAll("_", "::")`, which turns `Scope__member` into `Scope::::member`
- `TypeValidator` used `!includes("_")` as "not qualified", misjudging a global named `my_var`

## Judgment calls for review

- **E0227 left on a literal underscore.** ADR-063 makes the collision it guards against unrepresentable, so it looks obsolete — but retiring it changes what C-Next accepts and needs its own ADR. Behavior preserved verbatim.
- **Three sites reverted** to a literal `_`: `usedClampOps`/`usedSafeDivOps` keys (`add_u32`) are helper op keys, not qualifications. My batch conversion over-reached and suppressed every overflow/safe-div helper — caught by 120 integration failures.
- **Two tests called `Math_add(...)`**, spelling the generated C name in source. That only worked via the defect-2 leak; now `Math.add(...)` with a comment.
- Names with legitimate internal underscores are preserved: `Motor__MAX_SPEED`, `Handlers__on_point`, `GPIO_Controller__value`.

## Verification

| Gate | Result |
|---|---|
| Unit | **5978** pass |
| Integration | **1015** pass, real compile + execute |
| CLI integration | 28 pass |
| `validate:c` | cppcheck, clang-tidy, MISRA, flawfinder |
| Generated-file drift | **0** after re-running `--update` |
| oxlint / knip / cspell | clean |
| jscpd | 0.45% |
| Coverage (new analyzer) | 100% |

Docs verified by transpiling the README and language-guide examples rather than hand-editing them.

Also filed: jlaustill/vscode-c-next#8 (surface every diagnostic live; notes that `collectErrors()` drops `code`/`helpText` before they reach users).


src/transpiler/output/headers/adapters/HeaderSymbolAdapter.ts
```diff
@@ -204,16 +205,17 @@ class HeaderSymbolAdapter {
...
    * @example resolveArrayDimension("10") => "10"
    */
   private static resolveArrayDimension(dim: string): string {
-    // Qualified enum access (e.g., "EColor.COUNT") - convert dots to underscores
-    if (dim.includes(".")) {
-      return dim.replaceAll(".", "_");
+    // Qualified enum access (e.g., "EColor.COUNT") - join through the shared
+    // separator so the dimension matches the enum member emitted elsewhere.
+    if (dim.includes(QualifiedCName.SOURCE_SEPARATOR)) {
+      return QualifiedCName.join(dim);
```
jlaustill: **[important]** bug: `dim` here is the dimension **as written in C-Next source** (`State.COUNT`, `this.State.COUNT`), not a generated C name. `QualifiedCName.join(dim)` assumes the dotted path is already the full scope path — inside a scope it is not, so the header and the `.c` file derive the same name two different ways and the header does not compile.

Reproduced by running this branch:

```cnx
scope Motor {
    enum State { IDLE, RUN, COUNT }
    public u8[State.COUNT] counters;
}
```

```c
/* arrdim.c */ uint8_t Motor__counters[Motor__State__COUNT] = {0};
/* arrdim.h */ extern uint8_t Motor__counters[State__COUNT];
```

```
arrdim.h:30:32: error: 'State__COUNT' undeclared here (not in a function)
```

The `.c` path resolves through `CodeGenState.currentScope`; this one does not. With the `this.` spelling it is more visibly wrong — the header emits `extern uint8_t Motor__counters[this__State__COUNT];`, with `this` joined in as a name component.

The divergence is not a regression (the merge base emits `State_COUNT` vs `Motor_State_COUNT` — same shape, one underscore), but this PR rewrote this function body from `dim.replaceAll(".", "_")` to `QualifiedCName.join(dim)` and now documents `QualifiedCName` as the single source of truth for this decision. Routing a second, scope-blind derivation *through* the shared helper is exactly the "agree only by coincidence" case CLAUDE.md calls out — sharing the join function is not enough while each path independently decides *what to join*.

Suggest resolving the dimension through the same scope-aware resolution the `.c` path uses (strip `this.`/`global.`, prefix the declaring scope) so both sides share one derivation. `IVariableSymbol` already carries the owning scope.

Also worth noting the `@example` on line 211 (`resolveArrayDimension("EColor.COUNT") => "EColor__COUNT"`) only holds for a **top-level** enum; for a scoped one the correct answer is `Motor__State__COUNT`, which this function cannot currently produce.

src/utils/QualifiedCName.ts
```diff
@@ -0,0 +1,127 @@
...
+ * a different target language; see FormatUtils.getScopeSeparator().
+ */
+class QualifiedCName {
+  /**
+   * The separator between components of a qualified C name.
+   *
+   * ADR-063 reserves this sequence: no C-Next identifier may contain it, which is
+   * what makes the join injective. Enforced by E0201 (IdentifierSyntaxAnalyzer).
+   */
+  static readonly SEPARATOR = "__";
```
jlaustill, line 27: **[moderate]** design: `SEPARATOR = "__"` is applied unconditionally, including in `--cpp` output — and a double underscore is reserved differently in the two target languages.

- **C11 §7.1.3** reserves identifiers that *begin* with `__` (or `_` + uppercase). `Scope__member` does not, so **C mode is clean**.
- **C++ [lex.name]/3.2** reserves every identifier that *contains* `__` **anywhere**. So in `--cpp` mode every scope function, scope variable, enum member, bitmap and register is now an implementation-reserved identifier.

Confirmed with the toolchain this repo already ships:

```
$ clang-tidy rid.cpp -checks='-*,bugprone-reserved-identifier,cert-dcl37-c,cert-dcl51-cpp' -- -std=c++14
warning: declaration uses identifier 'S__buf', which is a reserved identifier [bugprone-reserved-identifier,cert-dcl37-c,cert-dcl51-cpp]
warning: declaration uses identifier 'S__init', which is a reserved identifier
$ clang-tidy rid.c  -checks='...' -- -std=c99      # nothing
$ cppcheck --addon=misra --std=c99 rid.c           # no 21.2
```

To be clear about the actual impact: **nothing breaks and nothing miscompiles.** No real toolchain collides with these names, and `npm run validate:c` will not flag it — `batch-validate.mjs` passes no `-checks=` to clang-tidy (so only the default `clang-analyzer-*` set runs), there is no `.clang-tidy` in the repo, and it fails only on `error:` lines. So this is a conformance property of the generated artifact, plus MISRA C++:2008 17-0-1 / AUTOSAR M17-0-1 for anyone running a C++ rule set.

The part worth acting on is the ADR, not the code. ADR-063 lines ~156/161 cite [lex.name]/3 as a reason to ban `__` in user identifiers and conclude the rule "removes an existing standards-compliance defect rather than introducing a new restriction." For `--cpp` output that is now backwards: before this PR the only way to emit a reserved identifier was a user writing `my__var` (0 occurrences across the 4541-identifier corpus); after, it is universal. Neither Consequences nor Open Questions records the tradeoff.

I would not re-litigate the separator over this — the injectivity argument is sound and C mode is the primary target. But the ADR should say what was actually chosen: either add it to Consequences and correct the line-161 claim, or (if `--cpp` conformance matters) emit real `namespace Scope { }` blocks in C++ mode so the flat form never reaches a C++ TU.

src/utils/QualifiedCName.ts
```diff
@@ -0,0 +1,127 @@
...
+  static join(...components: (string | undefined | null)[]): string {
+    return QualifiedCName.toParts(components).join(QualifiedCName.SEPARATOR);
+  }
+
+  /**
+   * Split a qualified C name back into its components.
+   *
+   * Exact inverse of join() for names built from ADR-063-conformant identifiers.
+   *
+   * @param qualifiedName A generated C identifier, e.g. "Motor_State_IDLE"
```
jlaustill: **[moderate]** quality: three JSDoc examples in this file still describe the **old** single-underscore separator — in the module the PR designates as the single source of truth. Verified by executing the branch file under `tsx`:

| Doc says | Actually returns |
|---|---|
| L42 `join(...) → "Motor_State_IDLE"` | `"Motor__State__IDLE"` |
| L53-54 `split("Motor_State_IDLE") → ["Motor","State","IDLE"]` | `["Motor_State_IDLE"]` |
| L97 `toCppQualified: A_B → A::B` | `"A_B"` unchanged; only `A__B` converts |

The `split()` one is the most consequential: it documents precisely the non-injective decomposition ADR-063 exists to abolish. A maintainer who trusts it and passes a legacy or C-header name like `Reg_flags` to `split()` expecting `["Reg","flags"]` gets `["Reg_flags"]`, `parts[0]` becomes the whole name, and the scope-prefix branch is silently skipped.

```suggestion
   * @param qualifiedName A generated C identifier, e.g. "Motor__State__IDLE"
```

Also: line 42's `@returns`, line 97's `A_B -> A::B`, and line 5's `Scope_member`. Two call sites this PR edited carry the same staleness — `memberAccessChain.ts:89` still says the separator is `"_"` three lines above the `return QualifiedCName.SEPARATOR` the PR introduced, and `FunctionUtils.ts:64` still documents `"Outer_Inner_name"`.

Minor, while you are in here: "Exact inverse of join()" is slightly stronger than what holds — `join` drops empty components and expands dotted paths, so `split(join("A", "", "b"))` yields 2 parts from 3 arguments. It is an exact inverse of `toParts`, not of the argument list.

src/transpiler/logic/analysis/__tests__/IdentifierSyntaxAnalyzer.test.ts
```diff
@@ -0,0 +1,202 @@
...
+      [
+        "for-loop variable",
+        `void fn() { for (u8 i_ <- 0; i_ < 3; i_ +<- 1) { } }`,
+      ],
+      ["struct name", `struct Bad_ { u8 a; }`],
+      ["struct member", `struct Good { u8 field_; }`],
+      ["enum name", `enum Bad_ { A }`],
+      ["enum member", `enum Good { A_ }`],
+      ["bitmap name", `bitmap8 Bad_ { a[8] }`],
+      ["bitmap member", `bitmap8 Good { a_[8] }`],
```
jlaustill, line 166: **[moderate]** testing: this `it.each` table covers 11 of the analyzer's 13 declaration handlers — `enterRegisterDeclaration` and `enterRegisterMember` have no row.

The handlers themselves are correct and do fire (I checked `RegisterDeclarationContext.IDENTIFIER()` / `RegisterMemberContext.IDENTIFIER()` are non-nullable in the generated parser), and no integration fixture covers them either — the only ADR-063 `.cnx` fixtures are the two bare-global cases. So the gap is purely regression protection, but it protects something load-bearing: register names sit in the **middle** of a qualified name (`Timer__SysTick__CTRL`), which is exactly where a trailing underscore reopens the collision class.

Delete `enterRegisterDeclaration` and both `npm run unit` and `npm test` stay green, after which:

```cnx
scope S { register R_ @ 0x40 { M:  u8 rw @ 0 } }   // S__R___M
scope S { register R  @ 0x40 { _M: u8 rw @ 0 } }   // S__R___M   (leading _ is legal)
```

Both collapse onto one C identifier — the original miscompile class, with nothing failing.

```suggestion
      ["bitmap member", `bitmap8 Good { a_[8] }`],
      ["register name", `register Bad_ @ 0x40 { a: u8 rw @ 0 }`],
      ["register member", `register Good @ 0x40 { a_: u8 rw @ 0 }`],
```

Consider also asserting the handler count (13) the way the ADR-109 assignment-handler tests assert their registration count, so adding a grammar declaration rule without a matching handler fails loudly instead of silently widening the hole.

src/transpiler/logic/analysis/IdentifierSyntaxAnalyzer.ts
```diff
@@ -0,0 +1,203 @@
...
+
+/**
+ * Pure function creating the help text for a violation.
+ */
+function formatIdentifierSyntaxHelp(
+  identifierName: string,
+  violation: TIdentifierViolation,
+): string {
+  if (violation === "consecutive") {
+    return `Use a single underscore instead (e.g. '${identifierName.replaceAll(/_+/g, "_")}')`;
```
jlaustill: **[nit]** bug: the `consecutive` branch only collapses underscore runs, but `classifyIdentifier` returns `"consecutive"` *before* checking `endsWith("_")` — so a name that breaks both clauses gets a suggestion that E0201 itself rejects.

Ran the two pure functions standalone:

| Input | Suggested | Legal? |
|---|---|---|
| `my__value_` | `my_value_` | ✗ trailing |
| `a__b_` | `a_b_` | ✗ trailing |
| `__` | `_` | ✗ trailing |
| `my__value` | `my_value` | ✓ |
| `__value` | `_value` | ✓ |

So `u8 my__value_ <- 1;` reports correctly, the user applies the hint verbatim, and gets a second E0201. Cost is one extra edit cycle — the diagnostic itself is accurate and generated C is unaffected — but sanitizing against the whole rule rather than one clause is a one-line fix:

```suggestion
    return `Use a single underscore instead (e.g. '${identifierName.replaceAll(/_+/g, "_").replace(/_+$/, "")}')`;
```

Worth a test asserting `classifyIdentifier(<suggestion>) === null` for `my__value_`, `a__b_` and `__`. The test at lines 78-83 already feeds `my__value_` but asserts only `violation`, so the bad suggestion slips through.

src/transpiler/output/codegen/helpers/CppConstructorHelper.ts
```diff
@@ -21,8 +22,11 @@ class CppConstructorHelper {
    * e.g., TestNS_MyClass -> TestNS::MyClass
```
jlaustill: **[nit]** quality: this doc line now describes the behavior the change immediately below it deliberately removed. Under the new guard, `TestNS_MyClass` is returned **unchanged** — only `TestNS__MyClass` converts (verified: `isQualified("Adafruit_BMP280") === false`, `toCppQualified("A_B", "::") === "A_B"`).

```suggestion
   * e.g., TestNS__MyClass -> TestNS::MyClass
   * A single underscore is part of a component name and is preserved:
   * Adafruit_BMP280 -> Adafruit_BMP280 (not Adafruit::BMP280).
```

The code change here is a genuine fix worth calling out — the old `replaceAll("_", "::")` mangled every single-underscore C++ class name (`Adafruit_BMP280` → `Adafruit::BMP280`), silently failing constructor detection. Leaving the doc asserting the old contract invites someone to "restore" it.

docs/decisions/adr-016-scope.md
```diff
@@ -245,17 +244,17 @@ struct UART {
 }
 
 // Free functions that operate on UART
-void UART_init(UART* self, u32 base, u32 baud) {
+void UART__init(UART* self, u32 base, u32 baud) {
```
jlaustill: **[moderate]** quality: the rename was applied to a ```cnx block — these are user-written **C-Next source** identifiers, not generated C, and `__` in a declared identifier is exactly what E0201 now rejects.

`enterFunctionDeclaration → check() → classifyIdentifier()` returns `"consecutive"` for `UART__init`, so this ADR now documents C-Next source that this same PR makes illegal:

```
error[E0201]: Identifier 'UART__init' cannot contain consecutive underscores.
```

Single-underscore names remain perfectly legal under ADR-063 (`tick_count`, `SysTick_Handler`), so the original spelling was correct and should be restored.

```suggestion
void UART_init(UART* self, u32 base, u32 baud) {
```

Same for `UART__send` / `UART__receive` at lines ~253, 257, 266-271, and the call-site prose at 442-443.

(For what it's worth, the block was already illustrative rather than compilable — `UART* self` and `&uart1` are raw pointers, which ADR-006 does not permit — so nothing regressed. The issue is that the mechanical sweep did not distinguish `cnx` blocks from `c` blocks, and this is the one place where that distinction inverts the correct answer.)

jlaustill (commented):
## Review Summary

**PR:** #1119 — fix(codegen)!: make scope-qualified names injective — `__` separator + E0201

### Overview

The core of this PR is right, and the reasoning behind it is unusually well documented. I checked the injectivity claim independently rather than taking the proof on faith, including the edge case the ADR calls out: `scope Foo { u8 _handler }` generates `Foo___handler`, and `"Foo___handler".split("__")` returns `["Foo", "_handler"]` — correct, because JS `split` scans left-to-right non-overlapping and `S` cannot end with `_`. Both regression tests assert **runtime values** with unique return codes rather than merely "it compiles," which is what it takes to catch the original silent wrong-value miscompile. Unifying ~80 construction sites behind `QualifiedCName` is the correct response to the no-duplicate-code-paths rule rather than a side quest, and the three deliberate reversions to a literal `_` (helper-op keys like `add_u32`) are correctly identified and commented.

The findings below cluster into three themes: **one real header/`.c` divergence in changed code**, **two sets of committed generated artifacts the mass regeneration skipped**, and **documentation that the mechanical rename either missed or over-applied**.

### Findings

| Severity | Count |
|----------|-------|
| Critical | 0 |
| Important | 3 |
| Moderate | 6 |
| Nit | 5 |

### Key Issues

**1. `HeaderSymbolAdapter.resolveArrayDimension` — header and `.c` derive the same name two different ways** (inline comment on `HeaderSymbolAdapter.ts:218`)

The strongest finding, and it lands in code this PR rewrote. Reproduced end-to-end by running the branch:

```cnx
scope Motor { enum State { IDLE, RUN, COUNT }  public u8[State.COUNT] counters; }
```

```c
/* arrdim.c */ uint8_t Motor__counters[Motor__State__COUNT] = {0};
/* arrdim.h */ extern uint8_t Motor__counters[State__COUNT];   /* undeclared */
```

`gcc: error: 'State__COUNT' undeclared here (not in a function)`. With `u8[this.State.COUNT]` the header emits `this__State__COUNT` literally. The divergence pre-dates the PR in shape (merge base emits `State_COUNT` vs `Motor_State_COUNT`), but this PR rewrote that exact line and now designates `QualifiedCName.join` as the single source of truth for this decision — so per the project's "touching it means owning it" rule this is the moment to unify it rather than re-express the divergent derivation through the shared helper.

**2. Committed generated artifacts under `examples/` were not regenerated**

`git diff --name-only <base> <head> -- examples/` is empty, yet these carry the `Generated by C-Next Transpiler` marker and still hold the old names:

- `examples/teensy4/blink.{c,h}` — `LED_on`, `LED_toggle`, `Teensy4_doSomething`, `#define Teensy4_GPIO7_DataRegister`
- `examples/nucleo-f446re/blink.{c,h}` — `SysTick_init`, `LED_on`
- `examples/nucleo-f446re/test-nucleo/src/main.c`
- `examples/generated/structs.c` — `Console_printInt`
- `examples/teensy4/blink.cnx:108` — a comment claiming "Scoped register generates: `Teensy4_GPIO7_DataRegister`, …", now false

This makes the repo self-contradictory in a user-facing way: `README.md:191` (added by this PR) tells users `LED_on()` becomes `LED__on()`, while `examples/teensy4/blink.h` in the same tree still declares `void LED_on(void);`. Repo convention is to regenerate these on codegen changes (`chore: regenerate blink.c with extern const fix`, `chore: regenerate blink example with current transpiler output`).

CI structurally cannot catch this: `scripts/__tests__/examples-transpile.test.ts` transpiles into `mkdtempSync(join(tmpdir(), "cnext-examples-"))` and only runs `gcc -fsyntax-only` — it never diffs the committed artifacts. Worth extending it to compare against the committed output (same `--update` convention as `tests/`) so the next codegen change cannot silently desynchronize them. The verification table's "Generated-file drift: 0" is accurate but scoped to `tests/`.

**3. `tests-gtest/` — the repo's only demonstration of the C++-entry-point workflow no longer builds**

`git diff --stat <base> <head> -- tests-gtest/` is empty. `tests-gtest/cpp-entry/led.test.h` is committed generated output still declaring `void LED_on(void); void LED_off(void); uint8_t LED_getState(void); void LED_toggle(void);`, and `led_test.cpp` calls those names at ~20 sites. `add_cnext_test()` in `tests-gtest/CMakeLists.txt` re-runs the transpiler into `CMAKE_CURRENT_SOURCE_DIR` via `add_custom_command`, so `cmake --build` overwrites the header with `LED__on` etc. and the C++ wrapper fails to compile:

```
led_test.cpp:19: error: 'LED_off' was not declared in this scope
```

Silent because `tests-gtest` is not wired into CI (no hits for gtest across `.github/`, `package.json`, `scripts/`). `tests-gtest/README.md:49` also teaches `MyScope_getValue()`.

### Also flagged (details inline)

- **`__` is a reserved identifier in C++** — `Scope__member` is fine in C (C11 §7.1.3 reserves only *leading* `__`), but C++ [lex.name]/3.2 reserves any identifier *containing* `__` anywhere. I confirmed with the toolchain the repo already ships: `clang-tidy -checks='bugprone-reserved-identifier,cert-dcl51-cpp' -- -std=c++14` flags every generated scope member; the same source as C99 reports nothing, and `cppcheck --addon=misra` finds no Rule 21.2 violation. Not caught by `validate:c` (no `-checks=` flag, no `.clang-tidy`, and it only fails on `error:` lines), and no real toolchain breaks — so this is a conformance property of the certification artifact, not a bug. The gap is that ADR-063 cites [lex.name]/3 as justification for the rule and concludes it "removes an existing standards-compliance defect," while neither Consequences nor Open Questions records that the chosen separator universalizes that same construct across `--cpp` output.
- **Stale JSDoc in `QualifiedCName` itself** — `split("Motor_State_IDLE") → ["Motor","State","IDLE"]` documents precisely the non-injective decomposition ADR-063 abolishes, in the file designated as the single source of truth.
- **Analyzer test table omits 2 of 13 handlers** — register name and register member are unexercised.
- **Docs the rename missed** — `docs/cnext-ai-reference.md` (5 spots incl. the Transpilation Summary table) and `docs/language-guide.md:216-221`.
- **Docs the rename over-applied** — `docs/decisions/adr-016-scope.md` rewrote C-Next *source* identifiers to `UART__init`, which E0201 now rejects.
- Two surviving literal-`_` sites the sweep missed: `PostfixExpressionGenerator.ts:1763` (`ctx.result.split("_").at(-1)` in a help-text hint — pre-existing and cosmetic) and `CodeGenerator.ts:3537` (`${CodeGenState.currentScope}_` in a provably dead bitmap fallback). Neither is commentable inline as both sit outside the diff hunks.

### What looks good

- The injectivity proof is correct and the ADR revision that produced it is a genuine improvement over the merged version — relaxing the rule to match what the proof actually requires, rather than migrating the corpus to fit an over-strict rule, is the right call.
- Moving enforcement from the lexer to an analyzer is well justified: it buys an error code, a source-accurate location, and the declaration-vs-reference distinction that keeps `__disable_irq()` working.
- `resolveIdentifier` collapsing into `CodeGenState.resolveIdentifier` and `SymbolRegistry.resolveFunction` dropping its hand-rolled `replaceAll(".", "_") + "_"` prefix are real duplicate-path eliminations beyond the mechanical sweep.
- E0201 is cleanly registered in `docs/error-codes.md` with no code collision and the range/total counts updated; the `test-error` fixtures prove it fires end-to-end through the real pipeline, not just in unit isolation.
- The `README.md` migration note is exactly the right thing to have written for a breaking ABI change.

### Considered and dismissed

For transparency, three candidate findings were investigated and rejected rather than reported: that `IdentifierSyntaxAnalyzer` hard-codes `"__"` instead of reading `QualifiedCName.SEPARATOR` (the analyzer enforces the ADR-063 *language* rule, which is strictly stronger — `endsWith("_")` is not derivable from the constant, and the ADR's own alternatives table shows forbidding `__` alone is **not** injective, so deriving the check from the constant would invert the dependency and silently drop the boundary clause); the absent CHANGELOG entry (every CHANGELOG edit in this repo's history lands in a `chore: release vX` commit, never in a feature PR); and the ADR-063 `Research → Accepted` status flip (unfalsifiable from the tree, and the branch name and merge base indicate it was directed).

---

docs/cnext-ai-reference.md
```diff
@@ -452,8 +452,8 @@ scope Counter {
 }
 
 // External usage
-Counter.increment();                // → Counter_increment()
-u32 v <- Counter.get();             // → Counter_get()
+Counter.increment();                // → Counter__increment()
```
jlaustill, line 455: **[moderate]** quality: this hunk is correct, but the same mapping is stale elsewhere in this file and the document now contradicts itself. Grepping the branch version:

| Line | Current text | Should be |
|---|---|---|
| 367 | ``an `extern const` header symbol named `Scope_NAME` `` | `Scope__NAME` |
| 372 | `// -> extern const uint16_t Protocol_PGN;` | `Protocol__PGN` |
| 375 | `uint16_t p = Protocol_PGN;` | `Protocol__PGN` |
| 501 | ``- Names prefixed: `ScopeName_memberName` `` | `ScopeName__memberName` |
| 1246-1248 | `static void S_f(void) {}` / `void S_f(void) {}` / `S_f()` | `S__f` |

Ground truth on this branch: `tests/scope/scope-const-inline.expected.h` emits `extern const uint8_t ConstTest__PUBLIC_MAX;` and `uint8_t ConstTest__getMaxValue(void);`.

Lines 1246-1248 are the "Transpilation Summary" table — the densest and most-quoted mapping in the file — and this document is designated in the user's global CLAUDE.md as *the* AI reference for writing `.cnx` and for interop advice. A stale row there propagates straight into generated interop code: someone follows line 372, writes `extern const uint16_t Protocol_PGN;`, and gets `undefined reference to 'Protocol_PGN'`.

A `grep -nE 'Scope_NAME|ScopeName_|S_f\(' docs/` before merge would catch all five alongside the two hunks that were updated.

docs/language-guide.md
```diff
@@ -177,12 +177,12 @@ Transpiles to:
...
+static uint8_t LED__brightness = 100;
 
-void LED_on(void) { GPIO7_DR_SET = (1 << LED_BIT); }
-void LED_off(void) { GPIO7_DR_CLEAR = (1 << LED_BIT); }
+void LED__on(void) { GPIO7__DR_SET = (1 << LED_BIT); }
+void LED__off(void) { GPIO7__DR_CLEAR = (1 << LED_BIT); }
 
-uint8_t LED_getBrightness(void) { return LED_brightness; }
+uint8_t LED__getBrightness(void) { return LED__brightness; }
 ```
```
jlaustill, line 186: **[moderate]** quality: this "Transpiles to" block is correct, but the switch-statement example ~30 lines below it was missed. Lines 216, 220, 221 still read:

```c
case EState_IDLE:
case EState_RUNNING:
case EState_STOPPED:
```

Enum members go through the same separator on this branch — `src/transpiler/__tests__/Transpiler.test.ts:255-257` asserts a top-level `enum Color` now produces `Color__RED`/`Color__GREEN`/`Color__BLUE` — so actual output is `case EState__IDLE:` etc. A block explicitly labelled "Transpiles to:" that does not match the transpiler is worth fixing in the primary language guide.

One caution against a blanket replace: `struct Controller Controller_init(void)` at line ~319 is **correct as-is** — `StructGenerator`'s ADR-029 init function genuinely still uses a single underscore, since it is not a scope qualification.

docs/decisions/adr-063-identifier-syntax.md
```diff
@@ -56,43 +56,66 @@ scope A  { u8 _B; }   /* "A"  + "__" + "_B" = A___B */
...
+| **Forbid trailing and consecutive `_` (chosen)** | **Yes**   | **0 existing identifiers violate the rule**                                                                                     |
 
 ## Decision
 
-**An underscore in a C-Next identifier must sit between two alphanumeric characters.**
+**A C-Next identifier may not end with `_`, and may not contain two or more consecutive underscores.**
 
 ```
-IDENTIFIER : [A-Za-z] [A-Za-z0-9]* ('_' [A-Za-z0-9]+)*
+IDENTIFIER : '_'? [A-Za-z] [A-Za-z0-9]* ('_' [A-Za-z0-9]+)*
```
jlaustill: **[nit]** quality: this production is stricter than both the prose three lines above it and the analyzer that enforces it, and line 111 declares the grammar "**specification, not implementation**" — i.e. normative.

Line 75 states the rule as trailing-`_` and consecutive-`_` only. But `'_'? [A-Za-z] …` additionally requires a *letter* after an optional leading underscore, whereas `grammar/CNext.g4` is `[a-zA-Z_] [a-zA-Z0-9_]*` and `classifyIdentifier()` tests only `includes("__")` and `endsWith("_")`. So `u8 _1count <- 0;` transpiles cleanly but is not derivable from the normative grammar — a reader implementing a second front end from this ADR (the VS Code extension grammar, say) would reject a program the transpiler accepts.

The extra restriction also buys nothing: the injectivity proof at lines 88-90 appeals only to the separator's left boundary and to the absence of `__`, never to what a component begins with — which is the same reasoning that correctly permits a leading underscore.

```suggestion
IDENTIFIER : '_'? [A-Za-z0-9]+ ('_' [A-Za-z0-9]+)*
```

Separately, the diagnostic samples at lines 168-172 do not match what the transpiler emits. ADR shows `identifier 'value_' may not end with an underscore` plus a `help:` line; `tests/analysis/identifier-trailing-underscore.expected.error` is `Identifier 'value_' cannot end with an underscore. A trailing underscore would make scope-qualified names ambiguous in generated C.` with no `help:` line at all — consistent with the `collectErrors()` limitation this ADR already documents at line 202, but worth reconciling so the ADR shows real output.

jlaustill:
Follow-up to my review above — I split the findings by whether they belong to this PR or pre-date it, verified each pre-existing one against `main` (`078686bf`) with a minimal reproduction, and filed the ones that are not this PR's job.

**Filed as issues (pre-existing on `main`, not introduced here):**

| Issue | Finding | Note for this PR |
|---|---|---|
| #1122 | Scoped enum as an array dimension resolves differently in the `.h` than the `.c`; output does not compile | ⚠️ This PR **edits the defective method** (`HeaderSymbolAdapter.resolveArrayDimension`, `dim.replaceAll(".", "_")` → `QualifiedCName.join(dim)`). It changes the separator but not the missing scope awareness, so the bug survives verbatim as `State__COUNT` / `this__State__COUNT`. Expect a textual conflict when #1122 is fixed. |
| #1123 | `tests-gtest/` is not wired into CI | This PR is the trigger that turns the latent gap into a live break — `led_test.cpp` has 20 `LED_*` call sites and CMake regenerates the header into the source tree on build. Refreshing those files here is enough; the CI wiring is #1123. |
| #1124 | Bitmap bracket-indexing diagnostic is duplicated; the copy suggests a truncated symbol | `PostfixExpressionGenerator.ts:1762` — not in a diff hunk, so I could not comment inline. Note the minimal `QualifiedCName.split` swap would not actually fix it; the member name needs to be passed rather than reverse-engineered. |

I also added evidence to **#1077** (*examples-transpile CI guard doesn't detect stale committed output*) rather than filing a duplicate — it already covers the `examples/` drift exactly. Worth reading alongside the "committed generated artifacts under `examples/`" point in my review: on this branch `README.md` says `LED__on()` while `examples/teensy4/blink.h` still declares `void LED_on(void);`.

**Not filed — these are in this PR's changed code and are review feedback, not separate bugs:** the stale `QualifiedCName` JSDoc, the two missing analyzer test rows (register name / register member), the E0201 help text that can suggest a name E0201 rejects, the `docs/` spots the rename missed (`cnext-ai-reference.md`, `language-guide.md`) and the one it over-applied (`adr-016-scope.md`'s ```cnx block), and the ADR-063 C++ reserved-identifier consequence.

Happy to open a separate ADR-follow-up issue for the C++ `__` reserved-identifier question if you'd rather decide that outside this PR — it is a documentation/scope call on ADR-063 more than a code change, so I left it as review feedback for now.

src/transpiler/output/headers/adapters/HeaderSymbolAdapter.ts
jlaustill: Fixed in 5dfb077c — and you were right that sharing the join function was not the same as sharing the decision.

`resolveArrayDimension` now takes the declaring scope and resolves the same way the `.c` path does: `global.` drops the marker and adds no prefix, `this.` prefixes the scope (so `this` never becomes a component), and a bare dotted path is prefixed only when that scope actually declares the enum.

Reproduced your case first, then verified all three shapes now agree:

```
.c   Motor__fromScope[Motor__State__COUNT]   .h  Motor__fromScope[Motor__State__COUNT]
.c   Motor__fromGlobal[Global__COUNT]        .h  Motor__fromGlobal[Global__COUNT]
.c   Motor__viaThis[Motor__State__COUNT]     .h  Motor__viaThis[Motor__State__COUNT]
```

Header compiles. Added `tests/enum/scoped-enum-array-dimension-header.test.cnx` (execution, C+C++ parity) covering all three, plus four unit tests, one per branch. The `@example` on the old line 211 is corrected too.

src/utils/QualifiedCName.ts
jlaustill: Verified independently before acting — clang-tidy gives 2 reserved-identifier warnings under `-std=c++14` and 0 under `-std=c99`, and `batch-validate.mjs` indeed passes no `-checks=` with no `.clang-tidy` in the repo, so our own CI would not catch it.

You are right that the line-161 claim is backwards for `--cpp`. Fixed in 5dfb077c: the ADR now separates the two targets — a net improvement for C, an accepted regression for C++ — with the clang-tidy evidence, the note that nothing breaks or miscompiles, and the MISRA C++:2008 17-0-1 / AUTOSAR M17-0-1 exposure. Emitting real `namespace Scope { }` blocks in C++ mode is logged as an open question.

Took your advice not to re-litigate the separator over it.

src/utils/QualifiedCName.ts
jlaustill: Fixed in 5dfb077c. Particularly worth catching in the module this PR calls the single source of truth.

All three corrected and each claim re-verified by executing the branch:

```
join(Motor,State,IDLE)     = Motor__State__IDLE
split(Motor__State__IDLE)  = ["Motor","State","IDLE"]
toCppQualified(A__B, ::)   = A::B
```

src/transpiler/logic/analysis/__tests__/IdentifierSyntaxAnalyzer.test.ts
jlaustill: Fixed in 5dfb077c — added `register name` and `register member` rows to the table, so all 13 handlers are now covered.

src/transpiler/logic/analysis/IdentifierSyntaxAnalyzer.ts
jlaustill: Fixed in 5dfb077c. Confirmed the ordering issue you describe: `classifyIdentifier` returns `"consecutive"` before the `endsWith` check, so `my__value_` was suggested `my_value_`.

Extracted `suggestLegalIdentifier()`, which always applies both normalizations regardless of which clause was reported. Added a test that asserts the suggestion is itself accepted by the analyzer, so this cannot regress.

Note it is not user-visible yet — `collectErrors()` drops `helpText` — but it becomes visible once jlaustill/vscode-c-next#8 lands, which is exactly why it was worth fixing now.

src/transpiler/output/codegen/helpers/CppConstructorHelper.ts
jlaustill: Applied in 5dfb077c, including your `Adafruit_BMP280` line — it makes the preserved-single-underscore case concrete. Verified: `toQualifiedName("Adafruit_BMP280")` returns it unchanged, `toQualifiedName("TestNS__MyClass")` gives `TestNS::MyClass`.

docs/decisions/adr-016-scope.md
jlaustill: Fixed in 5dfb077c — this was the one place my mechanical sweep inverted the correct answer, exactly as you say. Reverted `UART_init`/`UART_send`/`UART_receive` in the cnx block and in the prose at 442-443.

Then re-scanned every fenced cnx block across all docs for declared identifiers containing `__`. Only remaining hits are inside `/* ... */` comments in ADR-063's own counterexample, where `A___B` is the generated result being illustrated — those are correct.

Also confirmed the neighbouring `Motor__State__IDLE` block at ~347 is inside a ```c fence, so it correctly keeps the new separator.

docs/cnext-ai-reference.md
jlaustill: Fixed in 5dfb077c — all four: the `Scope_NAME` prose, the `Protocol_PGN` comment and consumer line, and `ScopeName_memberName` at 501. Grepped the file afterwards for any remaining single-underscore generated names; clean.

docs/language-guide.md
jlaustill: Fixed in 5dfb077c — `case EState__IDLE:` / `__RUNNING` / `__STOPPED`. You are right that a block explicitly labelled the transpiler's output is the worst place to leave stale, since it reads as ground truth.

docs/decisions/adr-063-identifier-syntax.md
jlaustill: Both fixed in 5dfb077c.

On the production: confirmed `u8 _1count <- 5;` transpiles cleanly on this branch, so it was indeed accepted-but-not-derivable. I did not take the suggested replacement literally though — `'_'? [A-Za-z0-9]+ ...` has the mirror problem, admitting a leading digit that the lexer rejects. Since a single production has to admit `_1count`, reject `__a`, and reject a leading digit, I restated it as the unchanged lexer rule plus the two constraints, and noted why a one-line production is easy to get subtly wrong. That also matches how it is actually enforced.

On the samples: replaced with the real text from `identifier-trailing-underscore.expected.error` and its consecutive counterpart, with a note that `helpText` exists but is dropped before display (cross-referencing the limitation already recorded in Open Questions).

jlaustill:
## Review feedback addressed — 5dfb077c

All 10 comments resolved. Every claim was reproduced before fixing and re-verified after.

| # | Severity | Item | Outcome |
|---|---|---|---|
| 1 | important | Header array dimension diverges from `.c` | **Fixed** — scope-aware resolution + execution test + 4 unit tests |
| 2 | moderate | `__` reserved in C++ mode | **ADR corrected** — tradeoff recorded, namespaces logged as open question |
| 3 | moderate | Stale `QualifiedCName` JSDoc | **Fixed** — all 3, each verified by execution |
| 4 | moderate | Register handlers untested | **Fixed** — 13/13 handlers covered |
| 5 | moderate | ADR-016 renamed C-Next source | **Reverted** + all cnx blocks re-scanned |
| 6 | moderate | Stale `cnext-ai-reference` refs | **Fixed** — all 4 |
| 7 | moderate | Stale language-guide switch block | **Fixed** |
| 8 | nit | Help text suggests illegal name | **Fixed** — suggestion now provably legal |
| 9 | nit | Stale `CppConstructorHelper` doc | **Fixed** — used your `Adafruit_BMP280` line |
| 10 | nit | ADR grammar stricter than analyzer | **Fixed** — restated; samples now show real output |

### Two places I did not follow the suggestion literally

**#10 production.** `'_'? [A-Za-z0-9]+ ('_' [A-Za-z0-9]+)*` has the mirror problem — it admits a leading digit the lexer rejects. Since the production must admit `_1count`, reject `__a`, *and* reject a leading digit, I restated the rule as the unchanged lexer rule plus the two constraints rather than ship a production with the opposite subtle error.

**#2 separator.** Took your advice not to re-litigate it. Fixed the ADR's claim instead, which was the actual defect.

### Found while fixing

cppcheck caught an out-of-bounds index in my own new regression test — `enum State { IDLE, RUN, COUNT }` makes `COUNT` 2, so index 2 is past the end. Fixed; `validate:c` green.

### Verification

5989 unit · 1016 integration · 15 bugs · 28 CLI · validate:c · oxlint · cspell · knip · jscpd 0.45% · zero generated-file drift.

Coverage on changed files:

| File | Stmt | Branch | Func | Line |
|---|---|---|---|---|
| `IdentifierSyntaxAnalyzer.ts` | 100 | 100 | 100 | 100 |
| `QualifiedCName.ts` | 100 | 100 | 100 | 100 |
| `HeaderSymbolAdapter.ts` | 97.95 | 93.93 | 100 | 97.87 |

Thanks — the array-dimension one was a genuine bug that would have shipped, and the ADR-016 catch was my sweep inverting the correct answer in the one place where cnx and c blocks disagree.

src/transpiler/output/headers/adapters/HeaderSymbolAdapter.ts
jlaustill: Following up on my own reply — your framing ("each side independently decided what to join") deserved more than a symptom fix, and on re-examination it was still true after my change.

Filed as #1127. Verified the two derivations are structurally independent by perturbing **only** the header rule:

```diff
- if (parts[0] === "this") {
+ if (parts[0] === "PERTURBED") {
```
```c
/* .c  — unchanged path */ uint8_t Motor__viaThis[Motor__State__COUNT] = {0};
/* .h  — perturbed      */ extern uint8_t Motor__viaThis[this__State__COUNT];
```

A one-line edit to one side silently desynchronises the header. So the fix in this PR makes them *agree*, not *unified* — sharing `QualifiedCName.join()` shares the separator, not the decision.

Worth noting the codebase already has the right shape for **const** dimensions: they are resolved once during symbol collection and stored as a number on `IVariableSymbol.arrayDimensions`, so `resolveArrayDimension()` is never called for them and there is no second derivation. Enum-qualified dimensions take the string branch, which is the only reason the duplicate exists. #1127 proposes extending that same pattern — the #914 fix shape.

Leaving the scope-aware fix in this PR since it resolves the live miscompile, with the architectural cleanup tracked separately.

Issue #1125, opened 2026-08-06
ADR-063 follow-up: record the `__` separator tradeoff (C++ reserved identifiers + MISRA 5.1 cost)


jlaustill:
## Summary

ADR-063 cites **C++ [lex.name]/3** as a reason to forbid `__` inside *user-written* C-Next identifiers, then adopts `__` as the *transpiler's own* qualified-name separator. Because [lex.name]/3.2 reserves every identifier containing `__` **anywhere** (not just as a prefix), in `--cpp` mode every scope function, scope variable, enum member, bitmap and register becomes an identifier reserved to the implementation.

C mode is unaffected — C11 §7.1.3 reserves only identifiers that *begin* with `__` or `_`+uppercase — so this is specific to the C++ output path.

Nothing miscompiles and no real toolchain breaks. The ask is **not** to change the separator: after working through the alternatives (below) I think `__` is the right choice and the ADR picked correctly. The ask is that ADR-063 record the tradeoff explicitly, because as written it asserts the opposite of what the implementation does.

**Origin:** raised during review of #1119 and deliberately deferred there as an ADR/scope decision rather than a code change. See https://github.com/jlaustill/c-next/pull/1119#pullrequestreview-4875195582.

## The ADR text at issue

`docs/decisions/adr-063-identifier-syntax.md`, *Alignment with C and C++* — line 156 on the #1119 branch, line 120 on `main`:

> **C++ §lex.name/3** — identifiers containing `__` _anywhere_ are reserved to the implementation. C-Next currently passes `my__var` straight through, so `--cpp` output today declares reserved identifiers.

and its conclusion (line 161 on the branch, line 123 on `main`):

> The rule therefore removes an existing standards-compliance defect rather than introducing a new restriction.

The first sentence is correct and the defect it names is real. The conclusion holds for **C-Next source**, but for **generated C++** it is backwards: E0201 removes the narrow case (a user writing `my__var`) and the separator choice replaces it with a universal one.

Neither *Consequences → Breaking* (which lists only the symbol-name change table and snapshot/caller churn) nor *Non-breaking* nor *Open Questions* (whose single bullet is leading-underscore-at-file-scope under C11 §7.1.3) records this. The same is true of the `main` revision, so the gap is not an artifact of the in-flight edits.

### Why the reservation exists, and why C-Next is not entitled to it

[...]

PR #1128, opened 2026-08-06
test(scope): regression test for #1117 collision class 3 (qualification-depth ambiguity)

jlaustill:
Adds a regression test for a **third** #1117 collision class that ADR-063 does not currently document. Targets `fix/1117-adr-063-corrections` rather than `main`, because the test only passes with the `__` separator.

## The gap

ADR-063's Context lists two classes: a global vs a scope member, and a scope vs another scope. Both involve an underscore sitting in a *scope-name* position. There is a third that needs no such underscore — both declarations live in the **same** scope, and what collides is how many components the qualified name has:

```cnx
scope A {
    public enum B { c, d }   // A . B . c    three components
    public u8 B_c <- 7;      // A . B_c      two components
}
```

Under the old separator both mangled to `A_B_c`, so the generated header declared one identifier twice — as an enum constant and as an object:

```c
    A_B_c = 0,
extern uint8_t A_B_c;
```

```
error: 'A_B_c' redeclared as different kind of symbol
```

## Why it is worth pinning

This variant is what rules out the cheaper alternatives to `__`. The obvious one — keep a single `_` separator and merely forbid `_` in scope and enum names — does **not** fix it: `A` and `B` are already underscore-free. Injectivity requires a separator that a leaf name cannot contain at all, which is exactly what this PR's rule provides.

Recording it as an executable test means the constraint cannot be quietly relaxed later.

## Verification

- Fails on the merge base `078686bf` with the redeclaration error above.
- Passes here: `A__B__c` (enum member) vs `A__B_c` (scope member).
- `// test-execution` with unique return codes, so a silent mis-resolution fails too, not only the non-compiling form.
- `PASS ... [C+C++ PARITY]`; full `tests/scope` suite green (48 passed).
- Full pre-push gate green: prettier, cspell, oxlint, knip, antlr, typecheck, integration, unit, CLI, grammar coverage.

Follow-up suggestion for the ADR itself (not included here, since it is your call): add this as a third class in ADR-063's Context, alongside the note in #1125 about why it constrains the alternatives.

Refs #1117, #1125


jlaustill (commented):
## Review Summary

**PR:** #1128 — test(scope): regression test for #1117 collision class 3 (qualification-depth ambiguity)

### Overview

A genuinely new collision class, correctly reduced, and worth pinning: `A.B.c` (3 components) against `A.B_c` (2 components) is not covered by either class in ADR-063's Context, and both mangled to `A_B_c` before the `__` separator. I checked out the branch and ran it — `PASS ... [C+C++ PARITY]`. All nine snapshot files are present and consistent (`.test.c` byte-identical to `.expected.c`, `.h`/`.hpp` identical, matching the sibling `issue-1117-*` tests).

The findings below are about the rationale comment and the strength of the assertions, not about whether the case is worth having.

### Findings

| Severity | Count |
| --- | --- |
| Critical | 0 |
| Important | 1 |
| Moderate | 3 |
| Nit | 0 |

### Key issues

1. **L22 contradicts ADR-063.** "Injectivity needs a separator that a leaf name cannot contain at all" is the option ADR-063's alternatives table marks **not injective** — boundary underscores (`A_`/`B` vs `A`/`_B`) defeat it, which is why the decision also forbids a trailing `_`. The comment is copied into four generated snapshots, so it lands in five committed files, and the same wording is in the PR description. Relatedly, this test does not actually discriminate the `__`-separator-plus-forbid-`__`-only alternative — it passes under it; what it rules out are the two single-`_` variants.
2. **Two of the five return codes are unreachable.** `current <- this.B.c` then `current = this.B.c` (L37-38), and `state <- A.B.d` then `state != A.B.d` (L58-59), are tautologies under any mangling. So L24's "values are asserted so a silent mis-resolution fails" is not carried by them — and for this class there is no silent form anyway, since an enum constant and an object sharing a name in one header is always a hard C error. Codes 1/2/5 do real work. An ordinal comparison would make 3/4 discriminating; inline suggestion included.
3. **`this.` on L33 and L37 hides the implicit-resolution path.** L33's bare form works (`return B_c;` → `return A__B_c;`) and covers exactly #1117's companion defect that ADR-063 lists as fixed — free coverage. L37's bare form is broken: `B current <- B.c;` emits an unqualified type and does not compile. That one is a real transpiler bug, pre-existing on `main` — filed as **#1130** — so it should stay, but with a comment saying why.

### One thing not in the diff

ADR-063's Context still documents two collision classes and its alternatives table is the artifact this test defends. You flagged the ADR update as deferred, which is fair — but leaving it means the ADR and the suite disagree on how many classes exist, and the next reader of that table re-derives the conclusion L22 already got wrong. Adding a third Context class plus a table footnote is a research/context/findings update, which the project's ADR rules say to do; only Status and Decision need your sign-off.

### What looks good

- The comment explains *why*, not just *what* — the right instinct for a regression test whose job is to stop a constraint being quietly relaxed.
- `// test-execution` with unique return codes rather than a transpile-only snapshot, so the case is pinned behaviourally as well as textually.
- Naming and file layout match the two sibling `issue-1117-*` tests exactly; nothing to reconcile.

---

tests/scope/issue-1117-depth-ambiguity.test.cnx
```diff
@@ -0,0 +1,63 @@
...
+//
+//     A_B_c = 0,             -- the enum member  A.B.c
+//     extern uint8_t A_B_c;  -- the scope member A.B_c
+//
+//     error: 'A_B_c' redeclared as different kind of symbol
+//
+// This is why forbidding `__` inside identifiers is not on its own sufficient, and why
+// the cheaper alternative — keeping a single `_` separator and merely forbidding `_` in
+// scope/enum names — does not work either: `A` and `B` are both underscore-free here.
+// Injectivity needs a separator that a leaf name cannot contain at all.
```
jlaustill: **[important]** quality: This sentence contradicts ADR-063, and it is the one sentence the rest of the comment builds to.

ADR-063's alternatives table lists exactly this option and marks it **not** injective:

| Option | Injective |
| --- | --- |
| `__` separator, forbid only `__` in identifiers | **No** |

Its "Why a longer separator alone is insufficient" section gives the counterexample — a separator no leaf can contain is necessary but not sufficient, because underscores at a component *boundary* still collide:

```cnx
scope A_ { u8 B;  }   /* "A_" + "__" + "B"  = A___B */
scope A  { u8 _B; }   /* "A"  + "__" + "_B" = A___B */
```

Neither `A_` nor `_B` contains `__`, so both survive that check. Injectivity needs **both** halves of the decision: no `__` inside an identifier *and* no trailing `_`.

Worth being precise about what this test actually discriminates, too. It rules out the two single-`_` alternatives (both mangle to `A_B_c` → redeclaration). It does **not** rule out `__`-separator-plus-forbid-`__`-only: under that rule `A.B.c` → `A__B__c` and `A.B_c` → `A__B_c` are still distinct, so this test passes. The trailing-`_` half is pinned by `tests/analysis/identifier-trailing-underscore`, not here.

Since the block is copied verbatim into four generated snapshots, the claim lands in five committed files — and the same wording is in the PR description.

```suggestion
// Injectivity needs BOTH halves of ADR-063's rule: a separator no leaf name may
// contain (`__`), and no leaf name ending in `_`. This test pins the first half —
// the second is pinned by tests/analysis/identifier-trailing-underscore, whose
// counterexample (`A_`/`B` vs `A`/`_B`, both `A___B`) this case does not reach.
```

tests/scope/issue-1117-depth-ambiguity.test.cnx
```diff
@@ -0,0 +1,63 @@
...
+//
+// They are now `A__B__c` and `A__B_c`. Values are asserted so a silent mis-resolution
```
jlaustill: **[moderate]** testing: Two of the five return codes cannot fire, so this claim is not carried by the assertions.

- L37-38: `current` is initialized from `this.B.c` and then compared against `this.B.c`. Always true → `return 3` on L55 is unreachable.
- L58-59: `state` is initialized from `A.B.d` and then compared against `A.B.d`. Always false → `return 4` is unreachable.

Both hold no matter what `A.B.c` mangles to, including under a mis-resolution where the enum member and the variable are the same object.

There is also no silent form to catch in this class. An enum constant and an object with the same name in one header is always a hard C error, unlike the class-2 sibling where `A_readIt()` genuinely returned the other scope's storage. Codes 1, 2 and 5 do real work (they would catch source-level leakage of `A.B_c` onto the enum member); 3 and 4 are compile-time coverage only.

If you want a genuinely discriminating value check, compare the ordinal — `A.B.c` is 0 while `A.B_c` is 7, so any shared-storage resolution shows up as a mismatch:

```cnx
A.B first <- A.B.c;
u8 asNumber <- first;
if (asNumber != 0) return 4;   // A.B.c is 0, A.B_c is 7 - not one object
```

Verified locally: this transpiles to `uint8_t asNumber = first;` and exits 0.

Otherwise, softening L24-25 to say the assertions pin usability of both names rather than "values are asserted so a silent mis-resolution fails" would keep the comment accurate.

tests/scope/issue-1117-depth-ambiguity.test.cnx
```diff
@@ -0,0 +1,63 @@
...
+// fails the test too, not only the non-compiling form.
+
+scope A {
+    public enum B { c, d }
+
+    public u8 B_c <- 7;
+
+    public u8 readVariable() {
+        return this.B_c;
```
jlaustill: **[moderate]** testing: The bare form works here and would add coverage for free.

There is no local or global named `B_c` in this file, so nothing needs disambiguating — and per ADR-057 (and the repo's guidance on reserving `this.` for real conflicts) `return B_c;` is the idiomatic form. I checked: it transpiles correctly to `return A__B_c;`.

That matters more than style here. ADR-063's Consequences list #1117's companion defect as fixed "as a consequence": bare `Reg_flags` no longer resolves to the scope member `Reg.flags`, because the registry key became `Reg__flags`. Implicit resolution is precisely the path that regression covers, and `this.` routes around it. Dropping the prefix on this line exercises it at zero cost.

```suggestion
        return B_c;
```

tests/scope/issue-1117-depth-ambiguity.test.cnx
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    public bool enumIsFirst() {
+        this.B current <- this.B.c;
```
jlaustill, line 48: **[moderate]** quality: This `this.` is load-bearing — it is working around a transpiler bug, not disambiguating.

The bare form does not work today. `B current <- B.c;` inside `scope A` emits:

```c
B current = A__B__c;    /* value qualified, type not */
```

```
error: unknown type name 'B'
```

The value side resolves and the type side does not. Filed as #1130 (it also causes `.c`/`.h` signature divergence and a fabricated `typedef struct B B;` in the header for the param/return cases). Reproduces on `main` @ `1733ba70`, so it predates this PR — not something to fix here.

Since the project rule is that a workaround carries a TODO linking the issue, a line noting why the prefix is required would stop the next reader from "simplifying" it back to the broken form:

```suggestion
        // this.B is required until #1130: a bare scope-local type name is emitted
        // unqualified (`B`), which does not compile. The value side (this.B.c) is fine.
        this.B current <- this.B.c;
```

jlaustill:
All four addressed. Each verified against the transpiler before implementing, rather than taken on faith.

**1 — L22 contradicted ADR-063.** Confirmed: `adr-063-identifier-syntax.md:67` marks "`__` separator, forbid only `__` in identifiers" as **not** injective, with the `A_`/`_B` → `A___B` counterexample at L48-58. The sentence was wrong as written. Reworded to state that this case pins the *first* half of the rule and to name where the second half is pinned (`tests/analysis/identifier-trailing-underscore`). Also dropped the overclaim from the PR description.

**2 — Two of five return codes were unreachable.** Confirmed tautologies. Took the ordinal suggestion: enum members are 0 and 1 while the variable holds 7, so all five codes now compare real values and a collapsed resolution fails on value rather than only on compilation.

```c
uint8_t inScopeOrdinal = A__firstOrdinal();
if (inScopeOrdinal != 0) return 3;
A__B second = A__B__d;
uint8_t secondOrdinal = second;
if (secondOrdinal != 1) return 4;
```

**3 — Bare `B_c`.** Confirmed it emits `return A__B_c;`. Taken — free coverage of the implicit-resolution path, and `readVariable()` now carries a comment saying that is deliberate so it does not get "fixed" to `this.` later.

**4 — `this.B` is load-bearing.** Confirmed: the bare form emits `B current = A__B__c;` → `error: unknown type name 'B'`. Kept, now as an explicit `TODO(#1130)` rather than a bare explanation, and I added a matching revert checklist to #1130 so the two sides point at each other — previously the workaround was recorded as context on the issue but tracked as work on neither side.

**On the ADR.** Agreed the Context should gain a third class, and the alternatives table a footnote noting this case rules out the single-`_` variants but not the `__`-only one. That is a research/context edit, so it does not need sign-off — but it changes a file outside this PR's scope, so say the word and I will do it as a separate PR rather than widening this one.

Verification: `PASS [C+C++ PARITY]`, `tests/scope` 48 passed, full pre-push gate green (prettier, cspell, oxlint, knip, antlr, typecheck, integration, unit, CLI, grammar coverage). No Sonar surface — `sonar.sources=src` and this PR touches only `tests/`.

Issue #1135, opened 2026-08-07
scripts/generate-cpp-snapshots.ts is non-functional: missing `input` config and calls the removed `transpileSource`


jlaustill:
## Summary

`scripts/generate-cpp-snapshots.ts` — documented in `CLAUDE.md` under "Generate C++ snapshots" — cannot generate a snapshot for any file. It only ever appears to succeed because files that already have a `.expected.cpp` skip before reaching the broken code.

Two independent defects, both in `transpileAndWriteCppSnapshot()`:

1. **`ITranspilerConfig.input` is never passed.** The call sites construct `new Transpiler({ inputs: [], ... })`. `inputs` is not a field on `ITranspilerConfig`; `input` is required. The constructor calls `resolve(this.config.input)` at `Transpiler.ts:153` and throws `ERR_INVALID_ARG_TYPE`.
2. **`transpileSource` no longer exists.** With defect 1 patched, the next call fails with `pipeline.transpileSource is not a function`. The single-entry-point refactor replaced it with `transpile({ kind: "source", ... })`; `scripts/generate-cpp-snapshots.ts:149` is the last remaining caller in the repo.

Neither is caught by CI: the script runs under `tsx`, which does not typecheck, and nothing exercises it.

## Reproduction

Any `.cnx` file lacking a `.expected.cpp` triggers it. Using files already in the repo:

```console
$ npx tsx scripts/generate-cpp-snapshots.ts tests/include --dry-run
ERROR   tests/include/helper-types.cnx (helper)
        The "paths[0]" argument must be of type string. Received undefined
ERROR   tests/include/issue-294-decoder.cnx (helper)
        The "paths[0]" argument must be of type string. Received undefined
```

Stack, with the script's `catch` widened to report `err.stack`:

```
TypeError [ERR_INVALID_ARG_TYPE]: The "paths[0]" argument must be of type string. Received undefined
    at resolve (node:path:1257:7)
    at new Transpiler (src/transpiler/Transpiler.ts:153:26)
    at transpileAndWriteCppSnapshot (scripts/generate-cpp-snapshots.ts:142:22)
```

After patching only defect 1 (`input: cnxFile`):

```console
$ npx tsx scripts/generate-cpp-snapshots.ts tests/include --dry-run
ERROR   tests/include/helper-types.cnx (helper)
        pipeline.transpileSource is not a function
```

Verified on `main` @ `2e2317b0`.

## Note: the script may be redundant

[...]

PR #1136, opened 2026-08-07
fix(naming): reserve the cnx_ prefix and key include identity on path

### LINKED ISSUE #1131: Generated temporaries shadow user globals — two verified silent miscompiles (`_<var>_len`, `_tmp<N>`)
## Summary

Two families of transpiler-invented temporaries are emitted with names a user is allowed to declare. When the names coincide, the generated temporary **shadows** the user's variable and every subsequent read silently binds to the wrong storage.

Both produce **wrong runtime values with `gcc -std=c99 -Wall -Wextra` completely clean** and the transpiler exiting 0. This is the same failure class as #1117 collision class 2 — the silent one — in a family ADR-063 did not cover.

| Family | Built at | Verified result |
|---|---|---|
| `_<var>_len` (strlen cache) | `CodeGenerator.ts:1293` — `` `_${varName}_len` `` | expected **47**, got **10** |
| `_tmp<N>` (slice-assignment unroll) | `CodeGenState.ts:1247` — `` `_tmp${this.tempVarCounter++}` `` | expected **7**, got **0** |

## Reproduction 1 — `_<var>_len`

```cnx
u8 _msg_len <- 42;

u8 count() {
    string<16> msg <- "hello";
    u8 total <- 0;
    if (msg.char_count > 2) {
        total <- msg.char_count + _msg_len;
    }
    return total;
}

u32 main() { return (u32)count(); }
```

Generated:

```c
uint8_t _msg_len = 42U;          /* the user's global */

uint8_t count(void) {
    uint8_t total = 0U;
    size_t _msg_len = strlen(msg);   /* shadows it */
    if (_msg_len > 2) {
        total = _msg_len + _msg_len; /* source said: msg.char_count + _msg_len */
    }
    return total;
}
```

Look at the emitted expression: `total = _msg_len + _msg_len`. The source was `msg.char_count + _msg_len`, i.e. `5 + 42 = 47`. Both operands now resolve to the shadowing local.

```console
$ gcc -std=c99 -Wall -Wextra -o lensilent lensilent.c     # no diagnostics at all
$ ./lensilent; echo $?
10                                                        # expected 47
```

## Reproduction 2 — `_tmp<N>`

```cnx
u8[4] buffer;
u8 _tmp0 <- 7;

void store(u32 magic) {
    buffer[0, 4] <- magic;
    buffer[0] <- buffer[0] + _tmp0;
}

u32 main() {
    store(0);
    return (u32)buffer[0];
}
```

Generated — the MISRA 21.15 slice unroll introduces `_tmp0` inside the function:

```c
uint8_t _tmp0 = 7U;                              /* the user's global */

void store(uint32_t magic) {
    const uint32_t _tmp0 = (uint32_t)(magic);    /* shadows it */
    buffer[0] = (uint8_t)(_tmp0);
    ...
    buffer[0] = buffer[0U] + _tmp0;              /* adds magic, not 7 */
}
```

```console
$ gcc -std=c99 -Wall -Wextra -o tmpsilent tmpsilent.c     # no diagnostics at all
$ ./tmpsilent; echo $?
0                                                         # expected 7
```

Both verified on `main` @ `1733ba70`.

## Why nothing catches it

- `-Wall -Wextra` do not include `-Wshadow`. With `-Wshadow` gcc does report *"declaration of '_msg_len' shadows a global declaration"* — but that flag is not used by this project's pipeline and is not on by default anywhere.
- E0201 does not apply: `_msg_len` and `_tmp0` contain no `__` and no trailing `_`, so they are perfectly legal C-Next identifiers. A **leading** underscore is explicitly legal under ADR-063.
- `scripts/batch-validate.mjs` only ever runs over this repo's own generated `tests/` output, never over a user's project, so it could not catch a user-side collision regardless.

## Proposed fix — make it unrepresentable, not diagnosed

Both names are **purely invented** — no user-visible contract depends on their spelling. So they can simply be moved into a shape E0201 already forbids in user source:

```ts
// CodeGenerator.ts:1293
- const tempVar = `_${varName}_len`;
+ const tempVar = QualifiedCName.join("cnx", "len", varName);   // cnx__len__msg

// CodeGenState.ts:1247
- return `_tmp${this.tempVarCounter++}`;
+ return QualifiedCName.join("cnx", `tmp${this.tempVarCounter++}`);  // cnx__tmp0
```

Because a user identifier may not contain `__`, `cnx__len__msg` and `cnx__tmp0` are **unrepresentable in C-Next source**. That closes the family structurally, with **no new diagnostic and no new rule** — reusing the guarantee ADR-063 already bought. It is strictly better than reserving a prefix, which would only turn a silent miscompile into an error message.

The same treatment applies to `_cnx_tmp_<N>` (`ArgumentGenerator.ts:118`), which has the same shape and is not covered by a `cnx_` prefix rule anyway because it starts with an underscore.

Note this requires regenerating snapshots — `_tmp0` … `_tmp5` and `_<var>_len` appear in committed expected output.

## Related

- #1117 / ADR-063 — same collision class, different family. ADR-063 fixed scope-qualified names and explicitly scoped itself there; these invented temporaries were never in scope.
- The `cnx_` prefix reservation issue — these two families are listed there as gaps the prefix rule does **not** close, because a prefix ban would produce a diagnostic where this fix produces impossibility.

## Acceptance

- [ ] `_<var>_len` and `_tmp<N>` (and `_cnx_tmp_<N>`) are renamed into a `__`-containing shape.
- [ ] Both reproductions above become impossible to express rather than merely rejected.
- [ ] Execution regression tests for both, asserting the **runtime values** (47 and 7), not just that they compile — a compile-only test would pass today.
- [ ] Snapshots regenerated.

### LINKED ISSUE #1132: Reserve the `cnx_` prefix (case-insensitive, prefix-only) for the transpiler — E0202
## Summary

C-Next emits identifiers into the generated C that no user name may collide with. ADR-063 solved the **scope-qualified** family structurally (`Scope__member`, injective, enforced by E0201). It explicitly scoped itself there and left the transpiler's other generated names unaddressed.

This issue proposes reserving the **`cnx_` prefix, case-insensitively, prefix-only** for the transpiler — and, just as importantly, records the full enumeration of what C-Next puts into that namespace, because a prefix rule is only as good as its enumeration. **`cnx_` is necessary but not sufficient**: it closes two families and leaves six, three of which fail silently.

## Verified hole this closes

A user can declare a helper name today:

```cnx
u8 cnx_clamp_add_u8(u8 a, u8 b) { return 42; }

u32 main() {
    u8 counter <- 250;
    u8 step <- 10;
    counter +<- step;     // triggers the generated helper of the same name
    return 0;
}
```

```console
$ npx tsx src/index.ts collide.cnx     # exit 0, no diagnostic
$ gcc -std=c99 -fsyntax-only collide.c
collide.c:19:9: error: conflicting types for 'cnx_clamp_add_u8';
  have 'uint8_t(uint8_t, uint8_t)'
collide.c:11:23: note: previous definition with type 'uint8_t(uint8_t, uint32_t)'
```

Same shape as #1117 collision class 1: transpiler succeeds, C compiler rejects.

## The rule

> **A declared C-Next identifier may not begin with `cnx_`, compared case-insensitively.**

Predicate: `/^cnx_/i` against the declared identifier text.

| Identifier | Verdict | Why |
|---|---|---|
| `cnx_foo`, `CNX_FOO`, `Cnx_Foo`, `cNx_bar` | **rejected** | case-insensitive prefix match |
| `cnxFoo`, `cnx`, `cnxvalue` | legal | prefix is `cnx_`, not `cnx` |
| `mycnx_foo`, `my_cnx_foo` | legal | prefix-only, not substring |
| `_cnx_foo` | legal under this rule | leading `_`; see gaps below |
| `cnx__foo` | already rejected | E0201 (`__`) |

**Migration cost: zero.** All 7 corpus occurrences of `cnx_*` in `.cnx` files are inside comments (`// Uses cnx_clamp_add_u32`), not declarations — verified across `tests/`, `examples/`, `bugs/` and docs code blocks.

## Why a diagnostic is right here, when ADR-063 rejected one

ADR-063 says:

> The obvious response — detect the collision and emit an error — was considered and rejected. It imports a C limitation into C-Next and makes the programmer responsible for working around an artifact of the code generator.

That reasoning does not carry over, and the distinction is the justification:

- `Reg_flags` is a **legitimate name a user might want**. Rejecting it punishes them for a codegen artifact → fixed structurally with `__`.
- `cnx_clamp_add_u8` is **squatting on the implementation's namespace**. Nobody needs it → reserve it, exactly as C reserves `str*`, `mem*`, `is*`, `to*`, `_[A-Z]*` and POSIX reserves the `_t` suffix.

Reserving a vendor prefix is the standard solution, not a workaround. Two mechanisms for two different problems: `__` makes collisions *unrepresentable*; `cnx_` makes squatting *illegal*.

### (A) Closed by this rule

| Family | Site |
|---|---|
| `cnx_clamp_{add,sub,mul}_{u8..i64}` | `OverflowHelperTemplates.ts:71`, `AtomicGenerator.ts:105,173`, `SpecialHandlers.ts:93` |
| `cnx_safe_{div,mod}_*` | `HelperGenerator.ts:24`, `CallExprUtils.ts:58` |

### (B) Already unreachable — this issue must NOT claim to fix these

E0201 bans `__` anywhere in a user identifier, so these cannot be collided with today:

`__cnx_get_PRIMASK`, `__cnx_set_PRIMASK`, `__cnx_disable_irq` (`CodeGenerator.ts:4976-5002`), `__old`, `__new`, `__primask` (`AtomicGenerator.ts`), `__bits_<name>` (`AssignmentValidator.ts:121`, `BitRangeHelper.ts:84`, `PostfixExpressionGenerator.ts:1988`, `FloatBitHelper.ts:82`).

*(Separately, the `__cnx_*` names are C-reserved identifiers under C11 §7.1.3 — tracked in #1129, not here.)*

### (C) Gaps this rule does NOT close — six families

| Family | Site | Failure mode |
|---|---|---|
| `_<var>_len` strlen cache | `CodeGenerator.ts:1293` | **silent wrong value** → #1131 |
| `_tmp<N>` slice unroll | `CodeGenState.ts:1247` | **silent wrong value** → #1131 |
| `_cnx_tmp_<N>` | `ArgumentGenerator.ts:118` | leading `_`, so `/^cnx_/i` misses it → #1131 |
| `<FILE>_H` include guard | `HeaderGeneratorUtils.ts:39,42` | **silent** — erases the declaration; also generated-vs-generated collisions |
| `<Struct>_init` (ADR-029) | `StructGenerator.ts:204` | loud (redefinition) |
| `<fn>_fp` typedef (ADR-029) | `CodeGenerator.ts:2881` | loud |

The first three are best fixed by *renaming into a `__` shape* rather than by any prefix ban — see #1131, where `cnx__tmp0` becomes unrepresentable under the existing E0201 with no new diagnostic at all. The guard family is its own problem and is tracked separately.

`<Struct>_init` and `<fn>_fp` are derived from user names, so a prefix rule cannot cover them; they need either the same `__` treatment (`<Struct>__init`, `<fn>__fp`) or an explicit post-generation collision check.

## Enforcement

Add as a **third `TIdentifierViolation`** in the existing `IdentifierSyntaxAnalyzer` rather than a new analyzer — it already walks all 13 declaration contexts through a single `check()` entry point, so a new analyzer would duplicate that traversal and split the "what may a declared identifier be called" decision across two files.

Error code **E0202** (verified unused; correct `E02xx` Identifier/Param Naming range per `docs/error-codes.md`).

**Implementation trap:** `formatIdentifierSyntaxHelp()` builds its suggestion by collapsing underscores. For an input like `cnx__foo` it would suggest `cnx_foo`, which E0202 then rejects. The suggestion must be sanitized against the whole rule set, not one clause — same bug shape already noted for the trailing-underscore case.

## Acceptance

- [ ] A declared identifier matching `/^cnx_/i` is rejected with E0202.
- [ ] `cnxFoo`, `mycnx_foo`, `cnx` remain legal.
- [ ] E0202 registered in `docs/error-codes.md` with the range count updated.
- [ ] `.cnx` integration fixtures (`// test-error` + `.expected.error`) covering `cnx_`, `CNX_`, and a legal `cnxFoo`.
- [ ] Help text suggestions are legal under both E0201 and E0202.
- [ ] An ADR records the reservation and the enumeration above, including the families the rule does not cover.

## Related

- #1117 / ADR-063 — the qualified-name family; the rule this complements.
- #1131 — the three invented-temporary families, including two verified silent miscompiles.
- #1129 — `__cnx_*` are C-reserved identifiers (leading `__`).
- #1125 — reserved identifiers in generated C++.

### LINKED ISSUE #1133: Include-guard macros are not injective — two headers can share a guard, silently skipping one (wrong runtime value)
## Summary

Include-guard macros are built by uppercasing a filename's basename and collapsing every non-alphanumeric character to `_` (`HeaderGeneratorUtils.ts:36-42`). That mapping is **not injective**, so two different C-Next files can produce the same guard. When a consumer includes both headers, the preprocessor silently skips the second — and the resulting program can return a **wrong value** with only one warning.

A separate variant: a *user* symbol whose name equals the guard is erased by the macro.

Note this is a **generated-vs-generated** collision, so the `cnx_` prefix reservation (#1132) cannot fix it — nothing a user does is at fault.

## Cause 1 — sanitization collapse (verified, wrong runtime value)

```ts
// HeaderGeneratorUtils.ts:36-42
const sanitized = base.toUpperCase().replaceAll(/[^A-Z0-9]/g, "_");
...
return `${sanitized}_H`;
```

`mod-a.cnx` and `mod_a.cnx` both sanitize to `MOD_A`:

```cnx
// mod-a.cnx
scope Alpha { public u8  one() { return 1; } }

// mod_a.cnx
scope Beta  { public u64 big() { return 4294967296; } }
```

```console
$ grep -h '#ifndef' mod-a.h mod_a.h
#ifndef MOD_A_H
#ifndef MOD_A_H
```

A consumer including both:

```c
#include "mod-a.h"
#include "mod_a.h"      /* silently skipped — guard already defined */
int main(void) { printf("%llu\n", (unsigned long long)(Alpha__one() + Beta__big())); }
```

```console
$ gcc -std=c99 -Wall -Wextra -o consumer consumer.c mod-a.c mod_a.c
consumer.c: warning: implicit declaration of function 'Beta__big'
$ ./consumer
1                      # expected 4294967297
```

The `u64` return is truncated through the implicit `int` declaration. One warning, wrong answer, exit 0. With a `u8`-returning function the same collision produces the *correct* answer, which makes this intermittently invisible.

Verified on `main` @ `1733ba70`.

## Cause 2 — same basename in different directories

`can/config.cnx` and `uart/config.cnx` both produce `CONFIG_H`, because the guard is derived from the basename only. Reachable across separate transpiler invocations plus a hand-written consumer. *(Within a single compilation this is currently masked by a different bug — see the companion issue on same-basename includes being dropped — so the two fixes interact.)*

## Cause 3 — user symbol equals the guard (silent erasure)

```cnx
const u8 GUARDCOL_H <- 5;
```

```c
#ifndef GUARDCOL_H
#define GUARDCOL_H              /* expands to nothing */
extern const uint8_t GUARDCOL_H;
```

```console
$ gcc -E consumer.c | grep 'extern const'
extern const uint8_t ;                       /* the symbol is gone */
$ gcc -std=c99 -Wall -Wextra -fsyntax-only consumer.c
warning: useless type name in empty declaration
```

Warning only; the export silently disappears. This variant *would* be prevented by moving guards inside the reserved namespace (below).

## Proposed fix

1. **Make the guard injective over file identity.** Derive it from the input-root-relative path rather than `basename`, and *encode* rather than collapse non-alphanumerics (e.g. map `-` and `.` to distinct escapes) so `mod-a` and `mod_a` cannot converge.
2. **Add a hard diagnostic** when two headers in one compilation would emit the same guard. Injectivity alone is fragile — a future sanitization tweak silently reintroduces the bug; a check makes it loud.
3. **Move guards into the reserved namespace** — e.g. `CNX_GUARD_<PATH>_H`. Combined with #1132's case-insensitive `cnx_` reservation, this also closes Cause 3, since a user could no longer declare a name matching the guard.

Modern alternative worth considering: `#pragma once` is supported by every compiler C-Next targets and sidesteps the naming problem entirely — though it is not standard C, which matters given #1129.

## Also worth cleaning up

`IHeaderOptions.guardPrefix` is dead in production — it is referenced only at `BaseHeaderGenerator.ts:59`, `IHeaderOptions.ts:6` and two unit tests, and nothing in the CLI ever supplies it. If guards are being reworked, either wire it up or delete it.

## Related

- #1132 — `cnx_` prefix reservation. Fixes Cause 3 only if guards move into the reserved space; cannot fix Causes 1 or 2.
- #1131 — the other silent-collision family (invented temporaries).
- The companion issue on same-basename `.cnx` includes being silently dropped, which currently masks Cause 2 inside a single compilation.

## Acceptance

- [ ] `mod-a.cnx` and `mod_a.cnx` produce distinct guards.
- [ ] Two headers in one compilation emitting the same guard is a transpiler error, not a silent skip.
- [ ] A user declaration matching a guard macro cannot erase itself.
- [ ] Regression test asserting the **runtime value** (4294967297) for the Cause 1 repro — a compile-only test passes today.

### LINKED ISSUE #1134: Two included .cnx files sharing a basename: the second is silently dropped and C-Next source syntax leaks into the .c output
## Summary

When two included `.cnx` files share a basename, the second is **silently dropped from the entire compilation**. The transpiler exits 0, never generates that file's `.c`/`.h`, and emits raw **C-Next source syntax** into the C output for calls into the dropped scope.

Not a naming-collision bug — this is include resolution keying on `basename` instead of the resolved path.

## Reproduction

```
v3/
├── app.cnx
├── can/config.cnx
└── uart/config.cnx
```

```cnx
// can/config.cnx
scope CanConfig  { public u8 bitRate()  { return 5; } }

// uart/config.cnx
scope UartConfig { public u8 baudRate() { return 9; } }

// app.cnx
#include "can/config.cnx"
#include "uart/config.cnx"

u32 main() {
    u8 a <- CanConfig.bitRate();
    u8 b <- UartConfig.baudRate();
    if (a != 5) return 1;
    if (b != 9) return 2;
    return 0;
}
```

```console
$ npx tsx src/index.ts app.cnx
Generated 3 output files:
  can/config.c
  app.c
  can/config.h
$ echo $?
0
```

Only three files — `uart/config.c` and `uart/config.h` were never generated. And the generated `app.c`:

```c
uint8_t a = CanConfig__bitRate();
uint8_t b = UartConfig.baudRate();     /* C-Next source syntax, emitted verbatim */
```

`UartConfig.baudRate()` is not valid C. It fails downstream:

```console
$ gcc -std=c99 -fsyntax-only -I. app.c
app.c:7:10: fatal error: uart/config.h: No such file or directory
```

Verified on `main` @ `1733ba70`.

## Root cause

`Transpiler.ts:1034` dedupes candidate includes on the **basename**:

```ts
if (cnextBaseNames.has(includeBaseName) || /* ...resolved-path check... */) continue;
```

The second disjunct already dedupes correctly on the resolved absolute path, so the basename disjunct is both redundant for its stated purpose and actively harmful when two distinct files share a name — which is extremely common in embedded projects (`can/config`, `uart/config`, `spi/driver`, `i2c/driver`).

Two things make this worse than a missing file:

1. **No diagnostic.** A whole translation unit vanishing should not be silent.
2. **Source syntax leaks into output.** `UartConfig.baudRate()` reaching the `.c` file means the scope-resolution path has no guard for "scope not found" and falls through to emitting the raw text. That is worth fixing independently of the include bug — any future resolution miss would produce the same garbage.

## Proposed fix

- Dedupe solely on the resolved absolute path at `Transpiler.ts:1034`; drop the `cnextBaseNames.has(includeBaseName)` disjunct.
- Audit the other `cnextBaseNames` uses at `Transpiler.ts:989` and `:1038`. `:989` matches a `#include "foo.h"` to the `foo.cnx` that will generate it, so basename keying there may be deliberate — it is **not** confirmed to have the same defect and should be checked rather than changed blindly.
- Make an unresolvable scope reference a hard error instead of emitting the source text.

## Interaction

This currently **masks** the same-basename variant of the include-guard collision in #1133 — with the drop fixed, `can/config.cnx` and `uart/config.cnx` will both generate headers guarded `CONFIG_H`, at which point #1133's Cause 2 becomes reachable inside a single compilation. The two should be fixed together, or #1133 first.

## Acceptance

- [ ] Two included `.cnx` files sharing a basename both compile, in one invocation.
- [ ] No scope reference ever reaches the `.c` output as C-Next source syntax; an unresolved scope is a transpiler error.
- [ ] Regression test with `can/config.cnx` + `uart/config.cnx`, executed (not transpile-only), asserting both scopes' values.

jlaustill:
Closes #1134, closes #1133, closes #1132, closes #1131.

Four issues, one root cause in three costumes: **a name-mangling function that is not injective.** ADR-063 already solved that for scope members by constraining the input domain rather than complicating the mapping. This applies the same discipline to file identity and to the names the transpiler invents for itself.

### #1134 — same-basename includes silently dropped

`Transpiler._processCnextIncludes` computed two identity checks and OR'd them:

```ts
const alreadyExists =
  cnextBaseNames.has(includeBaseName) ||                      // basename — WRONG
  cnextFiles.some((f) => resolve(f.path) === includePath);    // path — already correct
```

`uart/config.cnx` matched the basename of `can/config.cnx`, so it was judged already-included and dropped from the compilation entirely — exit 0, no output for it, and `UartConfig.baudRate()` (C-Next source syntax) emitted verbatim into the C.

`cnextBaseNames` has a second, legitimate consumer (header shadowing in `_collectHeaders`). I probed that path directly: control and colliding runs produce byte-identical output, so it is left alone with a comment recording that it is not a file identity.

### #1133 — include guards were not injective

Guards were the uppercased **basename**, so two files could share one and the preprocessor silently skipped the second. They are now built from the path relative to the **project root**, under the reserved prefix:

```
src/can/config.cnx   ->  CNX_SRC_CAN_CONFIG_H
src/uart/config.cnx  ->  CNX_SRC_UART_CONFIG_H
```

**Project root, not input directory** — this is the subtle part. Anchoring on the input directory means `cnext app.cnx` and `cnext can/config.cnx` disagree about the same file (`CNX_CAN_CONFIG_H` vs `CNX_CONFIG_H`). Any project compiling translation units individually would then reintroduce the collision, which is exactly the cross-invocation case #1133 calls out. The project root is stable across both.

Conversion to upper case stays lossy, so `mod-a.cnx`/`mod_a.cnx` and case-only differences still collide. Those are rejected by the new **E0203** rather than silently sharing a guard — the requirement was never perfect injectivity, only that it is never silent.

### #1132 — the `cnx_` prefix is reserved (E0202)

Declarations may not begin with `cnx_`, compared case-insensitively, prefix-only. Emitted by the same classifier and the same walk as E0201, so the two rules cannot drift apart. Zero user-declared identifiers in the corpus use the prefix.

### #1131 — generated temporaries shadowed user globals

Verified silent miscompiles: a global `_msg_len` was shadowed by the generated strlen cache and both operands of `msg.char_count + _msg_len` resolved to the shadowing local — expected 47, got 10, with a clean `-Wall -Wextra` compile.

Four name families each spelled their own convention, and one was already correct:

| Site | Before | After |
|---|---|---|
| `CodeGenState` | `_tmp<N>` | `cnx_tmp<N>` |
| `ArgumentGenerator` | `_cnx_tmp_<N>` | `cnx_tmp<N>` |
| `CodeGenerator` | `_<var>_len` | `cnx_len_<var>` |
| `OverflowHelperTemplates` | `cnx_clamp_*` | unchanged — already conformed |

They now share one definition in the new `ReservedCnxName` (the `cnx_` analogue of `QualifiedCName`), and `ArgumentGenerator` no longer increments `CodeGenState`'s counter behind its back.

## Why the prefix does not replace the underscore rule

Worth stating explicitly, because the two look interchangeable and are not. A prefix cannot deliver injectivity — prefixing `A_B.c` and `A.B_c` yields `cnx_A_B_c` either way. And neither underscore constraint can be relaxed given the prefix:

- Relax "no `__`": scope `A__B` member `C` and scope `A` member `B__C` both give `A__B__C`.
- Relax "no trailing `_`": scope `A_` member `B` and scope `A` member `_B` both give `A___B`.

`__` answers *which components built this name*; `cnx_` answers *who wrote it*. ADR-063 now carries the proof.

## Also in this PR

- **`guardPrefix` removed** — never set anywhere in `src/`, and an optional second prefix beside a mandatory `CNX_` is a duplicate path to one decision. Its two unit tests now cover path-based guards instead.
- **Examples regenerated** — they still carried the pre-ADR-063 single-underscore separator (`SysTick_Regs`). A concrete instance of the silent drift tracked in #1077.
- **Docs** — ADR-063 amended (Decision, orthogonality proof, guard scheme, alternatives, migration evidence, consequences); README and `docs/error-codes.md` updated for E0202/E0203.

## Verification

| Check | Result |
|---|---|
| Integration | 1025 passed, 0 failed |
| Unit | 6004 passed (222 files) |
| `validate:c` | cppcheck, MISRA, clang-tidy, flawfinder — all passed |
| `tsc --noEmit` | clean |
| oxlint / cspell | 0 issues |

Every fix was written test-first with a **watched RED**: `UartConfig.baudRate()` leaking into C (#1134), a `u64` truncating through an implicit `int` to exit 2 (#1133), "expected errors but transpilation succeeded" (#1132), and `total = _msg_len + _msg_len` exiting 1 (#1131).

The 1924 regenerated snapshots were **audited rather than trusted**: every added line is a guard or temp rename, insertions equal deletions exactly (6492 each), and zero `.expected.error` files changed.

## Filed separately

[#1135](https://github.com/jlaustill/c-next/issues/1135) — `scripts/generate-cpp-snapshots.ts` is non-functional (passes a config field that does not exist, and calls `transpileSource`, removed in the single-entry-point refactor). Found while this work needed a fresh `.expected.cpp`. No workaround was left in the tree; `npm test -- --update` covers the same ground.


jlaustill (commented):
## Review Summary

**PR:** #1136 — fix(naming): reserve the cnx_ prefix and key include identity on path

### Overview

This PR fixes four issues (#1131, #1132, ##1133, #1134) rooted in a single problem: the transpiler's name-mangling functions were not injective. The fix applies ADR-063's discipline — constrain the input domain rather than complicate the mapping — to file identity and invented names. A new `ReservedCnxName` utility unifies all transpiler-invented names under the `cnx_` prefix, include guards are now path-based with a `CNX_` prefix, and two new diagnostics (E0202/E0203) catch the residue that remains lossy.

### Findings

| Severity | Count |
|----------|-------|
| Critical | 0 |
| Important | 1 |
| Moderate | 1 |
| Nit | 2 |

### Key Issues

1. **[important]** `_collectHeaders` basename shadowing is inconsistent with the path-based identity model — `cnextBaseNames` accumulates basenames from C-Next includes across files, and `_collectHeaders` skips any resolved header whose basename matches ANY C-Next file's basename. After this PR's fix, file identity is path-based, so `can/config.cnx` and `uart/config.cnx` are correctly distinct in `_processCnextIncludes`. But `_collectHeaders` still uses basename matching: if `can/config.cnx` registers `"config"` in the set, `uart/config.h` (a completely different external header) will be skipped when processing `uart/config.cnx`. The PR author verified byte-identical output, so this doesn't break current tests, but it's a latent mismatch between the path-based model and the basename heuristic.

2. **[moderate]** The E0203 error location is always `line: 1, column: 0` — IDE error navigation won't jump to anything meaningful. Consider adding a `sourcePath` field with one of the colliding filenames so consumers can locate the issue.

3. **[nit]** `HeaderGenerator.generateFromSymbolTable` computes guards from basenames only (`sourceFile.replace(/\.[^.]+$/, "") + ".h"`), producing guards like `CNX_CONFIG_H` instead of `CNX_SRC_CAN_CONFIG_H`. The method is test-only, but inconsistent with the path-based model.

4. **[nit]** Files like `.config.cnx` produce guards with a leading underscore (`CNX__CONFIG_H`), which is the exact pattern ADR-063's underscore rule forbids in identifiers. Consider stripping leading dots before the sanitize pass.

### What Looks Good

- **Root-cause fix**: Four issues unified under one approach — injectivity via domain constraint, not escape-hatch complexity
- **Test-first methodology**: Every fix written with a watched RED, verified with execution tests that prove the failure mode
- **Comprehensive ADR-063 amendment**: Clearly documents the orthogonality proof, guard scheme, alternatives, and migration evidence
- **Clean architecture**: `ReservedCnxName` as single source of truth for all invented names, removing duplicate conventions and the dead `guardPrefix` option
- **Thorough verification**: 1025 integration tests pass, 6004 unit tests pass, all static analysis clean, 6492 snapshot insertions == deletions with zero `.expected.error` changes

---

PR #1138, opened 2026-08-07
fix(codegen): qualify bare scope-local type names in generated C output (ADR-057, #1130)

### LINKED ISSUE #1130: Bare reference to a scope's own type is not qualified — non-compiling C, mismatched .c/.h signatures, and a fabricated `typedef struct`
## Summary

Inside a scope, a **bare** reference to one of that scope's own types (enum or struct) is emitted verbatim in the generated C instead of being qualified with the scope. The *value* side resolves correctly (`B.c` → `A__B__c`), so only the **type** position is broken. The transpiler exits 0 and the generated C does not compile.

This is the same failure shape as #1117 / #1122: transpile succeeds, the C compiler rejects the output.

Reproduces on `main` @ `1733ba70`.

## Minimal reproduction

```cnx
scope A {
    public enum B { c, d }

    public bool isFirst() {
        B current <- B.c;
        if (current = B.c) {
            return true;
        }
        return false;
    }
}

void main() {
}
```

Generated `.c`:

```c
bool A__isFirst(void) {
    B current = A__B__c;      /* type not qualified; value is */
    if (current == A__B__c) {
        return true;
    }
    return false;
}
```

```
$ gcc -std=c99 -c minimal.c
minimal.c:13:5: error: unknown type name 'B'
   13 |     B current = A__B__c;
      |     ^
```

Per ADR-057 a bare name resolves local → scope → global, and the bare *variable* path already works (`return B_c;` inside `scope A` correctly emits `return A__B_c;`). Only the type path is missed.

## Extended reproduction — three distinct defects

```cnx
scope A {
    public enum B { c, d }

    public struct S { u8 x; }

    public B pick() {
        return B.d;
    }

    public void take(B value) {
        S local <- {x: 1};
        if (value = B.c) { return; }
        return;
    }
}

void main() {
}
```

Generated `.c`:

```c
B A__pick(void) {
    return A__B__d;
}

void A__take(B value) {
    S local = { .x = 1 };
    if (value == A__B__c) {
        return;
    }
    return;
}
```

Generated `.h`:

```c
/* External type dependencies - include appropriate headers */
typedef struct B B;

/* Enumerations */
typedef enum {
    A__B__c = 0,
    A__B__d = 1
} A__B;

/* Struct definitions */
typedef struct A__S {
    uint8_t x;
} A__S;

/* Function prototypes */
B A__pick(void);
void A__take(const B* value);
```

1. **Type positions are not qualified** — `B` instead of `A__B`, `S` instead of `A__S`, in local declarations, parameters and return types alike.
2. **`.c` and `.h` disagree on the signature** — the `.c` emits `void A__take(B value)` (by value) while the `.h` emits `void A__take(const B* value)` (by pointer). The unresolved name falls through to the unknown/opaque-type path in the header generator, which then applies struct-parameter promotion. This is a duplicate-code-path divergence of the kind #914 and #1127 cover.
3. **The header fabricates a type** — `typedef struct B B;` under "External type dependencies". The scope's own enum is classified as an unknown external *struct*.

## Workaround

Prefixing with `this.` resolves correctly and compiles:

```cnx
this.B current <- this.B.c;   /* -> A__B current = A__B__c;  */
```

That workaround is what `tests/scope/issue-1117-depth-ambiguity.test.cnx` (#1128) is using; found while reviewing that PR.

## Expected

Either the bare type name resolves to the scope's type (`A__B` / `A__S`), matching ADR-057 and matching the behaviour the value path already has — or, if bare type references are deliberately not supported, the transpiler emits a diagnostic. Silently emitting non-compiling C with exit 0 is the failure mode ADR-063 and #1117 exist to remove.

## Related

- #1117 / ADR-063 — qualified-name generation
- #1122 — scoped enum as an array dimension resolves differently in `.h` than in `.c`
- #1038 / #1039 — scoped type arrays (`this.Type[N]`) in struct fields
- #1127 — duplicate code path: enum-qualified array dimensions resolved independently for `.c` and `.h`

jlaustill:
Fixes #1130.

## The problem

Inside `scope A`, a bare `B` at a **type** position did not resolve to `A`'s own `B`. The value side already resolved correctly (ADR-057); the type side did not, so generated C carried a name that no longer existed:

- `.c` and `.h` disagreed on the same function's signature
- headers referenced undeclared types (`error: unknown type name 'S'`)
- a `typedef struct B B;` was fabricated for a type that isn't a struct

The transpiler exited 0 in every case.

## The fix

Type names are turned into strings in **two** layers, and the qualification has to happen at both — in the bare `userType()` branch only:

| Layer | Site |
|---|---|
| Symbols | `TypeUtils.dispatchTypeResolution()`, given an `isScopeType` predicate threaded from `CNextResolver` |
| Codegen | `CodeGenerator.getTypeName()`, `TypeRegistrationEngine`, `FunctionContextManager`, `TypeGenerationHelper` |

Three properties the earlier approach couldn't provide:

**1. `global.T` survives.** Qualification now happens while the parse tree still distinguishes `global.Mode` from a bare `Mode`. Once a type name is a resolved string those are byte-identical, so the previous struct-field post-pass rewrote both — silently retyping `global.` fields to the scope-local type with no compile error. That post-pass is deleted.

**2. The header is correct in multi-file builds.** `TSymbol` now carries the qualified name, so `HeaderSymbolAdapter` inherits it rather than re-deriving it from `CodeGenState.symbols` — which holds the *last transpiled file's* symbols at header-generation time, and so emitted bare names for every non-entry file. The header layer no longer touches that shared state.

**3. Declaration order doesn't matter.** New `CNextResolver` Pass 0b collects scope-declared enum/struct/bitmap names before any type is resolved. It collects **types only** — a scope function or variable named `Config` must not capture a global `struct Config` at a type position.

`CodeGenState.qualifyScopeType()` binds the predicate to `currentScope`, so codegen sites stop re-pairing the two.

## Also fixed along the way

`CodeGenerator.getTypeName()` never qualified its `userType()` branch, so `extractFunctionSignature` recorded a bare struct name and the **call site** passed a struct by value into a pointer parameter. `EnumAssignmentValidator` was comparing a bare enum prefix against a now-qualified declared type, and was updated to compare like-for-like.

## Testing

- `tests/scope/issue-1130-scope-type-qualification.test.cnx` promoted from `test-transpile-only` to **`test-execution`** — #1130 is "exits 0, output does not compile", so the C compiler has to be the guard. Extended to cover struct parameters, struct fields, scope variables, `global.T` fields, and non-type members shadowing a global type.
- `bugs/issue-1130-multi-file-scope-type/` — multi-file reproduction; the header defect is invisible to any single-file test.
- Unit tests for `QualifiedCName.qualifyScopeType()`, `CodeGenState.qualifyScopeType()`, and declaration-order independence in the resolver. Verified non-vacuous: the positive cases fail with the fix removed.

6034 unit + 1026 integration tests pass; `validate:c` (cppcheck, MISRA, clang-tidy, flawfinder) clean.

## Out of scope

Two **pre-existing** cross-file struct-parameter bugs surfaced while building the multi-file test — the `.h` drops the auto-const, and the cross-file call site omits the address-of. Both reproduce identically on `main` with a plain global struct, so they are unrelated to scope resolution. Tracked in #1139; the multi-file repro avoids that shape and says so inline.


src/transpiler/output/codegen/helpers/TypeGenerationHelper.ts
```diff
@@ -217,6 +219,15 @@ class TypeGenerationHelper {
       if (typeName === "cstring") {
         return "char*";
       }
+      // ADR-057: bare type name inside a scope — qualify if it's a scope member
+      const qualified = QualifiedCName.qualifyScopeType(
```
jlaustill, line 227: **[critical]** bug: `isCurrentScopeMember` answers *"is this any member of the scope?"*, but this is a **type position** — it needs *"is this a type declared in the scope?"*. `ScopeCollector.ts:88-166` pushes variables, functions, enums, bitmaps, structs and registers into the same list, so a scope member of any kind now captures a same-named global or C-header type.

Reproduced on this branch (compiles clean on `main`):

```cnx
struct Config { u8 x; }
scope A {
    private void Config() { }
    public void use() { Config c <- {x: 1}; }
}
```
`main` → `Config c = { .x = 1U };` · this branch → `A__Config c = { .x = 1U };` where `A__Config` is a function.
`gcc -std=c99`: `error: expected ';' before 'c'`

With a variable member it also regenerates defect #3 of #1130 — the header fabricates a type:

```cnx
struct Servo { u8 pin; }
scope Motor {
    private u8 Servo <- 3;
    public void configure(Servo unit) { unit.pin <- 1; }
}
```
`.h` → `typedef struct Motor__Servo Motor__Servo;`, `.c` → `static uint8_t Motor__Servo = 3U;`
`gcc`: `error: 'Motor__Servo' redeclared as different kind of symbol`

The same flaw is in all three sibling copies (`TypeUtils.ts:64`, `TypeRegistrationEngine.ts:189` and `:502`), so fixing it here alone will not be enough.

Suggested gate — type-ness rather than membership, using the qualified-name sets that already exist in `ICodeGenSymbols`:

```ts
const qualified = QualifiedCName.join(deps.currentScope, typeName);
if (deps.currentScope && deps.isScopeLocalType(qualified)) {
  return qualified;
}
```
where `isScopeLocalType` consults `knownEnums` / `knownStructs` / `knownBitmaps`. Please add a regression test with a scope member colliding with a global struct in type position.

src/transpiler/logic/symbols/cnext/collectors/FunctionCollector.ts
```diff
@@ -38,12 +38,21 @@ class FunctionCollector {
     // Get return type string and convert to TType
     const returnTypeCtx = ctx.type();
     const scopeName = scope.name === "" ? undefined : scope.name;
-    const returnTypeStr = TypeUtils.getTypeName(returnTypeCtx, scopeName);
+    const scopeMembers = scope.name === "" ? undefined : scope.members;
```
jlaustill: **[important]** bug: `scope.members` is still being built when you read it here, so this qualification is declaration-order dependent — while the codegen path is not.

`ScopeCollector.collect()` pushes each member's name and then *immediately* invokes that member's collector (`ScopeCollector.ts:88/105/124/135/150/166`). So `FunctionCollector` at this line only ever sees members declared **above** it in the source. `TypeGenerationHelper` / `TypeRegistrationEngine` instead read `CodeGenState.scopeMembers`, which is fully populated in an earlier stage. The two answers split as soon as a type is declared after its use:

```cnx
scope A {
    public B pick() { return B.d; }
    public enum B { c, d }
}
```
→ `.c`: `A__B A__pick(void)` · `.h`: `B A__pick(void);` plus a fabricated `typedef struct B B;`

Add a global type of the same name and it is a clean compiling → not-compiling regression:

```cnx
struct B { u8 g; }
scope A {
    public B pick(B v) { return v; }
    public struct B { u8 f; }
}
```
`main` → `B A__pick(const B* v)` in **both** files, `gcc -std=c99 -c` clean.
this branch → `.c`: `A__B A__pick(const A__B* v)` vs `.h`: `B A__pick(const B* v);`
`gcc`: `error: conflicting types for 'A__pick'; have 'A__B(const A__B *)'`

C-Next has no forward-declaration requirement, so "declare your types first" is not a rule a user can be expected to know — and the failure looks like a transpiler coin-flip. Either pre-scan `ctx.scopeMember()` to fill `scope.members` before running any sub-collector, or (preferred, see the comment on `TypeRegistrationEngine.ts:189`) have both layers consume one completed member table so the decision is made once.

tests/scope/issue-1130-scope-type-qualification.test.cnx
```diff
@@ -0,0 +1,23 @@
...
+scope A {
+    public enum B { c, d }
+
+    public struct S { u8 x; }
+
+    public B pick() {
+        return B.d;
+    }
+
+    public void take(B value) {
```
jlaustill: **[important]** testing: `take(B value)` takes an **enum**, which is scalar and so never exercises struct-parameter promotion — the exact path defect #2 of #1130 describes. Swapping in the struct shows the `.c`/`.h` signature mismatch is still there:

```cnx
scope A {
    public struct S { u8 x; }
    public u8 read(S cfg) { return cfg.x; }
}
```
→ `.c`: `uint8_t A__read(A__S cfg)` (by value) · `.h`: `uint8_t A__read(const A__S* cfg);` (by pointer)
`gcc`: `error: conflicting types for 'A__read'; have 'uint8_t(A__S)'`

Root cause is a fifth site making this same decision that the PR did not update: `FunctionContextManager.resolveParameterTypeInfo()` (`src/transpiler/output/codegen/helpers/FunctionContextManager.ts:227-235`) returns `typeCtx.userType()!.getText()` raw, and then evaluates `isStruct: callbacks.isStructType(typeName)` on that **unqualified** name — so the scope-local struct is never recognised as a struct in the `.c` path. Its `scopedType()` branch just below (line ~249) already qualifies via `CodeGenState.currentScope`, which is why `this.S cfg` works and `S cfg` does not.

This is broken on `main` too, so it is not a regression — but it is the headline defect of the issue this PR closes, so it should either be fixed here or the PR description should say it is deferred. Please add the bare-struct-parameter case to this test either way.

src/transpiler/output/codegen/helpers/TypeRegistrationEngine.ts
```diff
@@ -184,7 +184,11 @@ class TypeRegistrationEngine {
     }
 
     if (typeCtx.userType()) {
-      return typeCtx.userType()!.getText();
+      const typeName = typeCtx.userType()!.getText();
+      // ADR-057: bare type name inside a scope — qualify if it's a scope member
+      return QualifiedCName.qualifyScopeType(typeName, currentScope, (name) =>
```
jlaustill: **[important]** design: this is the third of four verbatim copies of the same decision — `TypeUtils.ts:64`, `TypeGenerationHelper.ts:223`, here, and `:502` — each carrying its own copy of the comment. `CodeGenState.resolveIdentifier()` (`CodeGenState.ts:892`) already implements exactly this: *"if in a scope and the name is a member, return `QualifiedCName.join(currentScope, name)`"*.

CLAUDE.md's rule is explicit that this is not a style preference: *"Single source of truth means the **decision**, not just the data... Paths that agree only 'by coincidence' are a latent divergence."* That is not hypothetical here — the copies read two different membership sources (`scope.members`, a partially-built array, vs `CodeGenState.scopeMembers`, a completed Set), and the divergence is already firing (see the comment on `FunctionCollector.ts:41`).

Concretely, the cost right now: the kind-awareness fix from the `TypeGenerationHelper.ts:223` comment has to be applied in four places, plus the fifth unpatched site in `FunctionContextManager`. Miss the uncovered copy at `:502` and `scope A { u8 Point <- 5; void go() { Point[2] pts; } }` stays wrong with no test failing.

Suggest one resolver — e.g. `ScopeTypeResolver.qualifyBareType(name)` over a single completed, type-filtered member set — called from all five sites, with the collectors passing nothing extra.

tests/scope/issue-1130-scope-type-qualification.test.cnx
```diff
@@ -0,0 +1,23 @@
+// test-transpile-only
```
jlaustill: **[moderate]** testing: `// test-transpile-only` turns off the one check this fix exists to guarantee. Issue #1130 is *"the transpiler exits 0 and the generated C does not compile"* — and this marker means the harness never hands the output to gcc/g++ (`scripts/test-utils.ts` short-circuits before the compile block). The test degrades to a snapshot diff against a file produced by the very code path under test.

It is also the only transpile-only file among the 49 `.test.cnx` in `tests/scope/`, and the marker is not needed: the committed snapshot compiles cleanly in both modes. Dropping that one line costs nothing and buys gcc + g++ validation on every future change to this rule.

Worth extending the case at the same time, since as written it covers only the three positions that happen to work:
- a bare scope-local **struct parameter** (`public u8 read(S cfg)`) — see the comment on line 15
- a scope whose types are declared **after** the functions that use them
- a scope member colliding with a global type in type position

src/transpiler/output/codegen/helpers/TypeRegistrationEngine.ts
```diff
@@ -490,7 +494,14 @@ class TypeRegistrationEngine {
     }
 
     if (arrayTypeCtx.userType()) {
-      return { baseType: arrayTypeCtx.userType()!.getText(), bitWidth: 0 };
+      const typeName = arrayTypeCtx.userType()!.getText();
+      // ADR-057: bare type name inside a scope — qualify if it's a scope member
+      const qualified = QualifiedCName.qualifyScopeType(
+        typeName,
```
jlaustill: **[moderate]** testing: of the four new qualification branches this is the only one no test reaches. The new `.test.cnx` contains no arrays at all, and a coverage run over the unit suite shows this `if` evaluated but never taken (the `return` at line 504 stays at 0 hits) — so these lines ship with no regression protection and count against the >= 80%-on-new-code gate.

The branch is load-bearing, not dead: it feeds the MISRA literal-suffix decision for scope-local struct/enum arrays. Suggest adding to `tests/scope/issue-1130-scope-type-qualification.test.cnx`:

```cnx
public void useArrays() {
    S[2] items;
    B[2] states;
    items[0].x <- 5;
    states[0] <- B.c;
}
```

plus a direct `TypeRegistrationEngine.test.ts` case that sets `CodeGenState.currentScope` and the scope members. (I confirmed the array path itself works — `A__B modes[3]` / `A__S items[2]` are emitted correctly — so this is purely a coverage gap, not a defect.)

jlaustill (commented):
## Review Summary

**PR:** #1138 — fix(codegen): qualify bare scope-local type names in generated C output

### Overview

The direction is right: issue #1130 asks that a bare scope-local type resolve like the value path already does, and for the happy path (types declared before use, enum/scalar positions, arrays, struct fields) this PR delivers exactly that. I reproduced the fix working for locals, parameters, return types, array element types and struct fields.

But the rule as implemented asks the wrong question. It asks *"is this name a member of the current scope?"* when it needs to ask *"is this name a **type** declared in the current scope?"* — and it asks it against a member list that is only half-built at the time the symbols layer reads it. Both gaps produce generated C that gcc rejects while the transpiler exits 0, which is the exact failure shape ADR-063, #1117 and #1130 exist to remove. Two of them are regressions: input that compiles on `main` does not compile on this branch.

All findings below were reproduced against a local checkout of `db19550b` and compared with `main` @ `056e1ba0`.

### Findings

| Severity | Count |
|----------|-------|
| Critical | 1 |
| Important | 3 |
| Moderate | 2 |
| Nit | 1 |

### Key issues

**1. (critical) The membership test is kind-agnostic.** `scope.members` / `CodeGenState.scopeMembers` hold *every* member name — variables, functions, registers, bitmaps, structs, enums (`ScopeCollector.ts:88-166`). Nothing checks type-ness, so a scope member of any kind captures a same-named global or C-header type in type position:

```cnx
struct Config { u8 x; }
scope A {
    private void Config() { }
    public void use() { Config c <- {x: 1}; }
}
```
`main`: `Config c = { .x = 1U };` — compiles. This branch: `A__Config c = { .x = 1U };` — `error: expected ';' before 'c'`.

With a *variable* member it also re-creates defect #3 from the issue (the header fabricates a type):
```cnx
struct Servo { u8 pin; }
scope Motor {
    private u8 Servo <- 3;
    public void configure(Servo unit) { unit.pin <- 1; }
}
```
`.h` gains `typedef struct Motor__Servo Motor__Servo;` while `.c` has `static uint8_t Motor__Servo = 3U;` → `error: 'Motor__Servo' redeclared as different kind of symbol`. On `main` this compiles clean.

**2. (important) The fix is declaration-order dependent, and the `.c` and `.h` disagree because of it.** `ScopeCollector.collect()` pushes each member name and immediately runs that member's collector, so `FunctionCollector` sees only the members declared *above* it. The codegen path reads the fully-built `CodeGenState.scopeMembers` instead. Move the type below its use and the two answers split:

```cnx
scope A {
    public B pick() { return B.d; }
    public enum B { c, d }
}
```
→ `.c`: `A__B A__pick(void)` · `.h`: `B A__pick(void);` plus a bogus `typedef struct B B;`

Add a global type of the same name and it becomes a clean compiling → not-compiling regression:
```cnx
struct B { u8 g; }
scope A {
    public B pick(B v) { return v; }
    public struct B { u8 f; }
}
```
`main`: `B A__pick(const B* v)` in both files, gcc clean. This branch: `A__B A__pick(const A__B* v)` in `.c` vs `B A__pick(const B* v);` in `.h` → `error: conflicting types for 'A__pick'`.

**3. (important) Issue #1130's struct-parameter case is not fixed.** `FunctionContextManager.resolveParameterTypeInfo()` (`src/transpiler/output/codegen/helpers/FunctionContextManager.ts:227-235`) is a fifth site making this same decision and was not updated, so `isStruct` is still decided from the *unqualified* name:
```cnx
scope A {
    public struct S { u8 x; }
    public u8 read(S cfg) { return cfg.x; }
}
```
→ `.c`: `uint8_t A__read(A__S cfg)` (by value) · `.h`: `uint8_t A__read(const A__S* cfg);` (by pointer) → `error: conflicting types for 'A__read'`. Still broken on `main` too, so not a regression — but this is precisely defect #2 in the issue this PR closes. The new test dodges it by using an enum parameter, which is scalar.

**4. (important) The decision is now spelled out four times** (`TypeUtils.ts:64`, `TypeGenerationHelper.ts:223`, `TypeRegistrationEngine.ts:189` and `:502`) over two independently-timed data sources — and `CodeGenState.resolveIdentifier()` (`CodeGenState.ts:892`) already encodes exactly this decision. Finding 2 *is* that divergence firing. Per CLAUDE.md, single source of truth means the decision, not just the data.

### Suggested shape of the fix

Resolve the type name **once**, kind-aware, behind one predicate that both layers consume — e.g. qualify only when `QualifiedCName.join(scope, name)` is in `knownEnums` / `knownStructs` / `knownBitmaps`, over a member table that is complete before any collector reads it. That single change would close findings 1, 2 and 4 together, and give finding 3 one place to hook into. Where a non-type scope member genuinely shadows a global type, the issue text explicitly allows a diagnostic as the alternative to resolution — anything but silent non-compiling C.

### What looks good

- The core insight is right and the happy path genuinely works — I verified qualification at variable, parameter, return, array-element and struct-field positions, in both C and C++ mode.
- Threading `scope.members` through all five `TypeUtils.getTypeName()` call sites was thorough; no collector was missed on that path.
- Committing all four snapshot flavours (`.c`/`.cpp`/`.h`/`.hpp`) matches project convention, and the snapshots do match actual output.

---

src/transpiler/output/codegen/__tests__/CodeGenerator.test.ts
```diff
@@ -763,10 +763,10 @@ describe("CodeGenerator", () => {
       it("should add temp declaration that can be flushed", () => {
         const generator = createMinimalGenerator(`void foo() { }`);
 
-        generator.addPendingTempDeclaration("int cnx_tmp1 = 0;");
+        generator.addPendingTempDeclaration("int _tmp1 = 0;");
```
jlaustill, line 766: **[nit]** quality: this looks like an unintended revert. Commit `2ab255c1` on `main` (four commits ago, ADR-063 part 2 / #1131) changed these exact two lines *to* `cnx_tmp1` when it reserved the `cnx_` prefix and routed every invented name through `ReservedCnxName`. This PR changes them back to `_tmp1`, and nothing else here touches temporaries.

`addPendingTempDeclaration` / `flushPendingTempDeclarations` are pass-throughs, so the test passes either way — which is exactly why it will not get caught. The cost is documentary: `_tmp1` is a name a user is still allowed to declare (`ReservedCnxName.isReserved("_tmp1") === false`), so the sample string now demonstrates the spelling #1131 was filed to abolish.

```suggestion
        generator.addPendingTempDeclaration("int cnx_tmp1 = 0;");
```

(line 769 needs the same restore). Might be worth diffing the branch against `main` for any other stale-buffer reverts of `2ab255c1`.

src/transpiler/output/headers/adapters/HeaderSymbolAdapter.ts
```diff
@@ -59,29 +59,48 @@ class HeaderSymbolAdapter {
...
-      type: TypeResolver.getTypeName(p.type),
-      isConst: p.isConst,
-      isArray: p.isArray,
-      arrayDimensions: p.arrayDimensions?.map((d) =>
-        typeof d === "number" ? String(d) : d,
-      ),
-      isAutoConst: p.isAutoConst,
-    }));
+    // ADR-057: qualify bare scope-local type names so .h matches .c
+    const isScopeType = (qn: string) =>
```
jlaustill: **[critical]** bug: This helper is local to `convertFunction`, so qualification reaches function parameters and return types only. Defects 2 and 3 from #1130 survive at every other type position.

`convertVariable` (line 121) still does `const typeStr = TypeResolver.getTypeName(variable.type)` with no qualification, and header struct fields come in through `IHeaderTypeInput.structFields`, which this PR does not touch.

**Scope-level variable** — the `.c` is now qualified and the `.h` is not, so they no longer even agree with each other:

```
scope A {
    public enum B { c, d }
    public struct S { u8 x; }

    public B current <- B.c;
    public S state <- {x: 1};

    public void touch() { current <- B.d; }
}

void main() { }
```

Generated `.h`:

```c
/* External type dependencies - include appropriate headers */
typedef struct B B;     /* fabricated — defect 3 */
typedef struct S S;

extern B A__current;    /* defect 2 */
extern S A__state;
```

Generated `.c`: `A__B A__current = A__B__c;`

```
p1.c:10:6: error: conflicting types for 'A__current'; have 'A__B'
p1.h:32:10: note: previous declaration of 'A__current' with type 'B'
p1.c:11:6: error: conflicting types for 'A__state'; have 'A__S'
p1.h:33:10: note: previous declaration of 'A__state' with type 'S'
```

**Struct field** — same input, plus `public struct Outer { B kind; u8 y; }`:

```c
typedef struct A__Outer {
    B kind;             /* unqualified */
    uint8_t y;
} A__Outer;
```

```
p2.h:24:5: error: unknown type name 'B'
```

Suggestion: rather than adding the same call at `convertVariable` and again in the struct-field map, resolve the type name once where it is produced — `TypeResolver.getTypeName(type)` already receives the symbol, and the symbol carries its `scope` — so every conversion inherits the qualification instead of each one opting in. Otherwise the next type position added to the header starts out broken again.

src/transpiler/output/codegen/helpers/FunctionContextManager.ts
```diff
@@ -226,10 +226,20 @@ class FunctionContextManager {
...
+        CodeGenState.currentScope,
+        (qn) =>
+          CodeGenState.symbols?.knownEnums.has(qn) ||
+          CodeGenState.symbols?.knownStructs.has(qn) ||
+          CodeGenState.symbols?.knownBitmaps.has(qn) ||
+          false,
+      );
+      return {
+        typeName: qualified,
+        isStruct: callbacks.isStructType(qualified),
```
jlaustill, line 234: **[critical]** bug: Qualifying `typeName` makes `isStruct` true for a bare scope-local struct here, but `ParameterInputAdapter.fromAST` still derives its classification from the **unqualified** name, so the signature and the body disagree:

```ts
// ParameterInputAdapter.ts:107-109, 146
const typeName = deps.getTypeName(ctx.type());       // "Outer"
const mappedType = deps.generateType(ctx.type());    // now "A__Outer"
...
const isKnownStruct = deps.isKnownStruct(typeName);  // false — asked with "Outer"
```

`isKnownStruct` false → `isPassByReference` false → `ParameterSignatureBuilder._buildUnknownParam` → pass by value, while member access here has already committed to `->`:

```
scope A {
    public struct Outer { u8 y; }

    public void fill(Outer o) { o.y <- 2; }
}
```

```c
void A__fill(A__Outer o) {   /* by value */
    o->y = 2U;               /* dereferenced */
}
```

```
p2.c:12:6: error: invalid type argument of '->' (have 'A__Outer')
p2.c:11:6: error: conflicting types for 'A__fill'; have 'void(A__Outer)'
p2.h:29:6: note: previous declaration of 'A__fill' with type 'void(A__Outer *)'
```

On `main` the `.c` was at least internally consistent (`void A__fill(Outer o) { o.y = 2; }`) — this failure shape is new. Writing `this.Outer` produces the correct `void A__fill(A__Outer* o) { o->y = 2U; }` in both files, which is what the bare form should match.

This is the "decision, not just the data" case from `CLAUDE.md`: `mappedType` now comes from the fixed path while `isKnownStruct` / `isPassByReference` / `isTypedefStructType` / `isOpaqueType` are all re-derived from the raw name. Qualify once, before the classification, so every consumer reads the same name — rather than qualifying at each site that happens to be noticed.

src/transpiler/output/codegen/CodeGenerator.ts
```diff
@@ -628,6 +628,11 @@ export default class CodeGenerator implements IOrchestrator {
         CodeGenState.symbolTable.checkNeedsStructKeyword(name),
       validateCrossScopeVisibility: (scope, member) =>
         ScopeResolver.validateCrossScopeVisibility(scope, member),
+      isScopeType: (qualifiedName) =>
```
jlaustill: **[important]** design: This predicate is inlined verbatim five times in this PR:

- `CodeGenerator.ts:631`
- `FunctionContextManager.ts:233`
- `TypeRegistrationEngine.ts:192`
- `TypeRegistrationEngine.ts:508`
- `HeaderSymbolAdapter.ts:65`

Adding a fourth type kind means editing five places — the anti-pattern `CLAUDE.md` marks ZERO EXCEPTIONS. `CodeGenState` already owns this question, and `isKnownStruct()` already folds bitmaps in:

```ts
// CodeGenState.ts:552
static isKnownStruct(name: string): boolean {
  if (this.symbols?.knownStructs.has(name)) return true;
  if (this.symbols?.knownBitmaps.has(name)) return true;
  ...
}
```

So all five copies are spelled-out equivalents of `CodeGenState.isKnownEnum(qn) || CodeGenState.isKnownStruct(qn)`. `HeaderSymbolAdapter` makes it especially visible: `resolveArrayDimension` calls `CodeGenState.isKnownEnum()` about 165 lines below the new inline lambda, so that one file now asks the same question two different ways.

Suggest one `CodeGenState.isKnownType(name)` next to the existing accessors. That also lets `qualifyScopeType(typeName, currentScope)` drop its third parameter at four of the five sites — keep the injected `deps.isScopeType` in `TypeGenerationHelper`, since that helper deliberately takes all its state through `ITypeGenerationDeps`, but wire it to the shared accessor.

jlaustill (commented):
## Review Summary

**PR:** #1138 — fix(codegen): qualify bare scope-local type names in generated C output (ADR-057, #1130)

### Overview

The direction is right and `QualifiedCName.qualifyScopeType()` is the correct place for this: checking the *qualified* name against the known-type sets (rather than the bare name against scope members) is a genuinely better rule than what the PR description proposes, because it stops a same-named function or variable member from capturing a global type at a type position.

But the fix is applied call-site by call-site rather than at the point where the name is resolved, and it stops at function signatures. Three of the positions #1130 names are still broken, and the partial application introduces a new failure shape that `main` did not have. I verified everything below by transpiling against this branch and compiling with `gcc -std=c99`.

### Findings

| Severity | Count |
|----------|-------|
| Critical | 2 |
| Important | 3 |
| Moderate | 2 |
| Nit | 1 |

### Key Issues

**1. Scope-level variables and struct fields are still unqualified** (inline: `HeaderSymbolAdapter.ts:65`). Only `convertFunction` qualifies. A `public B current <- B.c;` inside `scope A` now emits `A__B A__current` in the `.c` and `extern B A__current;` in the `.h`, plus the fabricated `typedef struct B B;` — defects 2 and 3 of #1130, verbatim, one position over:

```
p1.c:10:6: error: conflicting types for 'A__current'; have 'A__B'
p1.h:32:10: note: previous declaration of 'A__current' with type 'B'
```

Struct fields are the same story: `typedef struct A__Outer { B kind; ... }` → `error: unknown type name 'B'`.

**2. Struct parameters now generate `.c` that is internally inconsistent** (inline: `FunctionContextManager.ts:241`). `ParameterInputAdapter.fromAST` still classifies from the unqualified name while `mappedType` now comes back qualified, so a bare scope-local struct parameter emits a by-value signature with an arrow-dereferencing body:

```c
void A__fill(A__Outer o) {   /* by value */
    o->y = 2U;               /* dereferenced */
}
```

On `main` the `.c` was at least self-consistent (`void A__fill(Outer o) { o.y = 2; }`). This is the exact "single source of truth means the *decision*, not just the data" case from `CLAUDE.md` — the name is unified, the consequences are still re-derived independently.

**3. Five copies of one predicate** (inline: `CodeGenerator.ts:631`). `knownEnums || knownStructs || knownBitmaps` is inlined verbatim at five sites, when `CodeGenState.isKnownEnum()` / `isKnownStruct()` already answer exactly this (and `isKnownStruct` already folds in bitmaps). `HeaderSymbolAdapter` now asks the same question two different ways in one file.

**4. The regression test cannot regress** (inline: the `.test.cnx`). `// test-transpile-only` skips compilation, but #1130 *is* "exits 0, output does not compile." The three cases covered are also the three that work; the three that are broken are absent.

**5. `qualifyScopeType` has no unit test** (inline: `QualifiedCName.ts:105`). It is the only method in that class without a `describe` block, and its `isKnownType → true` branch is unit-uncovered against the 80% new-code gate.

### Also worth a look

The PR description does not describe this diff. It claims an `isCurrentScopeMember` dep taking `scopeMembers: readonly string[]`, and changes to `TypeUtils`, `FunctionCollector`, `StructCollector`, `VariableCollector`, and `RegisterCollector` — none of which are in the diff. The implemented design (`isScopeType(qualifiedName)`) is the better one, so the description just needs to catch up before merge.

### What Looks Good

- Putting the helper on `QualifiedCName` alongside `join`/`split`/`isInScope` is the right home, and the doc comment explaining *why* it keys on the qualified name is the kind of note that keeps a subtle rule from being "simplified" away later.
- `CodeGenerator.test.ts:8581` updating to `void Motor__setState(Motor__State s)` is a real assertion that the fix works, not a snapshot rubber-stamp.
- Injecting `isScopeType` through `ITypeGenerationDeps` rather than reaching for `CodeGenState` inside `TypeGenerationHelper` keeps that helper testable, consistent with how the rest of its deps are wired.

---

tests/scope/issue-1130-scope-type-qualification.test.cnx
jlaustill: **[important]** testing: `// test-transpile-only` skips compilation entirely (`scripts/test-utils.ts:307` → `TestUtils`), so this only snapshot-compares text. #1130 is specifically *"the transpiler exits 0 and the generated C does not compile"* — a transpile-only test structurally cannot catch a regression of that class. Dropping the marker makes the harness run gcc in both C and C++ modes, which is what actually pins the fix.

The three cases here — enum return type, enum parameter, struct **local** — are also the three that already work after the change. The positions that are still broken are the ones not covered:

```
scope A {
    public enum B { c, d }
    public struct S { u8 x; }
    public struct Outer { B kind; }   // header emits `B kind;`

    public B current <- B.c;          // header emits `extern B A__current;`

    public void fill(Outer o) {       // .c emits `A__Outer o` with `o->y`
        o.kind <- B.d;
    }
}
```

Adding those three declarations to this file turns it into a test that fails today, which is what a regression test for #1130 should do.

src/utils/QualifiedCName.ts
```diff
@@ -83,6 +83,37 @@ class QualifiedCName {
...
+   * bitmap) capture the name — a function or variable member with the same
+   * leaf name does not shadow a global type at a type position.
+   *
+   * @param typeName       Bare C-Next type name from the AST (e.g., "B")
+   * @param currentScope   The scope currently being processed, or null for global
+   * @param isKnownType    Check if the *qualified* name is a known type
+   *                       (e.g., isKnownType("A__B") → true for enum B in scope A)
+   * @returns Qualified C name (e.g., "A__B") or the bare name if not a scope type
+   */
+  static qualifyScopeType(
```
jlaustill, line 105: **[important]** testing: New public API with no unit test. `src/utils/__tests__/QualifiedCName.test.ts` has a `describe` block for `join`, `split`, `isQualified`, `isInScope`, `toCppQualified`, and ADR-063 injectivity — `qualifyScopeType` is now the only method without one.

The only unit-level exercise is `isScopeType: () => false` in `TypeGenerationHelper.test.ts:233`, and the other four call sites read `CodeGenState.symbols`, which is null in unit tests. So the `isKnownType(qualified) → true` branch — the entire point of the change — is unit-uncovered, against the >= 80% new-code gate.

Three cases worth pinning down, since the rest of the fix rests on these semantics:

```ts
expect(QualifiedCName.qualifyScopeType("B", null, () => true)).toBe("B");     // global: never qualified
expect(QualifiedCName.qualifyScopeType("B", "A", () => true)).toBe("A__B");   // scope declares it
expect(QualifiedCName.qualifyScopeType("B", "A", () => false)).toBe("B");     // global type not shadowed
```

That third one is the interesting behavior your doc comment describes — a function or variable member named `B` must not capture a global type `B` — and nothing currently asserts it.

src/transpiler/output/codegen/helpers/TypeGenerationHelper.ts
```diff
@@ -217,6 +223,15 @@ class TypeGenerationHelper {
       if (typeName === "cstring") {
         return "char*";
       }
+      // ADR-057: bare type name inside a scope — qualify if it's a scope type
+      const qualified = QualifiedCName.qualifyScopeType(
+        typeName,
+        deps.currentScope,
+        deps.isScopeType,
+      );
+      if (qualified !== typeName) {
```
jlaustill, line 232: **[moderate]** quality: The early return bypasses the rest of the `userType()` path, so `checkNeedsStructKeyword` / `generateUserType` become unreachable for any scope-local type. It happens to work because a scope struct always gets `typedef struct A__S { ... } A__S;`, but that is an unstated invariant, and the two branches now leave the function through different code — if a scope-local struct ever lands in `needsStructKeyword`, the `struct` keyword silently disappears with nothing to catch it.

Single exit, using whichever name is actually going to be emitted:

```ts
const qualified = QualifiedCName.qualifyScopeType(
  typeName,
  deps.currentScope,
  deps.isScopeType,
);
const needsStruct = deps.checkNeedsStructKeyword(qualified);
return TypeGenerationHelper.generateUserType(qualified, needsStruct);
```

Worth confirming `SymbolTable.needsStructKeyword` is keyed on generated C names before making the swap — if it is keyed on bare source names, the lookup needs `typeName` and the emitted name needs `qualified`, which is worth a comment either way.

src/transpiler/output/codegen/__tests__/CodeGenerator.test.ts
jlaustill, line 766: **[moderate]** quality: This reverts `cnx_tmp1` back to `_tmp1` (here and on line 769), undoing part of 2ab255c1 *"fix(naming): reserve the cnx_ prefix and key include identity on path"* — which is an ancestor of this branch, so it is a revert rather than a stale base.

The assertion passes either way, since `addPendingTempDeclaration` just stores and returns the string, so CI cannot see it. But it leaves the test contradicting the naming rule that commit established, and it has nothing to do with #1130. Restoring `cnx_tmp1` on both lines keeps the convention intact.

src/transpiler/output/codegen/helpers/TypeRegistrationEngine.ts
```diff
@@ -490,7 +500,18 @@ class TypeRegistrationEngine {
     }
 
     if (arrayTypeCtx.userType()) {
-      return { baseType: arrayTypeCtx.userType()!.getText(), bitWidth: 0 };
+      const typeName = arrayTypeCtx.userType()!.getText();
+      // ADR-057: bare type name inside a scope — qualify if it's a scope type
+      const qualified = QualifiedCName.qualifyScopeType(
+        typeName,
+        CodeGenState.currentScope,
```
jlaustill: Nit: **[nit]** quality: `_resolveBaseTypeWithCallbacks` (line 191) passes the `currentScope` **parameter** into `qualifyScopeType`, while this one reads the `CodeGenState.currentScope` **global** for the same question. Two sources for "which scope am I in" within one file is how they eventually disagree — worth settling on one. The parameter is the safer default, since it is what the caller actually intends rather than whatever the walker last set.

jlaustill (commented):
## Review Summary

**PR:** #1138 — fix(codegen): qualify bare scope-local type names in generated C output (ADR-057, #1130)

### Overview

The core idea is right — `QualifiedCName.qualifyScopeType()` is a good home for the type-side counterpart to the value-side resolver, and it genuinely fixes the reported case (bare enum/struct at local-declaration, return, and array-element positions). But the qualification is applied to a subset of the paths that resolve type names, so **all three defects listed in #1130 still reproduce on this branch** with slightly different inputs. Each reduction below was transpiled on `c03f2a81` and compiled with `gcc -std=c99 -c`.

### Findings

| Severity | Count |
|----------|-------|
| Critical | 3 |
| Important | 2 |
| Moderate | 3 |
| Nit | 0 |

### Key issues

**Three cases still emit non-compiling C:**

1. **Scope-local struct as a parameter** — `.c` emits `void A__take(A__S value)` with `value->x` in the body, `.h` emits `void A__take(const A__S* value)`. Two gcc errors: conflicting types, and `->` on a non-pointer. This is defect #2 from the issue, and structs are the *only* type category where pointer promotion applies — so the defect is unfixed exactly where it can occur. The `->` in the body is a **regression**: `main` produced a self-consistent `value.x`.
2. **Scope variable of a scope-local type** — `.h` still emits the fabricated `typedef struct B B;` (defect #3, explicitly claimed fixed) plus `extern B A__current;` against the `.c`'s `A__B A__current` → conflicting types. `HeaderSymbolAdapter.convertVariable()` never got the new qualification.
3. **Struct field of a scope-local type** — header emits `B kind; Inner nested;` → unknown type name. Field types reach the header via `structFields`, a path this PR doesn't touch.

**The unifying cause:** `TypeUtils.getTypeName()` still returns the bare name (`TypeUtils.ts:60-61`). The PR description says it was changed — it isn't in the diff — and it's the one edit that would close (1) and feed the header paths behind (2) and (3).

**Duplicate code path** — the `knownEnums || knownStructs || knownBitmaps` predicate is copy-pasted at five call sites while `CodeGenState.isKnownEnum/isKnownStruct/isKnownBitmap` already exist, and the copies are subtly *narrower* than `isKnownStruct`. jscpd won't flag it (the block is under the 6-line/50-token threshold), so the passing gate isn't evidence it's fine.

Separately, `CodeGenerator.test.ts` reverts `cnx_tmp1` → `_tmp1`, undoing 2ab255c1 ("reserve the cnx_ prefix").

### What looks good

- `QualifiedCName` is the right home for this, and the docstring explaining **why** the check keys on the *qualified* name — so a same-named function or variable member can't shadow a global type at a type position — is genuinely clarifying and a non-obvious call to get right.
- Keying on the joined name rather than raw scope membership is the conservative choice; it avoids widening ADR-057 resolution behavior as a side effect, which this project treats as requiring its own ADR.
- Snapshot discipline is correct — `.c`/`.h`/`.cpp`/`.hpp` all regenerated and committed per project convention.
- The array-element path (`_extractArrayBaseTypeInfo`) does work; `B[2] modes` and `S[2] items` both qualify and compile cleanly.

### Suggested next step

Push the qualification down into `TypeUtils.getTypeName()` / the shared type-resolution path so every consumer (`ParameterInputAdapter.fromAST`, `convertVariable`, the `structFields` header path) inherits it from one place rather than each site opting in — that also collapses the five predicate copies. Then extend the regression test to a struct parameter, a struct field, and a scope variable, and drop `test-transpile-only` so the C compiler actually guards the three defects the issue enumerates.

---

src/transpiler/output/codegen/helpers/FunctionContextManager.ts
jlaustill, line 234: **[critical]** bug: A scope-local **struct** parameter still produces non-compiling C — and the `.c` body regresses relative to `main`.

```cnx
scope A {
    public struct S { u8 x; }
    public void take(S value) { u8 y <- value.x; }
}
```

On this branch:

```c
/* .c */ void A__take(A__S value) { uint8_t y = value->x; }   /* by value, but -> */
/* .h */ void A__take(const A__S* value);                     /* by pointer       */
```

```
error: conflicting types for 'A__take'; have 'void(A__S)'
error: invalid type argument of '->' (have 'A__S')
```

Root cause: `isStruct` here now sees the qualified `A__S`, so the body switches to pointer member access — but the `.c` *signature* decision runs through `ParameterInputAdapter.fromAST` → `TypeUtils.getTypeName()` (`TypeUtils.ts:60-61`), which still returns the bare `S`. `isKnownStruct("S")` is `false`, so no pointer promotion. The `.h` goes through `fromSymbol`, which this PR *did* qualify — so the two adapters now disagree harder than before.

On `main` the `.c` was at least internally consistent (`void A__take(S value) { … value.x; }`); the `->` is new here.

This is defect #2 from #1130 ("`.c` and `.h` disagree on the signature"), and structs are the only category where pointer promotion applies, so it's unfixed where it matters. Qualifying inside `TypeUtils.getTypeName()` — which the PR description says was changed, but the diff doesn't contain — would fix the signature and the body together instead of moving the mismatch.

src/transpiler/output/headers/adapters/HeaderSymbolAdapter.ts
jlaustill: **[critical]** bug: `isScopeType` is only wired into `convertFunction`. `convertVariable()` (line 113) still emits the raw type at line 120 (`const typeStr = TypeResolver.getTypeName(variable.type)`), so a scope variable of a scope-local type yields a `.h` that contradicts the `.c` — and still carries the fabricated typedef this PR claims to remove.

```cnx
scope A {
    public enum B { c, d }
    public B current <- B.c;
}
```

```c
/* .c */ A__B A__current = A__B__c;

/* .h */ typedef struct B B;    /* defect #3 from #1130 — still emitted */
         extern B A__current;
```

```
error: conflicting types for 'A__current'; have 'A__B'
note: previous declaration of 'A__current' with type 'B'
```

Worth noting this is *worse* than `main` in one respect: there both sides said `B` (broken but agreeing), whereas now the `.c` says `A__B` and the `.h` says `B`, so it fails as a conflicting declaration. `convertVariable` needs the same `qualifyScopeType` treatment as the parameter/return path just below.

src/transpiler/output/headers/adapters/HeaderSymbolAdapter.ts
```diff
@@ -59,29 +59,48 @@ class HeaderSymbolAdapter {
...
+        isArray: p.isArray,
+        arrayDimensions: p.arrayDimensions?.map((d) =>
+          typeof d === "number" ? String(d) : d,
+        ),
+        isAutoConst: p.isAutoConst,
+      };
+    });
+
+    // Build signature with qualified return type and param types
+    const qualifiedReturn = QualifiedCName.qualifyScopeType(
```
jlaustill: **[critical]** bug: Struct **field** types aren't qualified either, so the generated header doesn't compile:

```cnx
scope A {
    public enum B { c, d }
    public struct Inner { u8 x; }
    public struct Outer { B kind; Inner nested; }
}
```

```c
typedef struct A__Outer {
    B kind;         /* should be A__B     */
    Inner nested;   /* should be A__Inner */
} A__Outer;
```

```
error: unknown type name 'B'
error: unknown type name 'Inner'
```

Field types reach the header through `structFields` (`IHeaderTypeInput.ts:18`), not through `convertFunction`, so this path is untouched — identical output to `main`. Defect #1 of #1130 ("type positions are not qualified") survives for struct fields.

This plus the `convertVariable` case above is the argument for qualifying once inside the shared type resolution rather than per-consumer: each new call site has to remember to opt in, and two of them didn't.

src/transpiler/output/codegen/CodeGenerator.ts
jlaustill: **[important]** design: This known-type predicate is copy-pasted at five sites — here, `FunctionContextManager.ts:232`, `TypeRegistrationEngine.ts:191` and `:507`, and `HeaderSymbolAdapter.ts:65`. Adding a sixth type kind (registers, typedefs, …) means editing five places in lockstep — the pattern CLAUDE.md calls "the WORST anti-pattern in this project."

`CodeGenState` already owns this decision: `isKnownEnum()`, `isKnownStruct()`, `isKnownBitmap()` (`CodeGenState.ts:552-578`). The five inlined copies also **diverge** from it — `CodeGenState.isKnownStruct()` additionally consults `symbolTable.getStructFields(name)`, which none of the copies do. They agree only coincidentally today, which is the "latent divergence" case the rule calls out.

One accessor would collapse all five:

```ts
// CodeGenState.ts, next to the existing isKnownEnum/isKnownStruct/isKnownBitmap
static isKnownType(name: string): boolean {
  return this.isKnownEnum(name) || this.isKnownStruct(name) || this.isKnownBitmap(name);
}
```

then pass `CodeGenState.isKnownType` at every call site. Note jscpd won't catch this (the block is under the configured 6-line / 50-token minimum, and project-wide duplication stays at 0.47% vs. the 0.86% threshold) — so the green gate isn't evidence the duplication is acceptable.

src/utils/QualifiedCName.ts
jlaustill, line 105: **[important]** quality: The helper itself is well-placed and the docstring is good — but the PR description doesn't describe this diff.

It says the fix touches `TypeUtils` (`dispatchTypeResolution()` accepting `scopeMembers: readonly string[]`), `FunctionCollector`, `StructCollector`, `VariableCollector`, and `RegisterCollector`, and that `ITypeGenerationDeps` gained `isCurrentScopeMember`. None of those are in the diff; the actual dep is `isScopeType`, and no collector or `TypeUtils` file is touched.

Worth reconciling before merge for two reasons: the body lands in the merge commit as the record of what changed, and the missing `TypeUtils.getTypeName()` change is exactly what the struct-parameter bug needs.

The "Generated C code compiles cleanly with `gcc -std=c99`" claim should be narrowed too — it holds for the one new test, but not for scope-local struct parameters, scope variables, or struct fields (see the other comments).

tests/scope/issue-1130-scope-type-qualification.test.cnx
jlaustill: **[moderate]** testing: The test covers a bare enum at return and parameter positions and a bare struct as a **local**, but not a bare struct as a **parameter** — which is the one shape where the `.c`/`.h` signature divergence (defect #2) can actually arise, and it's still broken. Two other shapes from the issue are also uncovered and still broken: a struct **field** of a scope-local type, and a **scope-level variable** of a scope-local type.

Adding a struct parameter here would have caught it:

```cnx
    public void takeStruct(S value) {
        u8 y <- value.x;
    }
```

Related: the file is `// test-transpile-only`, so nothing in it is ever handed to a C compiler. All three defects in #1130 are "transpiles fine, gcc rejects it" failures, so a transpile-only snapshot can't guard the class of bug the issue is about. Dropping the marker (the file already has a `main`, so it runs in both C and C++ modes) would put gcc back in the loop.

src/transpiler/output/codegen/__tests__/CodeGenerator.test.ts
```diff
@@ -763,10 +763,10 @@ describe("CodeGenerator", () => {
...
 
         const decls = generator.flushPendingTempDeclarations();
-        expect(decls).toBe("int cnx_tmp1 = 0;");
+        expect(decls).toBe("int _tmp1 = 0;");
```
jlaustill, line 769: **[moderate]** quality: This is unrelated to the fix and reverts recent work — `main` has `cnx_tmp1` here, from 2ab255c1 "fix(naming): reserve the cnx_ prefix and key include identity on path."

Because `addPendingTempDeclaration` is a pass-through, the test passes with either string, so nothing fails — it just quietly stops documenting the reserved prefix. Looks like a rebase artifact.

```suggestion
        generator.addPendingTempDeclaration("int cnx_tmp1 = 0;");

        const decls = generator.flushPendingTempDeclarations();
        expect(decls).toBe("int cnx_tmp1 = 0;");
```

src/transpiler/output/codegen/helpers/__tests__/TypeGenerationHelper.test.ts
```diff
@@ -230,6 +230,7 @@ describe("TypeGenerationHelper", () => {
       isCppScopeSymbol: () => false,
       checkNeedsStructKeyword: () => false,
       validateCrossScopeVisibility: vi.fn(),
+      isScopeType: () => false,
```
jlaustill, line 233: **[moderate]** testing: `isScopeType` is stubbed `() => false` here, and this is the only occurrence of `isScopeType` in any unit test — so the new qualification branch at `TypeGenerationHelper.ts:227-232` never executes with a `true` result under `npm run unit`. `QualifiedCName.qualifyScopeType()` has no test at all in the existing `src/utils/__tests__/QualifiedCName.test.ts`.

All the new behavior is currently covered only by one transpile-only integration snapshot. With the ≥80% new-code coverage gate, worth adding:

- a case here with `isScopeType: (qn) => qn === "A__B"` asserting the qualified result
- direct `qualifyScopeType()` tests for its three branches: `currentScope` null, qualified name known, qualified name unknown

src/transpiler/output/headers/adapters/HeaderSymbolAdapter.ts
```diff
@@ -59,29 +59,44 @@ class HeaderSymbolAdapter {
...
-      type: TypeResolver.getTypeName(p.type),
-      isConst: p.isConst,
-      isArray: p.isArray,
-      arrayDimensions: p.arrayDimensions?.map((d) =>
-        typeof d === "number" ? String(d) : d,
-      ),
-      isAutoConst: p.isAutoConst,
-    }));
+    // ADR-057: qualify bare scope-local type names so .h matches .c
+    const isScopeType = (qn: string) => CodeGenState.isScopeType(qn);
```
jlaustill: **[critical]** bug: `CodeGenState.symbols` is a single mutable global holding the **last transpiled file's** symbols. Headers are emitted in a separate later pass — `Transpiler._generateAllHeadersFromPipeline()` (`Transpiler.ts:979`) runs after every file's `.c` has been generated — so for every file except the entry file, `isScopeType()` here consults the wrong symbol table and the `.h` keeps the bare name.

`mergeExternalEnums()` masks this for enums (it merges `knownEnums` across includes) but it does **not** merge `knownStructs` / `knownBitmaps`, so struct and bitmap types fall straight through.

Repro — two files, one entry point:

```cnx
// dep.cnx
scope A {
    public struct S { u8 x; }
    public S build() { S out <- {x: 7}; return out; }
    public u8 read(S value) { return value.x; }
}
```
```cnx
// app.cnx
#include "dep.cnx"
scope Z { public enum Q { e, f } public Q other() { return Q.e; } }
void main() { }
```

`dep.c` — correct, this PR's fix:
```c
A__S A__build(void) { ... }
uint8_t A__read(const A__S* value) { ... }
```

`dep.h` — still #1130:
```c
S A__build(void);          /* error: unknown type name 'S' */
uint8_t A__read(S* value); /* and the const is gone too */
```

Swapping which file is the entry point flips which header comes out correct, which pins it on `CodeGenState.symbols` rather than on anything about the source. So defect 2 of #1130 is fixed only for single-file builds — and in one respect widened: on `main` both sides said `S` (consistently broken); now the `.c` says `const A__S*` and the `.h` says `S*`, so they disagree on the type *and* on constness.

See the summary for a single-point fix that removes the need for this call entirely.

src/transpiler/logic/symbols/cnext/adapters/TSymbolInfoAdapter.ts
```diff
@@ -162,6 +162,26 @@ class TSymbolInfoAdapter {
       }
     }
 
+    // ADR-057: Post-process struct fields to qualify bare scope-local type names.
+    // This must run AFTER all symbols are processed so knownEnums/knownStructs/
+    // knownBitmaps are complete.
+    // Iterate over structFields entries to extract scope from qualified struct name.
+    for (const [structQName, fields] of [...structFields]) {
```
jlaustill: **[critical]** bug: this pass runs over `structFields`, a flat `Map<string, Map<string, string>>` built *after* `TypeUtils.getTypeName()` has already collapsed `global.Mode`, `this.Mode` and bare `Mode` into the same string `"Mode"`. It therefore rewrites explicit `global.X` field types to `Scope__X`, silently.

```cnx
enum Mode { alpha, beta }

scope A {
    public enum Mode { one, two }

    public struct Holder {
        global.Mode outer;   // explicitly the GLOBAL enum
        Mode inner;          // scope-local (ADR-057)
    }

    public void use() {
        Holder h <- {outer: global.Mode.alpha, inner: Mode.one};
        return;
    }
}
```

`main` emits `Mode outer; Mode inner;` — `inner` wrong, which is what #1130 is about. This branch emits:

```c
typedef struct A__Holder {
    A__Mode outer;   /* was global.Mode — now the wrong enum */
    A__Mode inner;   /* correct */
} A__Holder;
```

while the `.c` still initialises `.outer = Mode__alpha`. Both are `int`-compatible in C, so nothing errors — the field simply has the wrong declared type and the header misstates the struct's contract to any C/C++ caller. A silent wrong-type is a worse failure mode than the non-compiling C #1130 started from.

I confirmed the attribution by deleting just this loop: `outer` goes back to `Mode` (and `inner` goes back to being wrong), so this is the source.

`bugs/issue-478/` is exactly this `global.X`-shadows-scope-`X` scenario, which is why it's worth guarding — it only covers return types today, so it stays green.

The underlying issue is that the decision is made here, downstream of the point where the distinction still exists. `TypeUtils.dispatchTypeResolution()` already branches on `globalType()` vs `scopedType()` vs `userType()`; qualifying there instead lets this whole loop be deleted.

src/transpiler/output/codegen/CodeGenerator.ts
```diff
@@ -3920,7 +3921,14 @@ export default class CodeGenerator implements IOrchestrator {
       generateType: (t) => this.generateType(t),
       generateExpression: (e) => this.generateExpression(e),
       callbackTypes: CodeGenState.callbackTypes,
-      isKnownStruct: (t) => this.isKnownStruct(t),
+      isKnownStruct: (t) => {
+        if (this.isKnownStruct(t)) return true;
+        // ADR-057: check qualified name for scope-local struct types only
```
jlaustill, line 3928: **[important]** design: this hand-rolls `QualifiedCName.join()` plus the null-scope guard instead of calling the helper this PR just added, and it checks only `knownStructs` where `CodeGenState.isScopeType()` checks enums, structs and bitmaps.

That makes four independent encodings of "is this bare name a scope-local type" on this branch:

1. `QualifiedCName.qualifyScopeType()` + `CodeGenState.isScopeType()` — enum/struct/bitmap
2. `TSymbolInfoAdapter.ts:171-181` — inline `knownEnums.has || knownStructs.has || knownBitmaps.has`
3. this block — inline join + `knownStructs` only
4. `TypeUtils.dispatchTypeResolution()` (`logic/symbols/cnext/utils/TypeUtils.ts:60`) — still returns the bare name

Per `CLAUDE.md`, sharing one helper is not enough when each path re-derives the consequences independently: extending the rule later (registers, opaque types, nested lookup) means landing the change in four places, and #3 and #1 already disagree about which type kinds count.

Minimally, route this through `QualifiedCName.qualifyScopeType()` with a struct-only predicate so the join and the null-scope guard live in one place. Better, qualify once during symbol collection so `t` arrives already qualified and this override isn't needed at all.

src/transpiler/output/codegen/helpers/FunctionContextManager.ts
```diff
@@ -226,10 +226,16 @@ class FunctionContextManager {
...
+      // ADR-057: bare type name inside a scope — qualify if it's a scope type
+      const qualified = QualifiedCName.qualifyScopeType(
         typeName,
-        isStruct: callbacks.isStructType(typeName),
-        isCallback: CodeGenState.callbackTypes.has(typeName),
+        CodeGenState.currentScope,
+        (qn) => CodeGenState.isScopeType(qn),
+      );
+      return {
+        typeName: qualified,
```
jlaustill, line 233: **[important]** design: the scalar `userType()` branch now qualifies, but the `arrayType()` → `userType()` branch in this same function (line 301) still uses the bare name for `typeName`, `callbacks.isStructType()` and `CodeGenState.callbackTypes.has()`. So a `B[4]` parameter inside `scope A` registers as `B` while a `B` parameter registers as `A__B`.

`TypeRegistrationEngine._tryRegisterUserTypeArray()` (line 512) has the same split — it passes the bare name to `_tryRegisterEnumOrBitmapType()`, so an array of a scope-local enum never matches there and falls through to `_extractArrayBaseTypeInfo()`, which *does* qualify.

To be straight about the impact: I could not produce a user-visible symptom. Emission goes through `TypeGenerationHelper`, which is qualified on both branches, and `S[3]` / `B[3]` parameters and locals all come out correct and compile:

```c
uint8_t A__sumStructs(A__S items[3]);
bool A__firstIsC(A__B modes[3]);
```

So this is latent, not broken today — but it is the same "two paths agree by coincidence" shape the rest of the PR is fixing, and worth closing while the code is open.

tests/scope/issue-1130-scope-type-qualification.test.cnx
jlaustill: **[moderate]** testing: `// test-transpile-only` makes the framework skip compilation entirely — `scripts/test-utils.ts:839-846` sets `compileSuccess = true` and returns before the compiler runs — so this test asserts only that the snapshot text matches.

#1130 is specifically *"the transpiler exits 0 and the generated C does not compile"*. Dropping the marker turns on the `gcc -fsyntax-only -std=c99` / `g++ -std=c++14` check, which is the guard that actually pins this class of bug and would fail loudly on any regression.

Both generated artifacts already compile clean, so the marker can just be removed:

```
$ gcc -fsyntax-only -std=c99 -Wno-unused-variable -Wno-main -I tests/include -I . issue-1130-scope-type-qualification.test.c    # ok
$ g++ -fsyntax-only -std=c++14 -Wno-unused-variable -I tests/include -I . issue-1130-scope-type-qualification.test.cpp          # ok
```

Separately, the case this test covers is single-file, which is why both critical findings above pass CI. A companion multi-file case (an included `.cnx` declaring the scope, plus a struct-typed parameter, plus a `global.X` field where the scope also declares `X`) would cover them.

src/utils/QualifiedCName.ts
jlaustill, line 105: **[moderate]** testing: `qualifyScopeType()` has no direct unit tests, even though `src/utils/__tests__/QualifiedCName.test.ts` exists and covers the rest of the class. Same for the new `CodeGenState.isScopeType()`. `TypeGenerationHelper.test.ts` only adds `isScopeType: () => false` to the shared mock, so the new true-branch is never exercised there either.

Four cases worth pinning directly, since they are the helper's actual contract:

- `currentScope === null` → returns `typeName` unchanged
- qualified name is a known type → returns the joined name
- qualified name is not a known type → returns the bare name (global fallthrough)
- the case the doc comment specifically claims: a scope **function or variable** whose leaf name matches a global *type* must not capture it at a type position

That last one is the interesting invariant and the reason this keys on the qualified name rather than on `scope.members` — it should have a test that fails if someone later "simplifies" it to a members lookup.

src/transpiler/output/codegen/__tests__/CodeGenerator.test.ts
jlaustill, line 766: **[moderate]** quality: this reverts `cnx_tmp1` → `_tmp1`, undoing the convention established by 2ab255c1 ("reserve the `cnx_` prefix", #1136), which landed on `main` two commits before this branch's base.

It is an opaque string literal passed through `addPendingTempDeclaration()`, so nothing functional changes — but it is an unrelated edit that contradicts a just-landed convention, and `_tmp1` is the reserved-identifier form that commit deliberately moved away from. Worth restoring both lines (769 too).

```suggestion
        generator.addPendingTempDeclaration("int cnx_tmp1 = 0;");
```

jlaustill (commented):
## Review Summary

**PR:** #1138 — fix(codegen): qualify bare scope-local type names in generated C output (ADR-057, #1130)

### Overview

The core idea is right and the `.c` side of #1130 is genuinely fixed — bare enum/struct names inside a scope now resolve to `A__B` at variable, parameter and return-type positions, and `A__S` / `const A__S*` come out correct. `QualifiedCName.qualifyScopeType()` is a good shape for a shared helper, and its doc comment does the hard work of explaining *why* it keys on the qualified name rather than the bare one.

Two problems keep it from actually closing #1130. I reproduced both against this branch:

1. In a **multi-file** build the `.h` still emits the bare name — defect 2 of the issue (`.c`/`.h` disagree) is only fixed when there is a single file.
2. `global.Type` struct fields are silently rewritten to the scope-local type — a new and quieter failure than the one being fixed.

Underneath both is one thing: the qualification decision is made **too late**, at several independent downstream sites, on data from which the `this.` / `global.` / bare distinction has already been erased — and, in the header's case, against a global that no longer describes the file being emitted.

### Findings

| Severity | Count |
|----------|-------|
| Critical | 2 |
| Important | 2 |
| Moderate | 4 |
| Nit | 1 |

### Key issues

- **[critical]** `HeaderSymbolAdapter` reads `CodeGenState.symbols`, which holds the *last transpiled file's* symbols by the time headers are emitted in `Transpiler._generateAllHeadersFromPipeline()`. Every non-entry file's `.h` keeps the bare name. `mergeExternalEnums()` hides this for enums but not structs/bitmaps.
- **[critical]** The new `TSymbolInfoAdapter` post-pass rewrites `global.Mode` struct fields to `A__Mode` when the scope declares its own `Mode`. No compile error — just a wrong declared type in the header while the `.c` still initialises it from the global enum.
- **[important]** Four independent encodings of "is this bare name a scope-local type" now exist (`qualifyScopeType` + `isScopeType`, the `TSymbolInfoAdapter` inline check, the `CodeGenerator` inline check, and `TypeUtils.dispatchTypeResolution` which still returns the bare name).
- **[important]** `FunctionContextManager.resolveParameterTypeInfo()` qualifies its scalar `userType()` branch but not the array branch in the same function; same split in `TypeRegistrationEngine._tryRegisterUserTypeArray()`.
- **[moderate]** `// test-transpile-only` on the new regression test skips compilation entirely — the one check that would pin this class of bug.

### The one-place fix

`TypeUtils.dispatchTypeResolution()` (`src/transpiler/logic/symbols/cnext/utils/TypeUtils.ts:60`) looks like the natural single point — and it is what this PR's own description says was done ("**`TypeUtils`**: Same check in `dispatchTypeResolution()`"). It is the only site that:

- still has the AST, so `scopedType` / `globalType` / `qualifiedType` / `userType` are distinguishable — it already handles the first three, and only the `userType()` branch returns a bare name;
- already receives `scopeName`;
- is funnelled through by **every** collector: `FunctionCollector.ts:41,114`, `StructCollector.ts:170`, `VariableCollector.ts:138`, `RegisterCollector.ts:51`, each doing `TypeUtils.getTypeName(ctx, scopeName)` → `TypeResolver.resolve(...)`.

Qualify there and struct fields, parameters, return types, scope variables and register members all carry the qualified name in `TSymbol` itself. `HeaderSymbolAdapter` then gets it from `TypeResolver.getTypeName(p.type)` for free (critical 1 dissolves — no dependence on `CodeGenState.symbols` at header time), the `TSymbolInfoAdapter` post-pass can be deleted (critical 2 dissolves, and `global.` survives), and the `CodeGenerator` inline variant becomes unnecessary.

One real catch, worth stating plainly: at collection time the known-type sets are not built yet, so the predicate cannot be `isScopeType()`. The description's `scopeMembers` answer works but over-captures — a scope *function* named `B` would capture a global type `B` at a type position, which is precisely what the `qualifyScopeType()` doc comment says it avoids. A cheap pre-pass collecting only type declarations per scope before types are resolved gets both properties. Either way, the decision wants to be made once, where the AST still records which of `B` / `this.B` / `global.B` was written.

### What looks good

- `QualifiedCName.qualifyScopeType()` is a genuinely good abstraction, and the comment explaining why it checks the qualified name against the known-type sets is the kind of reasoning that belongs in the code.
- Fixing `.c` and `.h` in one change, with matching `.expected.h` / `.expected.hpp` snapshots, is the right instinct for a defect that is fundamentally about the two sides disagreeing.
- Every ADR-057 site carries a comment naming the ADR — consistent with the compliance-annotation standard in `CLAUDE.md`.
- The `Motor__State s` update in `CodeGenerator.test.ts` is a real behavioural assertion, not a snapshot rubber-stamp.

### Two housekeeping notes

- **The PR description does not match the diff.** It describes `isCurrentScopeMember` on `ITypeGenerationDeps` and edits to `TypeUtils`, `FunctionCollector`, `StructCollector`, `VariableCollector` and `RegisterCollector` — none of which are in the change. Worth syncing before merge, especially since the described design is the stronger one.
- **Consider a `bugs/issue-1130-*/` multi-file reproduction.** Both criticals are invisible to any single-file test. `bugs/` already holds 32 such cases, and `bugs/issue-478/` is this exact `global.X`-shadowing-scope-`X` shape (return types only, which is why it stays green here).

---

src/transpiler/output/codegen/helpers/TypeGenerationHelper.ts
```diff
@@ -217,6 +223,15 @@ class TypeGenerationHelper {
...
+        return qualified;
+      }
       const needsStruct = deps.checkNeedsStructKeyword(typeName);
       return TypeGenerationHelper.generateUserType(typeName, needsStruct);
```
jlaustill, line 236: Nit: quality: the early return on the qualified path skips `deps.checkNeedsStructKeyword()`, so a scope-local type can never receive the `struct ` prefix.

That is correct today only because `needsStructKeyword` is populated exclusively from non-typedef'd C structs (`logic/symbols/c/collectors/StructCollector.ts:81`), which are always global — so the two branches agree by coincidence rather than by construction. Folding the qualified name back through the existing call keeps it true by construction and drops a branch:

```suggestion
      const needsStruct = deps.checkNeedsStructKeyword(qualified);
      return TypeGenerationHelper.generateUserType(qualified, needsStruct);
```

src/transpiler/output/codegen/helpers/TypeGenerationHelper.ts
jlaustill, line 232: Nit: quality: the early return on the qualified path skips `deps.checkNeedsStructKeyword()`, so a scope-local type can never receive the `struct ` prefix.

That is correct today only because `needsStructKeyword` is populated exclusively from non-typedef'd C structs (`logic/symbols/c/collectors/StructCollector.ts:81`), which are always global — so the two branches agree by coincidence rather than by construction. Folding the qualified name back through the call already below drops the branch and keeps it true by construction:

```ts
const needsStruct = deps.checkNeedsStructKeyword(qualified);
return TypeGenerationHelper.generateUserType(qualified, needsStruct);
```

(This was the 8th finding in my review summary — GitHub dropped it as a multi-line suggestion.)

jlaustill:
Addressed in `b8491e7c`. The reviews across all four rounds converged on the same diagnosis — the decision was being made too late, at several independent sites — so this reworks it rather than patching the sites.

**Both criticals from the latest round are fixed, and I reproduced both first.**

**1. `global.Mode` struct fields rewritten to `A__Mode`** — confirmed, and confirmed a regression: `main` emits `Mode kind;`, this branch emitted `A__Mode kind;`, silently, with no compile error. Root cause is that the post-pass ran over resolved *strings*, where `global.Mode` and a bare `Mode` are byte-identical — it could not have been made correct in place. Qualification now happens in `TypeUtils.dispatchTypeResolution()`'s `userType()` branch, while the parse tree still distinguishes them; `globalType()` keeps its own branch and is never qualified. The post-pass is deleted (which also cleared the two SonarCloud `S7747` findings that lived in it).

**2. `HeaderSymbolAdapter` reading the last file's symbols** — confirmed on a two-file build: `lib.h` had `uint8_t A__pick(S* v)` against `lib.c`'s `const A__S*` → `error: unknown type name 'S'`. Your read on `mergeExternalEnums()` was exactly right: enums survived, structs didn't. `TSymbol` now carries the qualified name from the symbols layer, so the header adapter inherits it and no longer consults `CodeGenState.symbols` at all.

**On the suggested one-place fix** — I took it, including the pre-pass. `CNextResolver` Pass 0b collects scope-declared enum/struct/bitmap names before any type is resolved, modelled on the existing `collectBitmapsPass1`. As you noted, `scope.members` would be wrong twice over: kind-agnostic, and still being populated while collectors read it.

One thing the analysis missed: `TypeUtils` alone isn't sufficient. `CodeGenerator.getTypeName()` never qualified its `userType()` branch either, so `extractFunctionSignature` recorded a bare struct name and the **call site** passed a struct by value into a pointer parameter. Type names get resolved in two layers and both needed the fix. `CodeGenState.qualifyScopeType()` now binds the predicate to `currentScope` so the codegen sites stop re-pairing the two.

**On `test-transpile-only`** — agreed, and raised further: the test is now `test-execution`, so it compiles *and* runs. Extended to struct parameters, struct fields, scope variables, `global.T` fields, and a non-type member shadowing a global type. Added `bugs/issue-1130-multi-file-scope-type/` per your suggestion, since both criticals were invisible to any single-file test. Unit tests added for `qualifyScopeType`, `CodeGenState.qualifyScopeType`, and declaration-order independence — verified non-vacuous by removing the fix and watching them fail.

**Two corrections.**

*The `cnx_tmp1` → `_tmp1` revert (raised in three rounds) is not real.* `main` has `_tmp1` at that line too — `git diff main...HEAD` on `CodeGenerator.test.ts` shows exactly one changed line, the intentional `Motor__State` assertion. That string is opaque input handed to `addPendingTempDeclaration()`, not a generated name; the `cnx_` convention from `2ab255c1` governs generated temporaries, and those still assert `cnx_tmp0` in `ArgumentGenerator.test.ts`.

*The remaining `.c`/`.h` mismatch on cross-file struct parameters is not this defect.* It reproduces identically on `main` with a plain global struct, in two parts: the header drops the auto-const, and the cross-file call site omits the address-of. Filed as #1139 with a minimal reproduction. The multi-file test deliberately avoids that shape and says so inline, rather than encoding the broken behaviour in a snapshot.

Full suite green: 6034 unit + 1026 integration, `validate:c` clean across cppcheck, MISRA, clang-tidy and flawfinder. PR description rewritten to match the diff.

PR #1140, opened 2026-08-09
fix: preserve isAutoConst for cross-file function parameters (issue #1139)

### LINKED ISSUE #1139: Cross-file struct parameter: header drops auto-const and call site omits address-of — non-compiling C
## Summary

When a scope function in an **included** `.cnx` file takes a **struct parameter**, the generated code is broken in two independent ways. Both reproduce with a plain global struct, so this is unrelated to scope-local type resolution (#1130) — it is the cross-file struct-parameter path itself.

1. **The header drops the auto-const.** The `.c` definition gets `const GS*` (auto-const, since the parameter is never modified) but the `.h` prototype gets `GS*` → `error: conflicting types`.
2. **The cross-file call site omits the address-of.** The caller emits `Sensors__readValue(sample)` passing a struct by value into a pointer parameter → `error: incompatible type for argument 1`.

The single-file equivalent of (2) is correct — the caller emits `&sample`. Only the cross-file path is affected.

## Reproduction

`glib.cnx`:
```cnx
struct GS { u8 x; }
scope Sensors {
    public u8 readValue(GS sample) { return sample.x; }
}
```

`gmain.cnx`:
```cnx
#include "glib.cnx"
u32 main() {
    GS sample <- {x: 42};
    u8 v <- Sensors.readValue(sample);
    if (v != 42) return 1;
    return 0;
}
```

```
cnext gmain.cnx
```

## Actual output

`glib.c`:
```c
uint8_t Sensors__readValue(const GS* sample) {
```

`glib.h`:
```c
uint8_t Sensors__readValue(GS* sample);
```

`gmain.c`:
```c
uint8_t v = Sensors__readValue(sample);
```

```
glib.c: error: conflicting types for 'Sensors__readValue';
  have 'uint8_t(const GS *)'
glib.h: note: previous declaration with type 'uint8_t(GS *)'
gmain.c: error: incompatible type for argument 1 of 'Sensors__readValue'
  note: expected 'GS *' but argument is of type 'GS'
```

The transpiler exits 0.

## Expected

`.h` prototype matches the `.c` definition (`const GS*`), and the call site emits `&sample`.

## Notes

- Verified on `main` @ `1733ba70`. Not a regression — the same output appears on older commits.
- Found while adding a multi-file regression test for #1130; that test currently has to avoid a cross-file struct parameter because of this bug.
- The single-file case works, so the divergence is in how the callee's parameter info is resolved for cross-file calls (`CallExprUtils.resolveTargetParam` falling back to the `SymbolTable`) versus the local `functionSignatures` path — and, for the const half, in `Transpiler.convertToHeaderSymbols()` not carrying `isAutoConst` into the prototype.

jlaustill:
When a scope function in an included .cnx file takes a struct parameter, the transpiler was dropping the auto-const qualifier in the generated header and omitting the address-of operator in cross-file call sites.

The fix ensures that CallExprUtils.resolveTargetParam preserves the isAutoConst property when resolving cross-file function parameters, allowing proper header generation and call site construction.

Fixes #1139

### DIFF
--- a/src/transpiler/output/codegen/generators/expressions/CallExprUtils.ts
+++ b/src/transpiler/output/codegen/generators/expressions/CallExprUtils.ts
@@ -102,6 +102,7 @@ class CallExprUtils {
               baseType,
               isConst: p.isConst,
               isArray: p.isArray,
+              isAutoConst: p.isAutoConst,
             },
             isCrossFile: true,
           };


src/transpiler/output/codegen/generators/expressions/CallExprUtils.ts
```diff
@@ -102,6 +102,7 @@ class CallExprUtils {
               baseType,
               isConst: p.isConst,
               isArray: p.isArray,
+              isAutoConst: p.isAutoConst,
```
jlaustill, line 105: **[critical]** bug: This does not typecheck, and CI's **Build** job has already failed on exactly these two errors (`npm run typecheck`):

```
CallExprUtils.ts(105,15): error TS2353: Object literal may only specify known properties,
  and 'isAutoConst' does not exist in type '{ name: string; baseType: string; isConst: boolean; isArray: boolean; }'.
CallExprUtils.ts(105,30): error TS2339: Property 'isAutoConst' does not exist on type
  'IParameterInfo | ICParameterInfo | ICppParameterInfo'.
  Property 'isAutoConst' does not exist on type 'ICParameterInfo'.
```

They are two independent problems, and neither is fixable by a cast:

1. **Write side.** The declared return type is `IFunctionSignature["parameters"][0]`, and `IFunctionSignature` (`src/transpiler/output/codegen/types/IFunctionSignature.ts`) declares only `name` / `baseType` / `isConst` / `isArray`. Carrying `isAutoConst` through here requires widening that interface — which in turn means deciding what the field means for the *local* signature path too, so the two branches of `resolveTargetParam` don't diverge.

2. **Read side.** `sym.parameters[paramIndex]` is `IParameterInfo | ICParameterInfo | ICppParameterInfo`. Only the C-Next `IParameterInfo` has `isAutoConst`; `ICParameterInfo` and `ICppParameterInfo` are plain `{ name, type, isConst, isArray }`. Auto-const is a C-Next-only concept (ADR-006), so this read needs narrowing to the C-Next symbol before it is valid — an imported C prototype has no auto-const to preserve.

src/transpiler/output/codegen/generators/expressions/CallExprUtils.ts
```diff
@@ -102,6 +102,7 @@ class CallExprUtils {
...
             },
             isCrossFile: true,
```
jlaustill, line 107: **[critical]** bug: Beyond the type errors, this change is a **no-op** — #1139 still reproduces in full on `c66cc662`.

I checked the branch out and ran the issue's exact reproduction through `npx tsx src/index.ts` (tsx skips typechecking, so the new line really does execute). Output is byte-identical to the "Actual output" section of #1139:

```c
/* glib.h  */ uint8_t Sensors__readValue(GS* sample);         /* const still dropped */
/* glib.c  */ uint8_t Sensors__readValue(const GS* sample) {  /* still conflicting   */
/* gmain.c */ uint8_t v = Sensors__readValue(sample);         /* still no &          */
```

`gcc` still emits both errors verbatim — `conflicting types for 'Sensors__readValue'` and `incompatible type for argument 1`.

The reason is that nothing consumes the new field: `grep -n isAutoConst CallExprGenerator.ts` returns no matches. The resolved param is only ever read for `.baseType`, `.isArray`, and the type predicates in `_shouldPassByValue`, so writing `isAutoConst` here cannot change generated output.

**Half 1 (header drops const) is also in a different layer entirely.** `glib.h` and `glib.c` are both generated from `glib.cnx`; that prototype never passes through `resolveTargetParam`, which resolves *call-site* target params. So even a compiling version of this change could not fix that half.

---

Root causes, in case they're useful — both verified on this branch:

**Half 1 — header drops auto-const.** This is a state-lifetime bug, not a lossy-copy bug. Transpiling `glib.cnx` **standalone** produces a correct header:

```c
uint8_t Sensors__readValue(const GS* sample);   /* glib.cnx alone — correct */
```

The const only disappears when `glib.cnx` is pulled in as an include of `gmain.cnx`. In `Transpiler.convertToHeaderSymbols()`, `shouldAutoConst` requires `unmodifiedParams.get(headerSymbol.name)` to be present, and that map comes from `CodeGenState.getUnmodifiedParameters()`, which iterates `CodeGenState.functionSignatures`. When the included file's header is emitted, that map no longer holds the included file's functions, so `unmodified` is `undefined` and auto-const is silently dropped — no warning, exit code 0.

**Half 2 — call site omits `&`.** A name-space mismatch, one layer above this line. `resolveTargetParam` is called with `funcExpr`, which for a scope function is the C-mangled `Sensors__readValue` (`QualifiedNameGenerator.forFunctionStrings`), but `SymbolTable` keys C-Next function symbols by their **bare** name plus a separate `scope` reference (`FunctionCollector`: `const name = ctx.IDENTIFIER().getText()`). So `symbolTable.getOverloads("Sensors__readValue")` returns `[]`, this whole block never runs, `param` stays `undefined`, and `generateFunctionArg(e, undefined)` has no type to decide address-of from.

Adding a top-level (unmangled) function to the same included file confirms it — the bare name resolves and gets its `&`, the mangled one doesn't:

```c
uint8_t a = topLevelRead(&sample);        /* bare name  -> symbol found     */
uint8_t b = Sensors__readValue(sample);   /* mangled    -> lookup miss, bug */
```

So the fix for this half is to make the lookup agree with how symbols are keyed (unmangle, or look up by scope + bare name), not to add fields to the object built after a lookup that never succeeds.

jlaustill (commented):
## Review Summary

**PR:** #1140 — fix: preserve isAutoConst for cross-file function parameters (issue #1139)

### Overview

One line added to `CallExprUtils.resolveTargetParam`. I checked out `c66cc662` and tested it against issue #1139's own reproduction. Two blocking problems: it does not compile, and — run via `tsx`, which skips typechecking so the new line actually executes — it does not change the generated output. Both halves of #1139 reproduce verbatim, with both `gcc` errors intact.

### Findings

| Severity | Count |
|----------|-------|
| Critical | 2 |
| Important | 2 |
| Moderate | 0 |
| Nit | 0 |

### Key issues

1. **Does not typecheck (critical, inline).** `npm run typecheck` fails with TS2353 and TS2339; CI's **Build** job already failed on exactly those. `IFunctionSignature` has no `isAutoConst` field, and `isAutoConst` does not exist on `ICParameterInfo` / `ICppParameterInfo` in the read union.

2. **No-op even if it compiled (critical, inline).** Nothing reads `isAutoConst` off the resolved param — `grep -n isAutoConst CallExprGenerator.ts` returns no matches. The field is written and never consumed, so it cannot affect codegen. Verified empirically against the issue's repro.

3. **Wrong layer for half 1 (important).** The dropped header const is produced by `Transpiler.convertToHeaderSymbols()` from the *same* file (`glib.cnx` → `glib.c` + `glib.h`). It never passes through call-site parameter resolution, so no change in `CallExprUtils` can reach it. The issue is explicit that #1139 is two independent bugs; this PR targets neither successfully, and a fix will need to touch both paths.

4. **No regression test (important).** `bugs/` is the project's convention for exactly this (`bugs/issue-<name>/`, committed with the fix, guarded by the **Bugs Regression** CI job), and #1139 ships a ready-made two-file reproduction. A test here would have caught points 2 and 3 before CI — the change passes `npm run build` (esbuild, no typecheck) and breaks no existing test, so the only signals that it doesn't work are `typecheck` and a repro that doesn't exist yet. Worth adding the `bugs/` case first and watching it fail.

### Root causes

Both verified on this branch; details and evidence are in the inline comment.

- **Half 1 (header const):** state lifetime, not lossy copying. `glib.cnx` transpiled *standalone* emits a correct `const GS*` prototype; the const is only lost when the file is transpiled as an include, because `CodeGenState.functionSignatures` no longer holds the included file's functions when `getUnmodifiedParameters()` is consulted.
- **Half 2 (missing `&`):** `resolveTargetParam` is queried with the C-mangled `Sensors__readValue`, but `SymbolTable` keys C-Next functions by bare name plus a `scope` reference — the lookup never hits, so the block this PR edits never executes for the reported case. A top-level function in the same file, whose name isn't mangled, resolves fine and gets its `&`.

### What looks good

- The instinct about `resolveTargetParam` is directionally sound: the cross-file branch does rebuild the parameter from scratch and silently drops every field outside the four hardcoded ones. That lossy-projection shape is worth fixing on its own — it just isn't what breaks #1139, and fixing it needs the type work described inline.
- Minimal, focused diff with a clear issue reference in both title and body.

---

